PR #64, opened 2022-09-12
replace sidebar table icon

### LINKED ISSUE #58: Arrows suggest table hierarchy
Tables listed in the sidebar show an arrow on their left side, which suggests kind of hierarchy / submenu of tables. 
Those arrows are used for example by the Notes app for categories, which can be opened or closed.
AFAIK (from an user view) there is no table hierarchy implemented on this tables app nor can tables be moved around. 
So I would suggest to remove the arrows.

ganimed-es:
fixes #58

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -90,3 +90,7 @@ p code {
 .align-right {
     text-align: right;
 }
+
+.app-navigation-entry__children {
+    display: none !important;
+}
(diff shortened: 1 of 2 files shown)


datenangebot (approved):
This is a preperation for the views that will go in the collapse part. But you are right, for now. So I will merge but I guess this will be replaced soon. :)

PR #75, opened 2023-01-05
Create dependabot.yml

juliushaertl:
To keep the dependencies up to date :)


datenangebot (approved):
Fine with that. Had a call about that to explain.

PR #88, opened 2023-01-05
ci: Add node build to GitHub actions

juliushaertl:
This will make sure that we have a step that builds the javascript so we can be sure that this continues to work


juliushaertl:
The build artefacts are not used in an way, this is just to ensure the build passes

PR #95, opened 2023-01-13
tests/psalm

juliushaertl:
Setup psalm for static code analysis of PHP code. I tried to mute only irrelevant errors due to external dependencies and fix the rest (mostly by changing annotations a bit). One fix is contained as a separate commit which psalm figured out that there was a wrong order of parameters being used.


datenangebot (approved):
Fine, thx!

PR #98, opened 2023-01-16
fix: Use NcContent component instead of div

juliushaertl:
This fixes a few issues with the container layout on Nextcloud 25, e.g. on smaller window height


datenangebot (approved):
Okay, seems to be fine

Issue #110, opened 2023-01-31
Add pagination for long tables


datenangebot:
https://github.com/nextcloud/tables/issues/279 covers the frontend part, this issue would cover the backend

juliushaertl:
This is actually a duplicate of https://github.com/nextcloud/tables/issues/941 which has some more insights already, let's close this one

Issue #109, opened 2023-01-31
Make table scrolling sticky


datenangebot:

datenangebot:
close as duplicate, track #613

Issue #107, opened 2023-01-31
Print layout for tables


datenangebot:

FahrJo:
@datenangebot do you already have any concrete plans/ideas for the print layout?

Issue #28, opened 2022-03-26
import data


datenangebot:
there are some errors if the column mapping is in trouble -> maybe 2 steps to import
1. paste and show
2. save in BE, if okay

datenangebot:
import for tables have to be redesigned

datenangebot:
we had to remove this kind of importing. So the issue is outdated.

PR #111, opened 2023-01-31
Feat/v0.3.0

datenangebot:
This is a PR draft for the upcoming tables release v0.3.0 with the invitation for testing.
With this release are some bugs fixed and the table is replaced with an own vue component.

Known bugs:
- #109
- #108

Missing features due to the table replacement
- #28 
- #107 
- on demand filters/sorting (will be covered with the views feature)
- resize columns  (will be covered with the views feature)
- #110


Makefile
```diff
@@ -104,23 +101,19 @@ lint-php-lint:
...
 lint-php-ncversion:
 	# Check min-version consistency
-	# TODO needs to be fixed
 	# php tests/nextcloud-version.php
 
 lint-php-phan:
 	# PHAN
-	# TODO needs to be fixed
-	# vendor/bin/phan --allow-polyfill-parser -k tests/phan-config.php --no-progress-bar -m checkstyle | vendor/bin/cs2pr --graceful-warnings --colorize
+	vendor/bin/phan --allow-polyfill-parser -k tests/phan-config.php --no-progress-bar -m checkstyle | vendor/bin/cs2pr --graceful-warnings --colorize
```
juliushaertl, line 108: php-cs-fixer might actually be able to replace this with the composer scripts defined by running `composer run cs:check`

Makefile
```diff
@@ -104,23 +101,19 @@ lint-php-lint:
...
 
 lint-php-phpcs:
 	# PHP CodeSniffer
-	# TODO needs to be fixed
-	# vendor/bin/phpcs --standard=tests/phpcs.xml $(php_dirs) --report=checkstyle | vendor/bin/cs2pr --graceful-warnings --colorize
+	vendor/bin/phpcs --standard=tests/phpcs.xml $(php_dirs) --report=checkstyle | vendor/bin/cs2pr --graceful-warnings --colorize
```
juliushaertl, line 112: php-cs-fixer might actually be able to replace this with the composer scripts defined by running `composer run cs:check`

Makefile
```diff
@@ -104,23 +101,19 @@ lint-php-lint:
...
 
 lint-php-cs-fixer:
 	# PHP Coding Standards Fixer (with Nextcloud coding standards)
-	# vendor/bin/php-cs-fixer fix --dry-run --diff
-	PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer fix --dry-run --diff
+	vendor/bin/php-cs-fixer fix --dry-run --diff
```
juliushaertl, line 116: Same as above :)

appinfo/info.xml
```diff
@@ -21,7 +21,7 @@ Share your tables with users and groups within your cloud.
 Have a good time and manage whatever you want.
 
 ]]></description>
-    <version>0.2.2</version>
+    <version>0.2.3</version>
```
juliushaertl, line 24:
```suggestion
    <version>0.3.0</version>
```

css/grid.css
```diff
@@ -4,9 +4,12 @@
 }
 
 .row.first-row {
-	padding-top: 35px;
-	padding-left: 45px;
+	padding-top: 10px;
```
juliushaertl, line 7: Not urgent but would be good to generally try to move to use css variables for this.

We have a `--default-grid-baseline` (4px) in the server where any spacing should be a multiple of that one (4/8/12/16/...)

e.g.: padding: `calc(var(--default-grid-baseline) * 3);` for 12px

css/style.css
```diff
@@ -1,5 +1,5 @@
 h1 {
```
juliushaertl, line 1: This seems to apply to all styles on the page as the stylesheet is loaded through PHP. I might be better to at some point to move those general styles to the vue components and make use of scoped styling.

lib/Db/Table.php
```diff
@@ -25,6 +26,7 @@ public function jsonSerialize(): array {
 		return [
 			'id' => $this->id,
 			'title' => $this->title,
+            'emoji' => $this->emoji,
```
juliushaertl, line 29: Indented with spaces instead of tabs ;)

lib/Migration/Version000203Date20230124000000.php
```diff
@@ -0,0 +1,39 @@
...
+	 * @return null|ISchemaWrapper
+	 * @throws SchemaException
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = 'tables_tables';
+		if ($schema->hasTable($table)) {
+            $table = $schema->getTable($table);
```
juliushaertl, line 30: Spaces but `composer run cs:fix` should be able to auto fix this ;)

lib/Migration/Version000203Date20230124000000.php
```diff
@@ -0,0 +1,39 @@
...
+			$table->addColumn('emoji', Types::STRING, [
+				'notnull' => false,
+				'length' => 20
```
juliushaertl, line 33: Any specific reason for the length of 20?

lib/Service/TableService.php
```diff
@@ -111,11 +111,12 @@ public function find($id) {
 	 * @throws \OCP\DB\Exception
 	 * @throws InternalError|PermissionError
 	 */
-	public function create($title, $template): Table {
+	public function create($title, $template, $emoji): Table {
```
juliushaertl, line 114:
```suggestion
	public function create(string $title, string $template, string $emoji): Table {
```

lib/Service/TableService.php
```diff
@@ -138,7 +139,7 @@ public function create($title, $template): Table {
 	 *
 	 * @throws InternalError
 	 */
-	public function update($id, $title, $userId) {
+	public function update($id, $title, $emoji, $userId) {
```
juliushaertl, line 142:
```suggestion
	public function update(string $id, string $title, string $emoji, string $userId) {
```

package.json
```diff
@@ -54,6 +55,8 @@
 		"@nextcloud/browserslist-config": "^2.3.0",
 		"@nextcloud/eslint-config": "^8.2.0",
 		"@nextcloud/stylelint-config": "^2.3.0",
-		"@nextcloud/webpack-vue-config": "^5.4.0"
+		"@nextcloud/webpack-vue-config": "^5.4.0",
+		"node-sass": "^8.0.0",
```
juliushaertl, line 59: Should be replaces with the `sass` package due to https://sass-lang.com/blog/libsass-is-deprecated

src/store/data.js
```diff
@@ -0,0 +1,190 @@
...
+				if (res.status === 200) {
+					const row = res.data
+					const rows = state.rows
+					const index = rows.findIndex(r => r.id === row.id)
+					rows[index] = row
+					commit('setRows', [...rows])
+				} else {
+					console.debug('axios error', res)
+					return false
+				}
```
juliushaertl, line 145: Might be preference but I feel in general handling errors first and returning early if something fails is more readable.

```suggestion
				if (res.status !== 200) {
					console.debug('axios error', res)
					return false
				}
				
				const row = res.data
				const rows = state.rows
				const index = rows.findIndex(r => r.id === row.id)
				rows[index] = row
				commit('setRows', [...rows])
				return true
```

src/store/data.js
```diff
@@ -0,0 +1,190 @@
...
+			} catch (e) {
+				console.error(e)
```
juliushaertl, line 147: Since you already catch and log here you could also just throw an error instead of the console.debug in line 137

src/shared/components/ncTable/sections/CustomTable.vue
```diff
@@ -0,0 +1,167 @@
...
+				<TableHeader :columns="columns"
+					:selected-rows="selectedRows"
+					:rows="rows"
+					@create-row="$emit('create-row')"
+					@create-column="$emit('create-column')"
+					@edit-columns="$emit('edit-columns')"
+					@download-csv="data => $emit('download-csv', data)"
+					@select-all-rows="selectAllRows" />
+			</thead>
+			<tbody style="overflow-x: auto;">
```
juliushaertl, line 14: Moving to the style block would be great :)

src/shared/components/ncTable/partials/TableCellStars.vue
```diff
@@ -0,0 +1,56 @@
...
+			} else if (v && v === 3) {
+				res = starFull + starFull + starFull + starEmpty + starEmpty
+			} else if (v && v === 4) {
+				res = starFull + starFull + starFull + starFull + starEmpty
+			} else if (v && v === 5) {
+				res = starFull + starFull + starFull + starFull + starFull
+			} else {
+				res = starEmpty + starEmpty + starEmpty + starEmpty + starEmpty
+			}
+			return res
```
juliushaertl, line 44:
```suggestion
			return starFull.repeat(v) + starEmpty.repeat(5-v)
```

src/modules/navigation/sections/Navigation.vue
```diff
@@ -0,0 +1,143 @@
...
+			return this.tables.filter(table => { return table.title.toLowerCase().includes(this.filterString.toLowerCase()) })
+		},
+	},
+	mounted() {
+		subscribe('create-table', this.createTable)
+		subscribe('edit-table', tableId => { this.editTableId = tableId })
+	},
+	beforeDestroy() {
+		unsubscribe('create-table', this.createTable)
+		subscribe('edit-table', tableId => { this.editTableId = tableId })
```
juliushaertl, line 105: Shouldn't that unsubscribe?

src/modules/navigation/sections/Navigation.vue
```diff
@@ -0,0 +1,143 @@
...
+	},
+	methods: {
+		createTable() {
+			this.showModalCreateTable = true
+		},
+		closeNav() {
+			if (window.innerWidth < 960) {
```
juliushaertl, line 112: Having that defined somewhere globally would be nice instead of the magic value

juliushaertl:
One more thing found when testing:

- Text columns should probably be limited in size:
  Screenshot 2023-02-04 at 00 34 23

juliushaertl:
🎉

PR #120, opened 2023-02-05
Data source for the Analytics app

Rello:
Analytics accesses data via data sources.
This pull adds the data source logic to Tables.

Within Analytics, the data source "Tables App" can be selected. Additionally, the required columns can be restricted.

The following restrictions apply:
- when persisting data within the Analytics app (automation -> data load), only 3 columns can loaded and the last one has to be a key figure. This might change in the future
- when using realtime reporting, no column restrictions apply
- for most use cases, realtime reporting should be used to avoid double data storage
- for high-volume reporting, Analytics datasets should be used to store the data for high performance reporting and filtering

> Bildschirm­foto 2023-02-05 um 20 05 42

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -3,18 +3,31 @@
 namespace OCA\Tables\AppInfo;
 
 use OCA\Tables\Listener\UserDeletedListener;
+use OCA\Tables\Listener\AnalyticsDatasourceListener;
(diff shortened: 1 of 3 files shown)


datenangebot (approved):
Changes are okay. Some checks failed, but that's only code hygiene. I will merge and fix that in the version-feature branch.

PR #121, opened 2023-02-08
PR for integration a very basic API and Analytics integrations

datenangebot:
With this PR I provide a basic API interface ([Docs](https://github.com/nextcloud/tables/wiki/API)).
This is useful and also be used by the app integrations for the analytics app, contribution by @Rello ->many thanks!


lib/Controller/Api1Controller.php
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+
+	/**
+	 * @NoAdminRequired
+	 * @CORS
+	 * @NoCSRFRequired
+	 */
+	public function index(): DataResponse {
+		return $this->handleError(function () {
```
marcelklehr, line 39: Generally, I believe this is usually implemented as middleware, but I see you've already got this in all your controllers, so :shrug:

appinfo/routes.php
```diff
@@ -2,10 +2,22 @@
...
 		['name' => 'page#index', 'url' => '/', 'verb' => 'GET'],
 
 		['name' => 'tableTemplate#list', 'url' => '/table/templates', 'verb' => 'GET'],
 
+		// API
+		['name' => 'api1#index', 'url' => '/api/1/tables', 'verb' => 'GET'],
+
+		['name' => 'api1#data', 'url' => '/api/1/table/{tableId}', 'verb' => 'GET', 'postfix' => 'tableOnly'],
+		['name' => 'api1#data', 'url' => '/api/1/table/{tableId}/limit/{limit}', 'verb' => 'GET', 'postfix' => 'tableAndLimit'],
+		['name' => 'api1#data', 'url' => '/api/1/table/{tableId}/limit/{limit}/offset/{offset}', 'verb' => 'GET', 'postfix' => 'tableAndLimitAndOffset'],
```
marcelklehr, line 19: GET parameters will automatically be used to populate route method params, so `/api/1/table/{tableId}` is technically enough. You can add limit and offset via query params then.

src/components/NcTable/NcTable.vue
```diff
@@ -0,0 +1,34 @@
...
+		</div>
+	</div>
+</template>
+
+<script>
+import Options from './sections/Options.vue'
+import CustomTable from './sections/CustomTable.vue'
+
+export default {
+	name: 'NcTable',
```
marcelklehr, line 17: these vue files seem unrelated, no?

lib/Controller/Api1Controller.php
datenangebot: Do you mean the error handling "layer" as middleware?
I saw it somewhere else and thought it's a nice idea to handle it that way...

appinfo/routes.php
datenangebot: what do you mean? maybe the [matching subURLs](https://docs.nextcloud.com/server/latest/developer_manual/basics/routing.html#matching-suburls)?
If I just use the first line, I can not curl with limit and offset...

src/components/NcTable/NcTable.vue
datenangebot: mh, yes. They are not related to the api and analytics integration. wondering.

appinfo/routes.php
datenangebot: Please tell me what do you mean, but for now I think it is okay to do it this way and I feel free to merge... :|

marcelklehr: > If I just use the first line, I can not curl with limit and offset...

If you just used the first line you could `curl /api/1/table/{tableId}?limit={limit}&offset={offset}`

Rello:
@datenangebot 
i just found a possible issue with the datasource in a certain combination (when scheduling unattended dataloads).
What is your release plan?

appinfo/routes.php
datenangebot: 🫣 Thanks, got it. Will do so...

marcelklehr: :)

PR #130, opened 2023-02-21
Add search provider and reference provider/widget

julien-nc:
Hey there,

This should work with NC 25 and 26. A NC version check is made in Application.php and NC 26 stuff is not loaded when using 25.

### What's new:
* Very basic search provider for tables
* Very basic reference provider to resolve internal table links
* Very basic reference widget to render table links
* Discoverable/searchable reference provider to be used by the link picker. It just declares the table search provider as supported.
* `TableService::search()` method which is implemented in a very inefficient and naive way

### What more could be done:
* [ ] Adjust the search provider and the reference widget to show relevant information about the tables
* [ ] Implement a custom picker component which would be displayed instead of the generic search in the link picker. For now, the generic one using the search provider seem good enough.

### Technical details/discussion
* `Vue` and `TableReferenceWidget` are lazy loaded to make the reference script smaller and prevent useless work for the browser if no table widget is displayed.
* There could be a new route for table links like
`['name' => 'page#table', 'url' => '/t/{id}', 'verb' => 'GET'],` ('/table/{id}' is already taken by the internal API) which would return the same as `page#index`.
This would then allow to generate table links in a cleaner way with `urlGenerator::linkToRouteAbsolute()`. Just my 2 cents


lib/Reference/TableReferenceProvider.php
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function matchReference(string $referenceText): bool {
+		if ($this->userId === null) {
+			return false;
+		}
+		$start = $this->urlGenerator->getAbsoluteURL('/apps/' . Application::APP_ID);
```
juliushaertl, line 44: Might be nice to just reuse the getTableIdFromLink method here since the matching logic should be similar.

lib/Reference/TableReferenceProvider.php
```diff
@@ -0,0 +1,139 @@
...
+	 */
+	public function resolveReference(string $referenceText): ?IReference {
+		if ($this->matchReference($referenceText)) {
+			$tableId = $this->getTableIdFromLink($referenceText);
+			if ($tableId === null) {
+				// fallback to opengraph if it matches but somehow we can't resolve
+				return $this->linkReferenceProvider->resolveReference($referenceText);
+			}
+			try {
+				$table = $this->tableService->find($tableId, $this->userId)->jsonSerialize();
```
juliushaertl, line 66: Any reason to serialise here here already instead of using the entity object further down?

julien-nc: `OCA\Tables\Db\Table` does not declare the getters so I didn't use them and went for accessing the info after having serialized. No special reason. Also, I figured this would be rewritten anyway :grin:. This PR's goal is just to provide a basic implementation to be polished then.

lib/Reference/TableReferenceProvider.php
julien-nc: :+1:

datenangebot (approved):
Many thanks @julien-nc 
I red it line by line and learned a lot.
After that I refactored some methods.
Ready to merge, finals tests will be integrated into the release testings...

lib/AppInfo/Application.php
```diff
@@ -22,6 +30,21 @@ public function __construct() {
...
 		$context->registerEventListener(BeforeUserDeletedEvent::class, UserDeletedListener::class);
 		$context->registerEventListener(DatasourceEvent::class, AnalyticsDatasourceListener::class);
+		$context->registerEventListener(RenderReferenceEvent::class, TablesReferenceListener::class);
+
+		$context->registerSearchProvider(SearchTablesProvider::class);
+
+		$container = $this->getContainer();
+		/** @var IConfig $config */
+		try {
+			$config = $container->get(IConfig::class);
```
julien-nc, line 40:
```suggestion
		try {
			/** @var IConfig $config */
			$config = $container->get(IConfig::class);
```

The comment help the IDE to know which class the $config var is.

lib/AppInfo/Application.php
```diff
@@ -22,6 +30,21 @@ public function __construct() {
...
+			if (version_compare($config->getSystemValueString('version', '0.0.0'), '26.0.0', '<')) {
+				$context->registerReferenceProvider(TableReferenceProvider::class);
+			} else {
+				$context->registerReferenceProvider(SearchableTableReferenceProvider::class);
+			}
+		} catch (NotFoundExceptionInterface|ContainerExceptionInterface $e) {
```
julien-nc, line 46: If IConfig is not found by the container, Nextcloud is not working at all :grin: . But it does not hurt to catch the exception.

lib/Db/TableMapper.php
```diff
@@ -47,4 +47,50 @@ public function findAll(?string $userId = null): array {
 		}
 		return $this->findEntities($qb);
 	}
+
+	/**
+	 * @throws Exception
+	 */
+	public function search(string $term = null, ?string $userId = null, ?int $limit = null, ?int $offset = null): array {
```
julien-nc, line 54: Awesomely more performant than filtering a list of all tables in Php!

lib/Reference/ReferenceHelper.php
```diff
@@ -0,0 +1,135 @@
...
+use OC\Collaboration\Reference\LinkReferenceProvider;
+use OCA\Tables\Service\TableService;
+use OCP\Collaboration\Reference\Reference;
+use OCA\Tables\AppInfo\Application;
+use OCP\Collaboration\Reference\IReference;
+use OCP\IConfig;
+use OCP\IURLGenerator;
+use Throwable;
+
+class ReferenceHelper {
```
julien-nc, line 15: Indeed nicer to factorize this. :+1:

lib/Reference/SearchableTableReferenceProvider.php
```diff
@@ -0,0 +1,107 @@
...
+	public function getCacheKey(string $referenceId): ?string {
+		return $this->referenceHelper->getCacheKey($referenceId);
+	}
+
+	/**
+	 * @param string $userId
+	 * @return void
+	 */
+	public function invalidateUserCache(string $userId): void {
+		$this->referenceHelper->getTableIdFromLink($userId);
```
julien-nc, line 105:
```suggestion
		$this->referenceManager->invalidateCache($userId);
```
Mistake or intentional?

lib/Reference/TableReferenceProvider.php
```diff
@@ -0,0 +1,58 @@
...
+	public function getCacheKey(string $referenceId): ?string {
+		return $this->referenceHelper->getCacheKey($referenceId);
+	}
+
+	/**
+	 * @param string $userId
+	 * @return void
+	 */
+	public function invalidateUserCache(string $userId): void {
+		$this->referenceHelper->getTableIdFromLink($userId);
```
julien-nc, line 56:
```suggestion
		$this->referenceManager->invalidateCache($userId);
```

lib/Search/SearchTablesProvider.php
```diff
@@ -0,0 +1,126 @@
...
+	public function getName(): string {
+		return $this->l10n->t('Nextcloud tables');
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getOrder(string $route, array $routeParameters): int {
+		if (strpos($route, Application::APP_ID . '.') === 0) {
+			// Active app, prefer GitHub results
```
julien-nc, line 74:
```suggestion
			// Active app, prefer Tables results
```

lib/Service/TableService.php
```diff
@@ -180,6 +180,28 @@ public function find(int $id, ?string $userId = null, bool $skipTableEnhancement
...
+	/**
+	 * @param string $term
+	 * @param int $limit
+	 * @param int $offset
+	 * @param string|null $userId
+	 * @return array
+	 */
+	public function search(string $term, int $limit = 100, int $offset = 0, ?string $userId = null): array {
+		try {
+			/** @var string $userId */
```
julien-nc, line 192: Why do you redefine the type of `$userId`? It is already stated in the method signature. Also it does not match :grin:.
Plus `preCheckUserId` actually accepts a `?string` param.

lib/Service/TableService.php
```diff
@@ -180,6 +180,28 @@ public function find(int $id, ?string $userId = null, bool $skipTableEnhancement
...
+			$this->permissionsService->preCheckUserId($userId);
+			$tables = $this->mapper->search($term, $userId, $limit, $offset);
+			foreach ($tables as &$table) {
+				/** @var string $userId */
+				$this->enhanceTable($table, $userId);
```
julien-nc, line 197: Same here, IMO no need to redefine the type.
Also, `$userId` can be null there. I would add a check or adjust `enhanceTable`.
Or maybe `InternalError` is thrown in this case?

julien-nc (commented):
I can't request changes because I'm the PR author.
Still requesting some :grin:
Nice!
A few minor things and comments.

src/views/TableReferenceWidget.vue
```diff
@@ -0,0 +1,126 @@
...
+				<strong>
+					<a :href="richObject.link" target="_blank">
+						{{ richObject.emoji + ' ' + richObject.title }}
+					</a>
+				</strong>
+			</div>
+			<div class="details">
+				<NcUserBubble :user="richObject.ownership"
+					:display-name="richObject.ownerDisplayName" />&nbsp;
+				<NcCounterBubble>{{ richObject.rowsCount }}&nbsp;rows</NcCounterBubble>
```
julien-nc, line 39:
```suggestion
				<NcCounterBubble>{{ n('tables', '{nb} row', '{nb} rows', richObject.rowsCount, { nb: richObject.rowsCount}) }}</NcCounterBubble>
```

julien-nc (commented):
Forgot this one.

lib/Reference/ReferenceHelper.php
datenangebot: thx

lib/AppInfo/Application.php
datenangebot: Is that a comment on your own code? 🙃

lib/Db/TableMapper.php
datenangebot: You pushed me to use the SQL "IN" command, otherwise the pagination would not work (fine).
And that's great at all, filtering directly in the db is the most performance way...

lib/Reference/TableReferenceProvider.php
datenangebot: mistake, thats why PRs are great. Thanks you really read it.

lib/AppInfo/Application.php
julien-nc: I'm taking about the try/catch :grin:

lib/Service/TableService.php
datenangebot: You are right. That was my first attempt to fix a psalm error. But https://github.com/nextcloud/tables/blob/8a0f7ddad89202a45739be814da066f1371ea53b/lib/Service/TableService.php#L196-L197 did it. I will remove it.

lib/Service/TableService.php
datenangebot: No, `enhanceTable` needs a string (not null) and that's correct.
With the `preCheck` it is ensured that we have an string (can be '') or a error is thrown.
For the psalm check I have to tell him, that we ensured a string...

datenangebot (approved):

lib/Reference/TableReferenceProvider.php
datenangebot: Done

lib/AppInfo/Application.php
datenangebot: Ah okay, but I need it to satisfy psalm...

PR #131, opened 2023-02-23
Behat and phpunit skeleton

juliushaertl:
- tests: Add behat skeleton
- tests: Add phpunit skeleton


.github/workflows/integration.yml
```diff
@@ -0,0 +1,154 @@
...
+      - 'composer.json'
+      - 'composer.lock'
+  push:
+    branches:
+      - main
+      - master
+      - stable*
+
+env:
+  APP_NAME: deck
```
juliushaertl, line 20:
```suggestion
  APP_NAME: tables
```

datenangebot: Well...

Issue #7, opened 2022-02-19
Add API


datenangebot:
Provide access to manipulate tables, columns and rows via api

Adding doc for usage

datenangebot:
> And if possible with a permissions management in order to have access only in public reading too.

Public sharing is not yet implemented. It is on the roadmap, but is not related to this issue.

datenangebot:
> public REST API

@stefan-niedermann Thanks for your issue comment. It's exactly what I am working on.
Here you can find a preview of the API. Some parts are still missing, will take about a week further...
https://petstore.swagger.io/?url=https://raw.githubusercontent.com/nextcloud/tables/enh/api/APIv1.yaml#/

All the changes are prepared in a pull request. Feel free to checkout and comment something:
https://github.com/nextcloud/tables/pull/170

And by the way, having an mobile app for tables will be very interesting!

stefan-niedermann:
Awesome! I'll have a look at it 🚀

stefan-niedermann:
My first thought was, that `Etag`s are missing - should I open a separate issue or is this just a documentation lack? For a sane performance this is must have in my opinion 🙂

datenangebot:
That would be a new issue.

datenangebot:
> do you plan on taking row level writes through the API

Yes, I am implementing it at this time... stay tuned.

stefan-niedermann:
How will we handle API versioning?
The current API version is `1`, the currently released app version for NC 25 is `0.3.3`.
- Will we have minor API jumps or only major API jumps?
- Will the API version be depending in any sense on the app version?
- How is the API version propagated?
  - In the `capabilities` endpoint? (Deck app is doing this)
  - As header on each request? (Notes app is doing this)

stefan-niedermann:
Thanks for your explanations!

> (hopefully) never change.

Sure :smile: :wink: 

> So API v1 will be stable forever. If we need something new, not just an addition, we create API v2.

Then how will clients be able to know that they can use the "additions" if there is no minor update of the API version to v1.1?

> Sorry, but what do you mean with "API version propagated"?

Well, the API version *will* change over the time - either backwards compatible (for example by adding new endpoints without touching the existing ones) or incompatible (removing deprecated endpoints or change the way how they work).
Clients *may* support multiple API versions, but even if they do not, they will need to know which API version the current version of the app supports.

[...]

PR #170, opened 2023-03-21
enh(api): add API interface for managing tables

datenangebot:
including integration tests

Closes https://github.com/nextcloud/tables/issues/7


APIv1.yaml
```diff
@@ -0,0 +1,254 @@
+openapi: 3.0.3
+info:
```
juliushaertl, line 2: @datenangebot Did you check with @provokateurin about generating those from the controller annotations? I'd be in favour of going the automated approach directly if possible.

datenangebot: Yes, we are in contact about that. The setup for auto generating is not yet ready to run. I like to start this way getting to know the openAPI things. But a migration to annotations is planned.

tests/integration/features/api/tablesapi.feature
```diff
@@ -1,17 +1,17 @@
...
     Then user "participant1" has the following tables
       | Tutorial |
-#  Scenario: User has tables
-#    Given as user "admin"
-#    Given user "admin" logs in
-#    Then user "admin" has the following tables
-#    | Staffing |
+
+  Scenario: User creates, rename and delete a table
+    Then user "participant1" creates a table with title "my new awesome table" and optionally emoji "🤓"
```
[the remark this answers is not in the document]
datenangebot: yes, I will add that.

juliushaertl (approved):

APIv1.yaml
```diff
@@ -0,0 +1,490 @@
...
+                    description: Share node object
+                    content:
+                        application/json:
+                            schema:
+                                $ref: "#/components/schemas/Share"
+
+    /table/{tableId}/rows:
+        get:
+            summary: Get an array of rows
+            description: The first row contains the column titles. There is a default row limit by 1000 rows. If you need more (or want to limit less), you can set a limit on your own. Additionally set an offset if needed, would be good for lazy loading etc. Shipped with v0.3.0.
```
stefan-niedermann, line 357: @datenangebot Are you sure this API is present since `v0.3.0` of the app? Calling this endpoint in `v0.3.3` of the app with a valid `tableId` in the path results in `HTTP 405 Method not allowed`.

APIv1.yaml
```diff
@@ -0,0 +1,927 @@
...
+                    type: string
+        Rows:
+            type: array
+            maxItems: 1000
+            items:
+                $ref: "#/components/schemas/Row"
+        Row:
+            type: array
+            description: Values depending on the table definition.
+            items: {}
```
stefan-niedermann, line 890: I assume we could be more specific here - it's not just any object, but it has `columnId` and `value` properties like in the web app, no? Otherwise mapping the values to the actual rows would be quite hard :wink: 

image

APIv1.yaml
datenangebot: Not exactly, I will change it to 'shipped with v0.4.0'. Thx

APIv1.yaml
datenangebot: specified

lib/Service/ColumnService.php
```diff
@@ -72,51 +77,58 @@ public function find(int $id): Column {
 	/**
 	 * @noinspection PhpUndefinedMethodInspection
 	 * @noinspection DuplicatedCode
-	 * @param int $tableId
-	 * @param string $title
+	 *
 	 * @param string $userId
+	 * @param int $tableId
```
juliushaertl, line 82: PHPdoc comments could be removed here as the type is already defined in the method signature

lib/Service/ColumnService.php
```diff
@@ -140,6 +150,8 @@ public function create(
 		$item->setNumberMin($numberMin);
 		$item->setNumberMax($numberMax);
 		$item->setNumberDecimals($numberDecimals);
+		$item->setNumberPrefix($numberPrefix !== null ?: '');
+		$item->setNumberSuffix($numberSuffix !== null ?: '');
```
juliushaertl, line 154: Couldn't the parameter then just be not-nullable and have an empty string as the default?

lib/Service/ColumnService.php
```diff
@@ -159,84 +171,125 @@ public function create(
...
+		?string $selectionDefault,
+		?string $datetimeDefault
 	):Column {
 		try {
+			$item = $this->mapper->find($columnId);
+
 			// security
-			if (!$this->permissionsService->canUpdateColumnsByTableId($tableId)) {
-				throw new PermissionError('update column id = '.$id.' is not allowed.');
+			if (!$this->permissionsService->canUpdateColumnsByTableId($item->getTableId())) {
```
juliushaertl, line 228: 👍 As otherwise there might have been the risk of a security issue with getting the table id if that is just passed from the controllers.

lib/Service/TableService.php
```diff
@@ -75,12 +96,11 @@ public function findAll(?string $userId = null, bool $skipTableEnhancement = fal
 						$newSharedTables[] = $sharedTable;
 					}
 				}
-			} else {
-				$newSharedTables = [];
 			}
-		} catch (\OCP\DB\Exception $e) {
+		} catch (\OCP\DB\Exception|InternalError $e) {
 			$this->logger->error($e->getMessage());
```
juliushaertl, line 101:
```suggestion
			$this->logger->error($e->getMessage(), ['exception' => $e]);
```

This way we also have the exception trace in the log

lib/Service/TableService.php
```diff
@@ -75,12 +96,11 @@ public function findAll(?string $userId = null, bool $skipTableEnhancement = fal
...
 			throw new InternalError($e->getMessage());
+		} catch (PermissionError $e) {
```
juliushaertl, line 103: Could still be interesting to log at least in debug level

lib/Service/TableService.php
```diff
@@ -238,28 +259,29 @@ public function create(string $title, string $template, string $emoji, ?string $
...
 	 * @param string $title
-	 * @param string $emoji
+	 * @param string|null $emoji
 	 * @param string|null $userId
 	 * @return Table
 	 * @throws InternalError
 	 */
-	public function update(int $id, string $title, string $emoji, ?string $userId): Table {
-		$userId = $this->permissionsService->preCheckUserId($userId, false); // $userId is set
+	public function update(int $id, string $title, string $emoji = null, ?string $userId = null): Table {
```
juliushaertl, line 269: For consistency we could also have the title nullable so you can update only the emoji, but no blocker

lib/Service/TableService.php
```diff
@@ -272,6 +294,31 @@ public function update(int $id, string $title, string $emoji, ?string $userId):
...
+				throw new PermissionError('PermissionError: can not change table owner with table id '.$id);
+			}
+
+			/** @noinspection PhpUndefinedMethodInspection */
+			$table->setOwnership($newOwnerUserId);
+			$table = $this->mapper->update($table);
+			$this->enhanceTable($table, $userId);
+			return $table;
+		} catch (Exception $e) {
+			$this->logger->error($e->getMessage());
```
juliushaertl, line 317: Add exception to the logger call

lib/Service/TableTemplateService.php
```diff
@@ -620,33 +710,30 @@ private function createColumn(int $tableId, array $parameters): ?Column {
...
-			// tableId
-			$tableId,
-
-			// title
-			(isset($parameters['title']) && $parameters['title'] != '') ? $parameters['title'] : $this->l->t('No title given'),
-
 			// userId
 			$this->userId,
 
+			// tableId
```
juliushaertl, line 716: Once PHP 8 is minimum required version this could make use of https://stitcher.io/blog/php-8-named-arguments

juliushaertl (approved):
Few comments inside but no larger blockers from my side. Nice work.

lib/Command/RemoveTable.php
```diff
@@ -0,0 +1,71 @@
...
+	 * @return int
+	 */
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$id = $input->getArgument('ID');
+
+		try {
+			$table = $this->tableService->delete($id, '');
+			$output->writeln('Table deleted.');
+		} catch (InternalError|PermissionError|Exception $e) {
+			$output->writeln('Error occurred: '.$e->getMessage());
```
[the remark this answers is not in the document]
datenangebot: @juliushaertl Do you have an example how to print the trace?

lib/Service/ColumnService.php
datenangebot: Yes it can, but I try to avoid any kind of logic in the method signature.
The method signature is quite long, I want to keep it as simple as possible.

lib/Command/RemoveTable.php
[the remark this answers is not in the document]
datenangebot: Oh I saw the example later on. Is fine.

lib/Service/TableTemplateService.php
datenangebot: For now this tables version have to be backwards compatible to nc25 with php 7.4

lib/Command/ChangeOwnershipTable.php
```diff
@@ -0,0 +1,82 @@
...
+
+			$arr = $table->jsonSerialize();
+			unset($arr['hasShares']);
+			unset($arr['isShared']);
+			unset($arr['onSharePermissions']);
+			unset($arr['rowsCount']);
+			unset($arr['ownerDisplayName']);
+			$output->writeln(json_encode($arr, JSON_PRETTY_PRINT));
+		} catch (InternalError $e) {
+			$output->writeln('Error occurred: '.$e->getMessage());
```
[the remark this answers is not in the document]
datenangebot: done

lib/Service/TableService.php
datenangebot: done

lib/Service/TableService.php
datenangebot: done

src/modules/main/modals/EditColumns.vue
```diff
@@ -219,7 +219,7 @@ export default {
 	},
 	filters: {
 		truncate(text, length, suffix) {
-			if (text.length > length) {
+			if (text?.length > length) {
 				return text.substring(0, length) + suffix
 			} else {
```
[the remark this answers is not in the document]
datenangebot: done

lib/Service/TableService.php
datenangebot: done

lib/Service/TableService.php
datenangebot: done

PR #190, opened 2023-04-03
Enh/column type selection

### LINKED ISSUE #10: Add column type for (multi) selections

datenangebot:
Let's add a new column type for selection/multi selections

Closes #10


juliushaertl:
@datenangebot Since I see a lot of rebase commits in there I think there is still something wrong. The rebase itself should not add additional commits, so maybe you still do a merge instead of the actual rebase somehow?

PR #192, opened 2023-04-04
enh(columnType): new column type: (multi)selection

### LINKED ISSUE #10: Add column type for (multi) selections

datenangebot:
enh(columnType): complete column type 'multi-selection'
- all parts for column type 'multi-selection'
- small adjustments in context to the new column type

enh(columnType): complete column type 'selection'
- all parts for column type 'selection'
- preparations of column type 'multi selection'
- small adjustments in context to the new column type

enh(columnType): add selection as column type and provide the initial form

closes #10


src/modules/main/modals/CreateColumn.vue
```diff
@@ -203,8 +201,8 @@ export default {
 				{ id: 'number-stars', label: t('tables', 'Stars rating') },
 				{ id: 'number-progress', label: t('tables', 'Progress bar') },
 
-				// { id: 'selection', label: t('tables', 'Selection') },
-				// { id: 'selection-multi', label: t('tables', 'Multiselect') },
+				{ id: 'selection', label: t('tables', 'Selection') },
+				{ id: 'selection-multi', label: t('tables', 'Multiselect') },
```
juliushaertl, line 205: Calling @jancborchardt for the wording here. My suggestion would be "Single select" and "Multiple select" to make it more clear what the difference is. This is also how airtable for example names them https://support.airtable.com/docs/supported-field-types-in-airtable-overview

juliushaertl:
Some testing feedback:

- [ ] Default value is not presented in the row edit modal:
Screenshot 2023-04-04 at 17 44 59
Screenshot 2023-04-04 at 17 45 03
- [ ] Multiselect value is not rendered in the table view after setting (only after reload)

juliushaertl:
Found a few bugs when testing, otherwise this seems really nice already :)

src/modules/main/modals/CreateColumn.vue
jancborchardt: Yeah good point – let’s do it like that, more clear and then also people coming from Airtable will understand it easily. :)

juliushaertl (approved):

PR #215, opened 2023-04-12
chore(capabilities): add some basic capabilities to the NC API

datenangebot:
In addition to https://github.com/nextcloud/tables/pull/170

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * @copyright Copyright (c) 2023 Florian Steffens <florian.steffens@nextcloud.com>
+ *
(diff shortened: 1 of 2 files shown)


lib/Capabilities.php
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getCapabilities() {
+		return [
+			'tables' => [
+				'enabled' => $this->appManager->isEnabledForUser('tables'),
+				'installed' => $this->appManager->isInstalled('tables'),
```
stefan-niedermann, line 49: If the app isn't installed, the whole entry would not exist, so isn't this *always* `true`?

stefan-niedermann (approved):

PR #220, opened 2023-04-14
fix(reference): fix reference provider and widget rendering

datenangebot:
- return random cache key if debug mode is enabled
- load working components from vue-richtext

### DIFF
--- a/lib/Reference/ReferenceHelper.php
+++ b/lib/Reference/ReferenceHelper.php
@@ -127,9 +127,18 @@ class ReferenceHelper {
 
 	public function getCacheKey(string $referenceId): ?string {
 		if ($this->config->getSystemValue('debug')) {
-			return '-';
+			return $this->randomString(10);
 		} else {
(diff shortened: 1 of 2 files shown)


lib/Reference/ReferenceHelper.php
```diff
@@ -127,9 +127,18 @@ public function getCachePrefix(string $referenceId): string {
 
 	public function getCacheKey(string $referenceId): ?string {
 		if ($this->config->getSystemValue('debug')) {
-			return '-';
+			return $this->randomString(10);
```
juliushaertl, line 130: How about using https://www.php.net/manual/en/function.uniqid.php ?

datenangebot: yes, might be better. But this is just a workaround. @julien-nc The component itself should handle the cache usage if debug mode is enabled.

juliushaertl (approved):

PR #233, opened 2023-04-19
cypress: use random user instead of admin

snehar97:
use a random user instead of admin

### DIFF
--- a/cypress/e2e/tables.cy.js
+++ b/cypress/e2e/tables.cy.js
@@ -6,8 +6,9 @@ describe('The Home Page', () => {
 	// })
 
 	beforeEach(function() {
-		const user = new User('admin', 'admin')
-		cy.login(user)
+		cy.createRandomUser().then(user => {
+			cy.login(user)
+		})
 		cy.visit('apps/tables')
 	})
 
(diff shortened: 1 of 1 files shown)


cypress/e2e/tables.cy.js
```diff
@@ -1,8 +1,9 @@
 import { User } from '@nextcloud/cypress'
 describe('The Home Page', () => {
 	it('successfully loads', () => {
-		const user = new User('admin', 'admin')
-		cy.login(user)
+		cy.createRandomUser().then(user => {
+			cy.login(user)
+		})
```
juliushaertl, line 6: Might be nice to move this out of the actual test and rather into the beforeEach hook https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Hooks

datenangebot: should be fixed with https://github.com/nextcloud/tables/commit/8a8d56e079804260080b077513043ec7d3f9e34e

juliushaertl:
Needs a manual rebase as there are conflicts 😉

juliushaertl (approved):

Issue #236, opened 2023-04-21
Previous value of multiline text is not cleared when adding items in a row


connium:
Adding multiple rows with "add more" enabled, value of multiline text input remains.

**Steps to reproduce the behavior**

1. Create a column with type "multiline text" for a table and set a default value.
2. Click on the button to create a new row in the table.
3. Write some text into the input field.
4. Activate the "Add more" button.
5. Save the row. The modal will be cleared to allow adding another row.
6. This time the input field for the column still contains previous value.

**Expected behavior**

The input field for the column should be cleared.

**Environment**

- Nextcloud version: 25.0.5
- Tables version: 0.4.0

PR #244, opened 2023-04-25
bugfix: #236 load default value for text-long on create row with "add more" mode

datenangebot:
- add a watcher to load the value
- make get method more precise if value is '' or false or 0
- reset "add more" state after closing modal

@juliushaertl This fix should be backported to stable0.4, whats the best way to do this?


juliushaertl:
Details on how it works can be found at https://github.com/nextcloud/backportbot

juliushaertl (approved):

Issue #235, opened 2023-04-21
Default value is removed when adding items in a row


connium:
Adding multiple rows with "add more" enabled, default values get removed.

**Steps to reproduce the behavior**

1. Create a column with type "single line of text" for a table and set a default value (e.g. "Germany").
2. Click on the button to create a new row in the table. The input field for the column is prefilled with the default value "Germany".
3. Activate the "Add more" button.
4. Save the row. The modal will be cleared to allow adding another row.
5. This time the input field for the column is not prefilled but empty.

**Expected behavior**

The input field for the column should be pre-filled each time.

**Environment**

- Nextcloud version: 25.0.5
- Tables version: 0.4.0

juliushaertl:
Seems fixed on main, will close

PR #246, opened 2023-04-25
bugfix: #235 load default value for text-line on create row with "add more" mode

datenangebot:


src/shared/components/ncTable/partials/rowTypePartials/TextLineForm.vue
```diff
@@ -28,6 +28,11 @@ export default {
 	computed: {
 		localValue: {
 			get() {
+				if (this.value === null) {
+					const newValue = this.column?.textDefault ? this.column.textDefault : ''
```
max-nextcloud, line 32: I think you could use the [`??` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) - aka the nullish coalescing operator as it's officially called.

max-nextcloud (approved):

src/shared/components/ncTable/partials/rowTypePartials/TextLineForm.vue
datenangebot: yea, next time. Thanks.

PR #250, opened 2023-04-26
Enh/column type text rich

datenangebot:
This pull adds a new column type "text-rich".
The used editor component is the main text component from nextcloud overall.
On the create and edit modal is a full-screen option to edit in full mode.

The option for inserting files and images is hidden by css.

Screenshot 2023-04-26 at 12 23 18
Screenshot 2023-04-26 at 12 23 53

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -7,10 +7,15 @@ use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\IRequest;
 use OCP\Util;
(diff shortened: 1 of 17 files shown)


juliushaertl:
Performance warning seems unrelated. I don't think the event dispatching should trigger additional queries so might be an upstream server change.

roliverio (commented):
This change introduces a duality in how the "Text-Line and Text long" should be translated in other languages, from the code is not 100% clear if these are actually datatypes of merely some internal naming for column data. Please clarify so as to keep translations consistent.

Issue #248, opened 2023-04-26
Template selector is not keyboard accessible


juliushaertl:
The tiles for table templates in the new table dialog are currently not accessible by keyboard navigation.

In addition we would also need to indicate to screen readers which the currently selected element is.

juliushaertl:
Retested now and the entries are keyboard navigatable but selecting them is not possible -> follow up fix in #505

PR #252, opened 2023-04-26
design(create table modal): add focus trapping #248

datenangebot:
- add tabbable
- add bold and border on hover/focus

You can now tab through the templates on the create table modal:
Screenshot 2023-04-26 at 12 54 39

### DIFF
--- a/src/shared/components/NcTile/NcTile.vue
+++ b/src/shared/components/NcTile/NcTile.vue
@@ -1,5 +1,5 @@
 <template>
-	<div class="tile" :class="{active: localeActive}" @click="$emit('set-template')">
+	<div :tabindex="tabbable ? 0 : null" class="tile" :class="{active: localeActive}" @click="$emit('set-template')">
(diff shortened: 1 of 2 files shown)


src/shared/components/NcTile/NcTile.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="tile" :class="{active: localeActive}" @click="$emit('set-template')">
+	<div :tabindex="tabbable ? 0 : null" class="tile" :class="{active: localeActive}" @click="$emit('set-template')">
```
juliushaertl, line 2: For screen readers https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected would probably be the best option here to have a grid role on the parent element and then indicate the selected state through aria so that it is read out.

juliushaertl: Not blocking as this could be a follow up, but was also mentioned in #248

PR #263, opened 2023-05-02
Feat/import

datenangebot:
Resolves https://github.com/nextcloud/tables/issues/196

Adds a new option in the three-dot-menus to open a modal for importing data.
A API endpoint is also provided.

Screenshot 2023-05-02 at 16 44 55

Screenshot 2023-05-02 at 16 43 27

There might be an error in the CI, have to check that, it runs fine locally.


lib/Service/ImportService.php
```diff
@@ -0,0 +1,176 @@
...
+
+		try {
+			$storageHome = $this->rootFolder->getUserFolder($this->userId);
+			if ($storageHome->nodeExists($path)) {
+				// TODO is this working if a object storage is used?
+				$completePath = $storageHome->get($path)->getPath();
+				$basePath = $config->getSystemValue('datadirectory');
+				$spreadsheet = IOFactory::load($basePath.$completePath);
+				$this->loop($spreadsheet->getActiveSheet());
+			}
```
juliushaertl, line 73: Ideally we could just call `$storageHome->get($path)->fopen('r')` but from a quick look at the phpspreadsheet code base it seems that it does not support reading files from a stream. Also https://github.com/PHPOffice/PhpSpreadsheet/issues/1931 is still open.

To make this work with object storage we will need a local copy of the file that is obtained through our storage abstractions. 

Something like this should work:

```suggestion
			$userFolder = $this->rootFolder->getUserFolder($this->userId);
			if ($userFolder->nodeExists($path)) {
				$file = $storageHome->get($path);
				$tmpFileName = $file->getStorage()->getLocalFile($file->getInternalPath());
				$spreadsheet = IOFactory::load($tmpFileName);
				$this->loop($spreadsheet->getActiveSheet());
			}
```

lib/Service/ImportService.php
```diff
@@ -0,0 +1,176 @@
...
+			$error = 'No user in context, can not import data. Cancel.';
+			$this->logger->debug($error);
+			throw new InternalError($error);
+		}
+
+		$this->tableId = $tableId;
+		$this->createUnknownColumns = $createMissingColumns;
+
+		/** @var IConfig $config */
+		$config = Server::get(IConfig::class);
```
juliushaertl, line 63: Please inject through DI, but might not be needed anymore with the below change.

lib/Service/ImportService.php
```diff
@@ -0,0 +1,176 @@
...
+	/**
+	 * @throws ContainerExceptionInterface
+	 * @throws NotFoundExceptionInterface
+	 * @throws InternalError
+	 * @throws NotFoundError
+	 */
+	public function import(int $tableId, string $path, bool $createMissingColumns = true): array {
+		if (!$this->userId) {
+			$error = 'No user in context, can not import data. Cancel.';
+			$this->logger->debug($error);
```
juliushaertl, line 55: Sounds more like something to log at a more serious level

lib/Controller/ImportController.php
```diff
@@ -0,0 +1,36 @@
...
+
+class ImportController extends Controller {
+
+	private ImportService $service;
+	private string $userId;
+
+	use Errors;
+
+
+	public function __construct(IRequest     $request,
```
juliushaertl, line 19:
```suggestion
	public function __construct(IRequest $request,
```

lib/Service/ImportService.php
```diff
@@ -0,0 +1,176 @@
...
+			$data[] = [
+				'columnId' => $this->columns[$i]->getId(),
+				'value' => json_decode($this->parseValueByColumnType($cell->getValue(), $this->columns[$i])),
+			];
+		}
+		try {
+			$this->rowService->createComplete($this->tableId, $data);
+			$this->countInsertedRows++;
+		} catch (PermissionError|Exception $e) {
+			$this->logger->error('Could not create row while importing.');
```
juliushaertl, line 152: Would it make sense to be a bit more verbose about which row (number) failed?

Especially when catching generic exceptions it should most of the time also pass the exception along with the log so that the trace is written.

src/modules/navigation/modals/Import.vue
```diff
@@ -0,0 +1,296 @@
...
+		<div class="modal__content">
+			<div class="row">
+				<div class="col-4">
+					<h2>{{ t('tables', 'Import') }}</h2>
+				</div>
+			</div>
+
+			<!-- Starting -->
+			<div v-if="!loading && result === null && !waitForReload">
+				<RowFormWrapper :title="t('tables', 'File')" :description="t('tables', 'Chose a file that should be imported. Supported formats are xlsx, xls, html, xml and csv.')">
```
juliushaertl, line 12:
```suggestion
				<RowFormWrapper :title="t('tables', 'File')" :description="t('tables', 'Choose a file that should be imported. Supported formats are xlsx, xls, html, xml and csv.')">
```

lib/Service/ColumnTypes/DatetimeBusiness.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace OCA\Tables\Service\ColumnTypes;
+
+use OCA\Tables\Db\Column;
+
+class DatetimeBusiness extends SuperBusiness implements IColumnTypeBusiness {
+
+	public function parseValue(string $value, ?Column $column = null): string {
+		return json_encode($this->isValidDate($value, 'Y-m-d H:i') ? $value : '');
```
juliushaertl, line 10: If we parse the value does it make sense to also store it in a standard format?

juliushaertl: Something to think about is if/how we can handle different date formats.

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -94,6 +94,103 @@ public function cleanupUsers() {
...
+	/**
+	 * @Given file :file exists for user :user with following data
+	 *
+	 * @param string $user
+	 * @param string $file
+	 * @param TableNode|null $table
+	 */
+	public function createCsvFile(string $user, string $file, TableNode $table = null): void {
+		$this->setCurrentUser($user);
+		$url = $this->baseUrl.'remote.php/webdav/files/'.$user.$file;
```
juliushaertl, line 108:
```suggestion
		$url = $this->baseUrl.'remote.php/dav/files/'.$user.$file;
```

lib/Service/ImportService.php
datenangebot: This is the first method called on the service layer. I guess it's a good idea to check for existence of the userId.
What would you suggest?

lib/Service/ImportService.php
```diff
@@ -0,0 +1,176 @@
...
+
+		} catch (NotFoundException|NotPermittedException|NoUserException|InternalError|PermissionError $e) {
+			throw new NotFoundError('path could not be found');
+		}
+
+		return [
+			'found columns' => count($this->columns),
+			'created columns' => $this->countCreatedColumns,
+			'inserted rows' => $this->countInsertedRows,
+			'errors (see logs)' => $this->countErrors,
```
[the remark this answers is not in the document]
datenangebot: added a _count

lib/Service/ColumnTypes/DatetimeBusiness.php
datenangebot: Will add support for iso strings. And docs: https://github.com/nextcloud/tables/wiki/Developing#date--time

lib/Service/ColumnService.php
```diff
@@ -325,4 +331,42 @@ public function delete(int $id, bool $skipRowCleanup = false, ?string $userId =
...
+				$result[$i] = '';
+			}
+			// if there are no columns at all
+			if(!isset($result[$i])) {
+				$result[$i] = '';
+			}
+			// if column was not found
+			if($result[$i] === '' && $createUnknownColumns) {
+				$description = $this->l->t('This column was automatically created by the import service.');
+				$result[$i] = $this->create($userId, $tableId, 'text', 'line', $title, false, $description, null, null, null, null, null, null, null, null, null, null, null, null, null);
```
blizzz, line 366: Naming: with the method called `find…` i do not expect that anything will be modified here. Would split the logic, or change the naming,

lib/Service/ColumnTypes/NumberStarsBusiness.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace OCA\Tables\Service\ColumnTypes;
+
+use OCA\Tables\Db\Column;
+
+class NumberStarsBusiness extends SuperBusiness implements IColumnTypeBusiness {
+
+	public function parseValue(string $value, ?Column $column = null): string {
+		return json_encode(intval($value));
```
blizzz, line 10: at least in PHP 7 times, casting (`(int)$foo`) was faster then using `intval($foo)`. We changed it in server everywhere, and for all types.

blizzz (commented):
Went over PHP code only, two remarks for your consideration.

lib/Service/ColumnService.php
datenangebot: Right, added a "orCreate"...

lib/Service/ColumnTypes/NumberStarsBusiness.php
datenangebot: Good to know. Will adjust...

src/modules/navigation/modals/Import.vue
```diff
@@ -0,0 +1,297 @@
...
+					console.debug('error while importing', res)
+					showError(t('tables', 'Could not import, needed resources were not found.'))
+				} else {
+					showError(t('tables', 'Could not import data due to unknown errors.'))
+					console.debug('error while importing', res)
+				}
+			} catch (e) {
+				console.error(e)
+				return false
+			}
```
max-nextcloud, line 229: I would expect axios to err out for any `4xx` response code.
You can change this with the [`validateStatus` config option](https://axios-http.com/docs/handling_errors).
I'd personally rather use the `.then()` function of the promise axios returns.
In that case you need to be a little careful because if you chain `.then().catch()` exceptions triggered in the `then()` callback will also be caught in the `catch()` function. So for network responses i try to use `.then()` [with two callbacks](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#using_the_then_method). That way either the success or the error callback are called but one is not followed by the other.

max-nextcloud (commented):
Started looking at the vue parts. Not done yet but I'll share what i have for now.

src/modules/navigation/modals/Import.vue
max-nextcloud: https://github.com/nextcloud/text/commit/ea17f697135e1e274673b4c2523004a8d7e3def2 is an example where i started using then with two args to avoid handling errors triggered in `handleResponse`

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -1042,6 +1042,12 @@ public function sendRequestFullUrl($verb, $fullUrl, $body = null, array $headers
 			'Accept' => 'application/json',
 		]);
 
+		/*		print_r([
+					'verb' => $verb,
+					'url' => $fullUrl,
+					'option' => $options
+				]);*/
+
```
blizzz, line 1050: leftover?

lib/Service/ImportService.php
juliushaertl: If the userId is only passed in through dependency injection we can safely assume that the user exists I'd say since the server code would only inject that if a valid user session exists.

juliushaertl: However if this server might be reused for importing through occ or in a background job it might be good to check against IUserManager::get

src/modules/navigation/modals/Import.vue
datenangebot: Whats the advantage of your approach? And I also need a conditional switching for the toast messages, right?

lib/Service/ImportService.php
datenangebot: Yes, I thought about occ and empty userIds. Good point to check against UserManager. Will add that.

jancborchardt:
Very nice! Design details:

- [ ] Destructive actions like "Delete table" should always be the last item in an action menu as per design guidelines
- [ ] The "Choose a file that should be imported" sentence can be cut, format info is enough. Header could instead be adjusted to "File import"
- [ ] Typo "Columns are identifie**s** by" → identified
- [ ] The "Note" paragraph below could be an NcNoteCard component. Sidenote: We don't have a blue "Info" state for that? Cause yellow warning state seems much for this …

jancborchardt (changes requested):
Very nice! Design details:

- [x] Destructive actions like "Delete table" should always be the last item in an action menu as per design guidelines
- [ ] The "Choose a file that should be imported" sentence can be cut, format info is enough. Header could instead be adjusted to "File import"
- [ ] Typo "Columns are identifie**s** by" → identified
- [ ] The "Note" paragraph below could be an NcNoteCard component. Sidenote: We don't have a blue "Info" state for that? Cause yellow warning state seems much for this.

juliushaertl (approved):

juliushaertl:
@datenangebot Rebase would be nice instead of a merge commit to have a cleaner git history

src/modules/navigation/modals/Import.vue
datenangebot: We talked about that, filed in #281

Issue #269, opened 2023-05-04
Fix height with overflow for template tiles


datenangebot:
### Steps to reproduce

1.Create table
2.
3.

### Expected behavior

-

### Actual behavior

-

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

![Screenshot 2023-05-04 at 08 58 54](https://user-images.githubusercontent.com/55329475/236132774-c03cd16a-936b-422e-ba50-7b763cb3856f.png)

PR #272, opened 2023-05-05
fix: fix height and add overflow #269

datenangebot:
see #269


mejo- (approved):
Code changes seem sensible to me.

PR #277, opened 2023-05-10
Enh/#266 design

### LINKED ISSUE #266: Design review results
## Sorting

- [x] Sorting "Stars rating" and "Checkmark", initial click should sort best rated / checked on top

Since we will introduce a 3-dot menu (see below "Filter/Search"):
- [x] The whole table heading should be click sensitive for sorting
- [x] Also the sorting options should be in the 3-dot menu, like in GitHub. Then divider, then the operator list.

## Filter/Search

Main issue: Current way of filtering via typing @ 3 times (and knowing you have to do that) is very complex

- [x] Move @-filtering out of field, but into heading of the tables via 3-dot menu

Ideal future solution is still to have it all in the search bar, but the heading display is also nice and discoverable, so we can keep that in the future too.

## Mark cells with searched/filtered content

- [x] color: var(--color-primary-element-light) as background
- [x] Move from padding-right: 16px to padding left and right each 8px

## Show filter / search

- [x] Reduce text complexity of filter feedback, e.g. from Is greater than "@★★☆☆☆" to > ★★☆☆☆
- [x] Adjust placeholder for search input

## Further input

Header should ideally be sticky on vertical scroll as well, together with options bar. Title can be scrolled out of view. #109 

- [x] Checkbox highlight on the leftmost column is not fully circular -> check if this is limited by the component (hint: null != '')

- [x] Multiselect field list has square bullet style, round standard would be better (needs triage)

If there are not enough rows to fill out the whole viewport, the horizontal scrollbar is directly below the last entry. Expected would be that the scrollbar is always on the bottom, so the sensitive area for horizontal scroll is bigger. Related to #109

datenangebot:
Closes #266 

------------------------------------------------------------------

SCR-20230510-orkj

------------------------------------------------------------------

SCR-20230510-oqyk

------------------------------------------------------------------

SCR-20230510-orcu

------------------------------------------------------------------


juliushaertl:
Works like a charm. One small suggestion would be to put the active filters in a flex layout: 

Screenshot 2023-05-11 at 09 17 25

juliushaertl:
I think gap is still better as it handles first/last items properly out of the box. Maybe just use something like `gap: 0 8px` then to have a different row and column gap.

src/shared/components/ncTable/partials/TableHeaderColumnOptions.vue
```diff
@@ -0,0 +1,238 @@
...
+			console.debug('getOperators requested')
+			const possibleOperators = this.getPossibleOperators(this.column)
+
+			// preselect first operator, even if it's not displayed
+			if (this.operator === '') {
+				console.debug('operator is empty, try to set first option', possibleOperators)
+				// eslint-disable-next-line vue/no-side-effects-in-computed-properties
+				this.operator = possibleOperators[0]?.id ?? ''
+				console.debug('operator', this.operator)
+			}
```
max-nextcloud, line 142: I would not expect a change to a data value inside a computed.

It's probably fine as there's only few triggers to `getOperators` and setting a default should be fine almost anytime.
However I think it would be more clean to have a separate computed for example `operatorWithPreselect` that returns `this.operator || possibleOperators[0].id ?? ''` or so.

max-nextcloud (approved):
Code looks good to me. Just some nitpicking in one case.

jancborchardt:
Wow that looks really nice! Great job @datenangebot! :sparkles:  :)

PR #314, opened 2023-05-23
fix: error handling for axios requests

### LINKED ISSUE #281: axios error handling
### Steps to reproduce

-

### Expected behavior

-

### Actual behavior

-

### Tables app version

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

Some error handling are not effective because 4xx status codes are handled as error codes.

Have to check all axios request.

datenangebot:
Closes #281

### DIFF
--- a/src/modules/navigation/modals/CreateTable.vue
+++ b/src/modules/navigation/modals/CreateTable.vue
@@ -63,6 +63,7 @@ import '@nextcloud/dialogs/dist/index.css'
 import axios from '@nextcloud/axios'
 import { generateUrl } from '@nextcloud/router'
 import NcTile from '../../../shared/components/NcTile/NcTile.vue'
+import displayError from '../../../shared/utils/displayError.js'
 
 export default {
 	name: 'CreateTable',
@@ -180,17 +181,7 @@ export default {
 				const res = await axios.get(generateUrl('/apps/tables/table/templates'))
 				this.templates = res.data
 			} catch (e) {
-				const res = e.response
-				if (res.status === 401) {
-					showError(t('tables', 'Could not load templates, not authorized. Are you logged in?'))
-				} else if (res.status === 403) {
-					showError(t('tables', 'Could not load templates, no permissions.'))
-				} else if (res.status === 404) {
-					showError(t('tables', 'Could not load templates, resource not found.'))
-				} else {
-					showError(t('tables', 'Could not load templates, unknown error.'))
-				}
-				console.error(e)
+				displayError(e, t('tables', 'Could not load templates.'))
 			}
 		},
 	},
--- a/src/modules/navigation/modals/CreateTable.vue
+++ b/src/modules/navigation/modals/CreateTable.vue
@@ -58,7 +58,7 @@
 
 <script>
 import { NcModal, NcEmojiPicker, NcButton } from '@nextcloud/vue'
-import { showError, showWarning } from '@nextcloud/dialogs'
+import { showError } from '@nextcloud/dialogs'
 import '@nextcloud/dialogs/dist/index.css'
 import axios from '@nextcloud/axios'
 import { generateUrl } from '@nextcloud/router'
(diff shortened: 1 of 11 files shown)


src/modules/navigation/modals/CreateTable.vue
```diff
@@ -178,14 +178,19 @@ export default {
...
+				if (res.status === 401) {
+					showError(t('tables', 'Could not load templates, not authorized. Are you logged in?'))
+				} else if (res.status === 403) {
+					showError(t('tables', 'Could not load templates, no permissions.'))
+				} else if (res.status === 404) {
+					showError(t('tables', 'Could not load templates, resource not found.'))
+				} else {
+					showError(t('tables', 'Could not load templates, unknown error.'))
+				}
 				console.error(e)
```
max-nextcloud, line 193: This seems to be repeated quite a bit. How about introducing a helper function to handle axios errors?
```suggestion
				handleAxiosError(e, t('tables', 'Could not load templates'))
```

The helper would look somewhat like this:

```javascript
function handleAxiosError(e, message) {
	const status = e.response?.status
        showError(message + ', ' + statusMessage(status))
}

function statusMessage(status)
	if (status === 401) {
		return t('tables', 'not authorized. Are you logged in?')
	} else if (status === 403) {
		return t('tables', 'no permissions.')
	} else if (res.status === 404) {
		return t('tables', 'resource not found.')
	} else {
		return t('tables', 'unknown error.')
	}
	console.error(e)
}

**Comment by max-nextcloud** [comment]
Thanks for keeping track of this and fixing it. :heart: 

I added a refactoring proposal inline. Matter of taste i guess.
One thing you should make sure to cover is that even though errors can only come from axios they may not always have a `e.response`.

File: src/modules/navigation/modals/CreateTable.vue (hunk 1 above)
Comment (datenangebot) [reply src/modules/navigation/modals/CreateTable.vue:193]: Hi @max-nextcloud ,

thanks for the feedback. I agree with the idea to extract that to a helper function. But can you help me how to do it best? I thought putting it into a helper mixin so  I can use it wherever needed in components, but that does not work for vuex.

I guess plain js would do the trick, but unfortunately I don't know how to make use of toasts in there...?

Comment (max-nextcloud) [reply src/modules/navigation/modals/CreateTable.vue:193]: https://github.com/nextcloud/collectives/blob/0c7f87167cf460c0bdaff46a152859f0e17bd643/src/util/displayError.js#L27 might server as an example.

File: src/modules/sidebar/mixins/shareAPI.js [hunk 2]
```diff
@@ -33,50 +28,27 @@ export default {
...
 				const res = await axios.post(generateUrl('/apps/tables/share'), data)
-				if (res.status === 200) {
-					await this.$store.dispatch('setTableHasShares', { tableId: res.data.nodeId, hasSHares: true })
-				} else {
-					showWarning(t('tables', 'Sorry, something went wrong.'))
-					console.debug('axios error', res)
-					return false
-				}
-				// console.debug('new share was saved', res)
+				await this.$store.dispatch('setTableHasShares', { tableId: res.data.nodeId, hasSHares: true })
```
max-nextcloud, line 32: Any exception triggered in this line will also be handled by the `catch` block. In this case the request was successful. The share was created on the server but the update of the local state failed. However the UI will still display `Could not create share`.

But why would a dispatch fail you say...
Maybe due to a typo :wink: 
```suggestion
				await this.$store.dispatch('setTableHasShares', { tableId: res.data.nodeId, hasShares: true })
```

This typo will most likely not cause an exception to be thrown... I'd still recommend keeping the `try` block as small as possible, `throw` or `return` at the end of the `catch` block and move the update of the ui after the `catch` block to make sure errors during the ui update are not misinterpreted as request errors.

datenangebot: @max-nextcloud 
Well, okay thanks for the Typo. 🙄

To minimize the try block makes the code more readable. Great.
What do you think about the new version?

(To check all the functionality I will start writing e2e tests right now.)

src/modules/sidebar/mixins/shareAPI.js
```diff
@@ -1,82 +1,56 @@
...
-				}
-				// console.debug('new share was saved', res)
+				tableId = res.data.nodeId
 				showSuccess(t('tables', 'Saved new share with "{userName}".', { userName: res.data.receiverDisplayName }))
 			} catch (e) {
-				console.error(e)
-				showError(t('tables', 'Could not create new share'))
+				displayError(e, t('tables', 'Could not create share.'))
 			}
+			await this.$store.dispatch('setTableHasShares', { tableId, hasShares: true })
```
max-nextcloud, line 38: Cool... exceptions from the dispatch will not be caught anymore.

But the dispatch will happen even when network exceptions are caught. It will not work because tableId will be null - but I think it would be better to return from the catch block so the dispatch does not cause another error.

max-nextcloud: Hehe... I'm just crazy lucky. :wink: 
This is the only case i looked at. :dart:

max-nextcloud (approved):
Looks good!

PR #315, opened 2023-05-23
fix(Options): prevents buttons to get cut off

### LINKED ISSUE #289: reduce search field width
### Steps to reproduce

-

### Expected behavior

reduce input field width on smaller screens

### Actual behavior

-

### Tables app version

0.5.0

### Browser

_No response_

### Client operating system

_No response_

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

SCR-20230513-gyxf

Hephi2:
Closes #289 

When selecting a row in tables, the button can be cut with a specific screen width. This is solved by this fix.

The design should be reviewed. See #289

### DIFF
--- a/src/shared/components/ncTable/partials/SearchForm.vue
+++ b/src/shared/components/ncTable/partials/SearchForm.vue
@@ -60,7 +60,6 @@ export default {
 <style scoped>
 
 :deep(.input-field input) {
-	width: 30vw !important;
 	border-color: var(--color-primary-element);
 }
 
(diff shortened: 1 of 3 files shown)


Hephi2:
New Design:
image
Needs to be reviewed @datenangebot-

Hephi2:
@datenangebot That is not universal solution but rather a hardcoded fix, since it might not work for languages with longer translations. But I can't think of a better solution right now so I think we can do it like that for now.

datenangebot (approved):
👍

Issue #299, opened 2023-05-17
New text-rich is missing on NC25


datenangebot:
### Steps to reproduce

create new column

### Expected behavior

-

### Actual behavior

<img width="897" alt="SCR-20230517-hxol" src="https://github.com/nextcloud/tables/assets/55329475/85ab6270-08e6-47dc-8450-a8f8222480ae">

### Tables app version

0.5.0

### Browser

FF

### Client operating system

Mac

### Web server

None

### PHP engine version

None

### Database

None

juliushaertl:
The text editor API is only available with 26

datenangebot:
> The text editor API is only available with 26

Really, I thought it was shipped with 25. :(

-> NC25 should be fully supported with this app, so have to use own tiptap instance here (make use of 'text-long' components)

datenangebot:
@juliushaertl Is it possible to backport the new text-editor API to NC25 with reasonable effort?

Otherwise I suggest to use the "old" text-long columns on NC25.

PR #317, opened 2023-05-24
fix(capabilities): show the correct version for text-long or text-ric…

### LINKED ISSUE #309: Capabilities should show column types by nc-version
### Steps to reproduce

-

### Expected behavior

-

### Actual behavior

-

### Tables app version

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

https://github.com/nextcloud/tables/blob/main/lib/Capabilities.php#LL54C6-L55C19

datenangebot:
…h depending on the nextcloud system (in addition to #299)

closes #309 

Did a test with postman that worked for me, don't need a test case for it.

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -25,6 +25,11 @@ namespace OCA\Tables;
 
 use OCP\App\IAppManager;
 use OCP\Capabilities\ICapability;
+use OCP\IConfig;
+use OCP\Server;
+use Psr\Container\ContainerExceptionInterface;
+use Psr\Container\NotFoundExceptionInterface;
+use Psr\Log\LoggerInterface;
 
 /**
  * Class Capabilities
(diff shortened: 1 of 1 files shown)


lib/Capabilities.php
```diff
@@ -33,15 +38,28 @@
...
 
 	/**
 	 * @inheritDoc
 	 */
-	public function getCapabilities() {
+	public function getCapabilities(): array {
+		$textColumnVariant = 'text-rich';
+		/** @var IConfig $config */
+		try {
+			$config = Server::get(IConfig::class);
```
blizzz, line 55: why don't you inject it in the constructor?

datenangebot: Well, because I copied the code and forgot it. 😱
Will adjust it.

lib/Capabilities.php
```diff
@@ -33,15 +38,28 @@
...
 	 * @inheritDoc
 	 */
-	public function getCapabilities() {
+	public function getCapabilities(): array {
+		$textColumnVariant = 'text-rich';
+		try {
+			if (version_compare($this->config->getSystemValueString('version', '0.0.0'), '26.0.0', '<')) {
+				$textColumnVariant = 'text-long';
+			}
+		} catch (NotFoundExceptionInterface|ContainerExceptionInterface $e) {
```
blizzz, line 59: this try-catch is also not necessary anymore as the exceptions would pop up only from autoloading.

datenangebot: I see 👍

PR #322, opened 2023-05-25
feat(Filtering): Enables to filter for empty cells

### LINKED ISSUE #302: Add filter operator "is empty"

Yes, filter for "is equal" '' filters fuzzy.

Would like to filter for column without a value set.

_No response_

### Additional context

_No response_

Hephi2:
Closes #302 

Enhancements:
- Enables filtering every column where it does make sense for it to be empty.
- After a filter is set, only other filters can be chosen that can be combined with the old filters in a meaningful way (constraint for usability)

### DIFF
--- a/src/shared/components/ncTable/mixins/columnsTypes/datetimeDateMixin.js
+++ b/src/shared/components/ncTable/mixins/columnsTypes/datetimeDateMixin.js
@@ -55,6 +55,10 @@ export default {
 				cell.filterFound = true
 				return true
 			}
+			if (filter.operator === 'is-empty' && !valueDate) {
+				cell.filterFound = true
+				return true
+			}
 			return false
 		},
 
--- a/src/shared/components/ncTable/mixins/columnsTypes/datetimeMixin.js
+++ b/src/shared/components/ncTable/mixins/columnsTypes/datetimeMixin.js
@@ -55,6 +55,10 @@ export default {
 				cell.filterFound = true
 				return true
 			}
+			if (filter.operator === 'is-empty' && !valueDate) {
+				cell.filterFound = true
+				return true
+			}
 			return false
 		},
 
--- a/src/shared/components/ncTable/mixins/columnsTypes/datetimeTimeMixin.js
+++ b/src/shared/components/ncTable/mixins/columnsTypes/datetimeTimeMixin.js
@@ -55,6 +55,10 @@ export default {
 				cell.filterFound = true
 				return true
 			}
+			if (filter.operator === 'is-empty' && !valueTime) {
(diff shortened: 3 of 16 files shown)


Hephi2:
> 1. There was a logic that avoids to show up operators if there is only one operator possible. That's gone :(

I can revert that, but couldn't it be confusing for users if they don't get a hint as to which rule is being used for filtering?

> 2. You added some logic to avoid filter-collisions. That's an interesting idea. But not that easy to implement:
>    Why can't we filter for `greater than` in the example below? 

The idea was to eliminate all filters that don't make sense. If you already filter for values > x, it does not make sense to also filter for values > y. One of those filters would be redundant. 

> And if a column is filtered the user can not get an idea what filter possibilities are served, even if the combination might be senseless. 🤷🏻‍♂️ Another problem I see is, that you have to adjust all filter-configurations if you add a new one. That's not a good practice. 🤨

That is true. But isn't it the same with the `goodFor` elements if you add a new column datatype? 
Since this "filter incompatibilty" (as well as the goodFor relationship) is bijective and an n-m-relationship, it might make sense to outsource those into a different place in a form a list. What do you think?
I could also could have a look into the `class` and `enum` functionalities of js (right now I only know the from different languages), and refactor the file `searchAndFilterMixin.js` which could make the work with those different filter types easier. Whould that make sense?

> Another idea would be to allow all filter-combinations and have a check if filters exclude each other and mark them as error or whatever?

In my studies I learned that you shoud prefer constraints to error messages, but here it might make sense if the users otherwise would not understand, why they can't add a specific filter. On the other hand in a real-world scenario it should not happen that someone needs to filter, e.g., for values >5 and >8 at the same time. What do you think?

Hephi2:
> It's not, because the only operator will be the expected behavior, (search for words in text, equal for checkboxes, etc.). I think it's not a good way to show an option where is actually no option to chose. We just could think about a hint for the users how the filter will operate. But I am not sure if we need that.

I see. Then I'll for now just revert and we'll see if a hint is necessary.
 

> Yes, I agree, but we need more flexibility here. It should be possible to filter for value > x AND value < y.

That should be possible with the current commit. I think I only eliminated the combinations of filters that really don't make any sense.

> Good point, yes and no. Yes you have to define all the allowed operators. But no, you don't have to adjust other types, you stay in the logic for your added item. And it's like a whitelisting process, much better than blacklisting.

👍
 
> Btw: just an idea came to me, we could not take the unwandet operators out, we could deactivate them. So the users see whats possible and avoid unwanted filter-combinations...?

That's a good idea. I'll try to find a solution in a cleaner style with this feature.

PR #333, opened 2023-05-29
Fix : Translation of table row creation button's label

Jerome-Herbinet:
Call "tables" resource (Transifex) instead of "table".


max-nextcloud (approved):
Thanks for looking into this and the fix @Jerome-Herbinet !

PR #338, opened 2023-05-31
refactor: OO approach for better handling of columns and more

### LINKED ISSUE #337: Refactor code for better operability and maintainability

The current codebase often uses numerous similar hardcoded objects, e.g., for columns, operators/filters and magicfields, imported as mixins into vue files to be accessible. This resticts the maintainability and extensibility and quickly leads to errors.

Use an object oriented approach for columns and something like enums for operators and magic fields to bind the respective functionalities to the objects that are to apply them. Abstract common functionalities to improve maintainability. Prepare for extensibility and possible future use as a plug-in.

_No response_

### Additional context

_No response_

Hephi2:
Closes #337


juliushaertl:
Haven't had a full review yet, but looks like a great refactoring :)

Hephi2:
Sure, I'll do that

src/modules/main/modals/CreateRow.vue
```diff
@@ -124,6 +89,24 @@ export default {
...
+			case ColumnTypes.Number: return 'NumberForm'
+			case ColumnTypes.NumberStars: return 'NumberStarsForm'
+			case ColumnTypes.NumberProgress: return 'NumberProgressForm'
+			case ColumnTypes.Selection: return 'SelectionForm'
+			case ColumnTypes.SelectionMulti: return 'SelectionMultiForm'
+			case ColumnTypes.SelectionCheck: return 'SelectionCheckForm'
+			case ColumnTypes.Datetime: return 'DatetimeForm'
+			case ColumnTypes.DatetimeDate: return 'DatetimeDateForm'
+			case ColumnTypes.DatetimeTime: return 'DatetimeTimeForm'
+			default: throw Error('No form exists for the column type ' + column.type)
```
datenangebot, line 107: Would it be possible to use some magic here? Advantage: no need to adjust if we get a new column type.

pseudo code
```
const type = column.substr(0, -4)
if (classexists(type)) {
    return type
} else {
    // TODO
}
```

src/modules/main/modals/EditRow.vue
```diff
@@ -143,6 +108,24 @@ export default {
...
+			case ColumnTypes.Number: return 'NumberForm'
+			case ColumnTypes.NumberStars: return 'NumberStarsForm'
+			case ColumnTypes.NumberProgress: return 'NumberProgressForm'
+			case ColumnTypes.Selection: return 'SelectionForm'
+			case ColumnTypes.SelectionMulti: return 'SelectionMultiForm'
+			case ColumnTypes.SelectionCheck: return 'SelectionCheckForm'
+			case ColumnTypes.Datetime: return 'DatetimeForm'
+			case ColumnTypes.DatetimeDate: return 'DatetimeDateForm'
+			case ColumnTypes.DatetimeTime: return 'DatetimeTimeForm'
+			default: throw Error('No form exists for the column type ' + column.type)
```
datenangebot, line 126: Same as above?!

src/shared/components/ncTable/mixins/columnClass.js
datenangebot, line 1: Awesome

src/shared/components/ncTable/mixins/columnParser.js
```diff
@@ -0,0 +1,33 @@
...
+	case 'number': return new NumberColumn(col)
+	case 'number-stars': return new NumberStarsColumn(col)
+	case 'number-progress': return new NumberProgressColumn(col)
+	case 'selection': return new SelectionColumn(col)
+	case 'selection-multi': return new SelectionMutliColumn(col)
+	case 'selection-check': return new SelectionCheckColumn(col)
+	case 'datetime': return new DatetimeColumn(col)
+	case 'datetime-date': return new DatetimeDateColumn(col)
+	case 'datetime-time': return new DatetimeTimeColumn(col)
+	default: throw Error(col.type + '-' + col.subtype + ' is not a valid column type!')
```
datenangebot, line 31: Magic possible?

src/shared/components/ncTable/mixins/exportTableMixin.js
```diff
@@ -32,11 +29,7 @@ export default {
 			this.$papa.download(csv, moment().format('YY-MM-DD_HH-mm') + '_' + tableTitle)
 		},
 		getValueByColumnType(set, column) {
-			const methodName = 'getValueStringFor' + this.ucfirst(column.type) + this.ucfirst(column.subtype) || ''
-			if (this[methodName] instanceof Function) {
-				return this[methodName](set, column)
-			}
-			return set ? set.value : ''
+			return column.getValueString(set)
```
datenangebot, line 32: Can we replace this directly in line 20?

`rowData[column.title] = set ? column.getValueString(set) : ''`

datenangebot (changes requested):
Wow, what a great refactoring. Many thanks! 👏

- [ ] I put some comments below, maybe we can simplify some things. 🙇🏻‍♂️

- [ ] What do you think about a dev-documentation for the wiki with the topic about `How to create or adjust a (new) column type`. This could give us a documentation at which points we can/have to adjust the business logic for a column and on the other side where we find the framework that renders and make some magic FE stuff for columns.

Btw: It's so great to see that the CI is green with all that changes. 🙄

src/shared/components/ncTable/mixins/columnParser.js
Hephi2: I could not find a good working solution. The only thing that worked was loading the corresponding module dynamically:

```
const column = snakeToCamel(columnType) + 'Column'
const importedModule = await import(`./columnsTypes/${snakeToCamel(columnType, false)}.js`)
const columnClass = importedModule.default

if (columnClass) {
return new columnClass(col)
} else {
throw Error('Column for type ' + columnType + ' does not exist')
}
```

However this means, that the method would be async and that would make the table loading process more complex. I don't think that would be a good solution. Maybe you have a better idea or somehow can get it work with the static imports. I somehow cannot access them dynamically.

datenangebot: Async is not what we need here I would say. What was about the eval thing?
Maybe @juliushaertl can help here?

juliushaertl: You could probably have a default export in /columnsTypes/index.js that exports all type classes and then do something like this:

```js
export default {
	// list all type classes here
}
```

```js
import Types from './columnTypes/index.js

// ...

const column = snakeToCamel(columnType) + 'Column'
const columnClass = Types[column]

if (columnClass) {
return new columnClass(col)
} else {
throw Error('Column for type ' + columnType + ' does not exist')
}
```

Not tested though. However I'd also think a mapping function for type to class would be fine

Hephi2: Yeah. I thought about that as well.
But the idea was that when you create a new column type, you have to adjust as little code as possible. Here you would still have to add the class to the list again.
@datenangebot should I just leave it or use Julius' solution then?

datenangebot: @Hephi2 I would say we shouldn't investigate too much time here as this could be a great refactoring later on.
Just choose your favorite code and lets merge.

But from a general perspective, there has to be a smooth solution for that... 🙃

Hephi2: Okay, then let's leave it as it is.

datenangebot (approved):

PR #339, opened 2023-06-01
fix(NcMultiselect): Replaces NcMultiselects by NcSelects

### LINKED ISSUE #286: Move from NcMultiselect to NcSelect
NcMultiselect has lots of problems in regards to accessibility so at some point all usages should be switched to the new NcSelect component.

Hephi2:
Closes #286

### DIFF
--- a/cypress/e2e/tables-share.cy.js
+++ b/cypress/e2e/tables-share.cy.js
@@ -30,7 +30,7 @@ describe('Manage a table', () => {
 		cy.get('.NcTable table tr th').last().find('button').click({ force: true })
 		cy.get('.v-popper__popper.v-popper--theme-dropdown.action-item__popper.v-popper__popper--shown').contains('Share').click({ force: true })
 		cy.get('.sharing input').type(localUser2.userId)
-		cy.get('ul.multiselect__content li').first().click()
+		cy.wait(1000).get('.sharing input').type('{enter}')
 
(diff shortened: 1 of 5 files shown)


cypress/e2e/tables-share.cy.js
```diff
@@ -30,7 +30,7 @@ describe('Manage a table', () => {
 		cy.get('.NcTable table tr th').last().find('button').click({ force: true })
 		cy.get('.v-popper__popper.v-popper--theme-dropdown.action-item__popper.v-popper__popper--shown').contains('Share').click({ force: true })
 		cy.get('.sharing input').type(localUser2.userId)
-		cy.get('ul.multiselect__content li').first().click()
+		cy.wait(1000).get('.sharing input').type('{enter}')
```
juliushaertl, line 33: Maybe we can get the list entry before pressing enter to avoid timing issues. cy.wait is usually a bad idea and makes tests less stable

juliushaertl (commented):
Nice, just a small comment on the test

cypress/e2e/tables-share.cy.js
datenangebot: @juliushaertl Any hints or can I merge here?

juliushaertl: I'd say merging is fine, if this turns out to make tests flaky we can investigate deeper.

PR #347, opened 2023-06-09
fix(numbers input): respect the number of decimals configured

### LINKED ISSUE #285: Number column: `Decimals` not respected
### Steps to reproduce

1. Add column of type `Number`, set `Decimal` to `4`
2. Add row, type `12345.12345`
3. Row is added without any validation

### Expected behavior

Validation error?

### Actual behavior

Number is added and returned by the REST API with all 5 numbers before and all 5 numbers after the `.`

### Tables app version

0.5.0

### Browser

Firefox

### Client operating system

Linux

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

_No response_

datenangebot:
- add the step size related to the numbers of decimals
- round the input value for the given span

closes #285

### DIFF
--- a/src/shared/components/ncTable/partials/rowTypePartials/NumberForm.vue
+++ b/src/shared/components/ncTable/partials/rowTypePartials/NumberForm.vue
@@ -7,7 +7,7 @@
 			type="number"
 			:min="column.numberMin"
 			:max="column.numberMax"
-			:step="column.numberDecimals === 0 ? '' : 'any'">
+			:step="getStep">
 		<div v-if="column.numberSuffix" class="suffix">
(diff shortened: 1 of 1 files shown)


stefan-niedermann (commented):
:+1: There should be a migration for data that has been added with previous tables versions to ensure that the data format is actually valid :)

juliushaertl:
I'd agree that a migration would be nice (or) a conversion on the backend side to only return valid number of decimals for existing data.

juliushaertl (approved):
Looks good, I'd say we can file a follow up ticket for the migration of existing data.

PR #350, opened 2023-06-09
tests: adjust

datenangebot:


juliushaertl:
Maybe only do that for the ones that cause trouble to still take advantage of the self hosted runners for the rest?

PR #355, opened 2023-06-12
enh(commands): occ cleanup row data sets

### LINKED ISSUE #200: command to remove data where the column was deleted
It seems that the data from rows were not deleted if a column was deleted.
Idea is to have a command to clean up that.

datenangebot:
- add occ command to clean up data sets from rows at db level

closes #200 

It might happen (route cause analysis runs in parallel) that the rows hold data from columns that are already deleted.
With this command is it possible to clean up the rows data. Use `occ tables:clean -d` for a dry run and `occ tables:clean -e` to print out only the errors.

### Run `occ tables:clean`, should be all "green"
SCR-20230612-ishe

### Manipulate the db
- delete the column from `oc_tables_columns` with the id 1 (or whatever).

### Run `occ tables:clean -d`, should look like this
 
SCR-20230612-iska

### Run `occ tables:clean`, to clean up the data

SCR-20230612-ismg

### Run `occ tables:clean -d` again, should be good again

SCR-20230612-isnw

### DIFF
--- a/lib/Command/Clean.php
+++ b/lib/Command/Clean.php
@@ -1,0 +1,221 @@
+<?php
+/**
+ * @copyright Copyright (c) 2023 Florian Steffens <flost-dev@mailbox.org>
+ *
+ * @author Florian Steffens <flost-dev@mailbox.org>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Tables\Command;
+
+use OCA\Tables\Db\Row;
+use OCA\Tables\Db\RowMapper;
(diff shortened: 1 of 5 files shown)


lib/Command/Clean.php
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	/**
+	 * @param InputInterface $input
+	 * @param OutputInterface $output
+	 * @return int
+	 */
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$this->dry = !!$input->getOption('dry-run');
+		$this->showOnlyErrors = !!$input->getOption('only-errors');
```
juliushaertl, line 97: Would be nice to rather use the verbosity levels here that are provided by symfony https://symfony.com/doc/current/console/verbosity.html

lib/Command/Clean.php
```diff
@@ -0,0 +1,221 @@
...
+			$this->print("");
+			$this->print("columnId: " . $date->columnId . " -> " . substr($date->value, 0, $this->truncateLength) . $suffix, self::PRINT_LEVEL_INFO);
+
+			try {
+				$this->columnService->find($date->columnId, '');
+				if(!$this->showOnlyErrors) {
+					$this->print("column found", self::PRINT_LEVEL_SUCCESS);
+				}
+			} catch (InternalError $e) {
+				$this->print("😱️ internal error while looking for column", self::PRINT_LEVEL_ERROR);
```
juliushaertl, line 160: Are those logged elsewhere, otherwise it might be nice to still put the exception into the log for further details.

juliushaertl (changes requested):
Just smaller comments, otherwise good from my side 👍

lib/Command/Clean.php
datenangebot: Did not know that, great. Will use -v from there.

lib/Command/Clean.php
datenangebot: Yes, but only for "real" errors.

juliushaertl (approved):

PR #358, opened 2023-06-12
chore(CI): shrink test matrix to reduce CI load

datenangebot:


.github/workflows/integration.yml
```diff
@@ -26,14 +26,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.4', '8.0', '8.1', '8.2']
+        php-versions: ['7.4', '8.2']
         databases: ['mysql', 'pgsql', 'sqlite']
-        server-versions: ['stable25', 'stable26', 'stable27', 'master']
```
marcelklehr, line 31: In case you still want to cover all of these (which makes sense if you want to support them), you can use an `include:` rule to have e.g. one run per php version, one run per database and one run per server version.

.github/workflows/cypress.yml
```diff
@@ -24,17 +24,13 @@ jobs:
...
-          - php-versions: 7.4
-            server-versions: stable26
-          - php-versions: 7.4
-            server-versions: stable27
-          - php-versions: 7.4
-            server-versions: master
+        databases: [ 'mysql' ]
+        server-versions: [ 'stable25', 'master' ]
+        include:
+          - php-versions: 8.1
```
marcelklehr, line 30: Last time I tried something like this it didn't work

.github/workflows/integration.yml
```diff
@@ -26,18 +26,24 @@ jobs:
...
-          - php-versions: 7.4
-            server-versions: stable26
-          - php-versions: 7.4
-            server-versions: stable27
-          - php-versions: 7.4
-            server-versions: master
-          - php-versions: 8.2
-            server-versions: stable25
+        include:
+          - php-versions: '8.2'
```
marcelklehr, line 31: I think you want to move this up into the matrix, probably?

.github/workflows/phpunit-mysql.yml
```diff
@@ -38,8 +38,11 @@ jobs:
 
     strategy:
       matrix:
-        php-versions: ['8.0', '8.1', '8.2']
-        server-versions: ['master']
+        server-versions: ['stable25', 'master']
+        include:
+          - php-versions: 8.1
```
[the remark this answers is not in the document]
datenangebot: @marcelklehr According to the last sentence here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude

> Note that the original matrix values will not be overwritten, but added matrix values can be overwritten.

That's why I added it as additional values and not directly into the matrix and tests seems to run fine?!

.github/workflows/integration.yml
```diff
@@ -26,18 +26,24 @@ jobs:
...
+            databases: 'mysql'
+          - server-versions: 'stable27'
+            php-versions: '8.1'
+            databases: 'mysql'
+          - server-versions: 'master'
+            databases: 'sqlite'
+          - server-versions: 'master'
+            databases: 'mysql'
+          - server-versions: 'master'
+            databases: 'pgsql'
```
marcelklehr, line 46: For the current commit github only runs 4 workflows out of all of these here

.github/workflows/phpunit-mysql.yml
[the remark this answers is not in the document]
marcelklehr: So, for me what you wrote down here seems intended to run 

- php8.1 stable25
- php8.1 master
- php74 stable25

But it actually only runs 

- php8.1 master
- php74 stable25

> tests seems to run fine?!

If this is what you want, don't mind me.

marcelklehr (approved):

PR #375, opened 2023-06-21
Enh/#234 extend link column type

### LINKED ISSUE #234: Reference ressources
This issue tries to collect all ideas and approaches to get a structured idea of what is possible and hopefully a result to go for.
We have to distinguish if we are talking about
- linking data from other tables (within the tables app)
- **linking to other resources, managed by nextcloud but not by the tables app**
- linking to external resources (links)

This issue takes the second option into account.

Related issue so far:
- #217 
- #74 

Delimitation:
- #180 
- #172 
- external links

I would like to have an generic approach to link to references from the whole nextcloud system.

What data do we need for this:
- identifier (uri, app name + id, ...)
- searchable content (otherwise we can not search and filter; eg title, name, ...)

Technical approaches:
- A: Pull, the tables app asks for data and implement the logic for itself
- B: Push, apps and other parts can register as an provider and implement an interface (like search provider, reference picker)
- C: Use (at least one) of the existing data provider structures (eg. maybe we can use the search providers and logic)

Thoughts for scenario C
* (+) not one more provider logic to implement for all other maintainers
* (+) can be used out of the box
* (+) extends automatically driven by the search context
* (+) can be extended if something is missing
* (+) have all the necessary data (have to check that)
* (-) new needed provider will always accessible via the search

So far, to discuss

datenangebot:
closes #234 

This PR adds the feature to use all data that can be found by a search. This way we can link to any resources within Nextcloud that provides a search-provider.

| Screenshots |
| --- |
| SCR-20230621-muiz |
| SCR-20230621-munz |
| SCR-20230621-muqg |

**Todo**

- [x] Bug on open modal (create row), if link is first input field
SCR-20230621-mmyo

- [x] Placeholder text on first startup, not always "Url"
- [x] Writing tests
- [x] Smoke tests with more data provider
     - [x] Calendar
     - [x] Deck
     - [x] Contact
     - [x] ...
 - [x] Writing docs in wiki for this column type
 - [x] Import compatibility
 - [x] Export compatibility
 - [x] Search compatibility
 - [x] filtering compatibility
 - [x] API compatibility


cypress/e2e/column-text-link.cy.js
```diff
@@ -0,0 +1,79 @@
...
+		cy.get('.button-vue__text').contains('Create column').click({ force: true })
+		cy.get('.modal-container').get('input[placeholder*="Enter a column title"]').clear().type('Test plain url')
+		cy.get('.columnTypeSelection .vs__open-indicator').click({ force: true })
+		cy.get('.multiSelectOptionLabel').contains('Link').click({ force: true })
+		// deactivate unwanted provider
+		cy.get('.typeSelection span label').contains('Files').click()
+		cy.get('.typeSelection span label').contains('Contacts').click()
+		cy.get('.modal-container button').contains('Save').click()
+
+		cy.wait(10).get('.toastify.toast-success').should('be.visible')
```
juliushaertl, line 38: Maybe intercept the request and wait for that instead of the seconds ;)

datenangebot: How would that look like? Do you have an example?

cypress/e2e/column-text-link.cy.js
```diff
@@ -0,0 +1,79 @@
...
+	beforeEach(function() {
+		cy.login(localUser)
+		cy.visit('apps/tables')
+	})
+
+	it('Create', () => {
+		cy.contains('.app-menu-entry--label', 'Tables').click()
+		cy.contains('button', 'Create new table').click()
+		cy.get('.tile').contains('Custom table').click({ force: true })
+		cy.get('.modal__content input[type="text"]').clear().type('Test text-link')
```
juliushaertl, line 20: It might be nice to move common selectors or logic (e.g. to create a table into helper functions)

cypress/e2e/column-text-link.cy.js
juliushaertl: https://docs.cypress.io/api/commands/wait#Wait-for-a-specific-request-to-respond

Basically:
- intercept the request, give it a name
- perform the action like clicking save
- wait for the intercepted alias name

datenangebot: Well, it's just about 10 millisenconds. And we need them to make sure the toast is loaded?!

cypress/e2e/column-text-link.cy.js
datenangebot: We moved some basic logic and registered them globally.

cypress/e2e/column-text-link.cy.js
```diff
@@ -0,0 +1,69 @@
...
+			cy.uploadFile('NC_server_test.pdf', 'application/pdf', '/NC_server_test.pdf')
+		})
+	})
+
+	beforeEach(function() {
+		cy.login(localUser)
+		cy.visit('apps/tables')
+	})
+
+	it('Create', () => {
```
juliushaertl, line 19: Something I forgot during the call to mention, but non blocking from my perspective. Tests should not depend on each other so that in theory you could run them in random order or independently and makes them retryable. In this spec file every test creates the requirements for the next ones. This also has the benefit that you can run them separately with .only for debugging

juliushaertl (approved):

PR #385, opened 2023-06-29
Updating cypress.yml workflow from template

nextcloud-command:
Automated update of the cypress.yml workflow from https://github.com/nextcloud/.github


juliushaertl:
Let me file an issue as it would require some more work on restructuring how we run the cypress tests

PR #435, opened 2023-07-31
Enh/views menu and views modal

datenangebot:


lib/Controller/ColumnController.php
```diff
@@ -40,6 +40,7 @@ public function index(int $tableId, int $viewId): DataResponse {
 
 	/**
 	 * @NoAdminRequired
+	 * @NoCSRFRequired
```
juliushaertl, line 43: Why would this be needed?

lib/Controller/TableController.php
```diff
@@ -33,6 +33,7 @@ public function __construct(
 
 	/**
 	 * @NoAdminRequired
+	 * @NoCSRFRequired
```
juliushaertl, line 36: Why would this be needed?

lib/Controller/ColumnController.php
datenangebot: not needed

lib/Controller/TableController.php
datenangebot: not needed

PR #446, opened 2023-08-08
feat: more gentle icon

skjnldsv:
I find the current icon a bit out of our icon style.
Using material design icon, I suggest we use this :)

| Before | After |
|:---------:|:------:|
|||

Quick PR, feel free to discard :wink:

### DIFF
--- a/img/app-dark.svg
+++ b/img/app-dark.svg
@@ -1,12 +1,3 @@
-<svg fill="#000" version="1.0" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 415 415" style="enable-background:new 415 415 415 415;">
-	<rect x="150" y="115" width="120" height="60" />
-	<rect x="150" y="315" width="120" height="60" />
-	<rect x="300" y="115" width="120" height="60" />
-	<rect x="150" y="215" width="120" height="60" />
-	<rect x="0" y="215" width="120" height="60" />
-	<rect x="300" y="315" width="120" height="60" />
-	<rect x="300" y="215" width="120" height="60" />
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Looks like a waffle

jancborchardt (commented):
Agree that the current icon can use improvement – the suggestion looks a bit too much like a calculator though @skjnldsv?

What do you think about the Material Design "table" icon: https://pictogrammers.com/library/mdi/icon/table/
I would approve :)

skjnldsv:
Done! See updated first comment @jancborchardt

juliushaertl:
>  But it looks a bit like a window or a positioning indicator for windows on a desktop...

Which would somehow also fit the larger no code "app" picture. The apps box looks a bit too generic in my oppinion, the table one for me also brings some association with structured data, which I'd say is a better fit - and as you mentioned it could also be generic enough to be matching grid/tile layouts.

skjnldsv:
> We already use that icon in a very present way, because it is included in any text editor all over nextcloud:

Which is fine as it is referring to... tables ;)

skjnldsv:
Considering Julius and Jan's comments as approval. Le'ts go :)

jancborchardt:
Yes, we can always adjust.
Further down the line, a new icon which reflects additional no-code capabilities can even go along with a "Tables 2"-like announcement which I’m sure @jospoortvliet would be happy about. :)

ChristophWurst:
(reflex)

jancborchardt:
@datenangebot I think "New idea 2" would fit the Nextcloud style best. :)

mortee:
Just my 2c, "new idea 2" seems the best to me. Apparently matches the style of the other icons, while also doesn't cause potential confusion.

Issue #465, opened 2023-08-11
 [API] CORS preflight returns "Internal Server Error 500"


logianer:
### Steps to reproduce

(Tested on v0.6.0-beta.1 and v0.5.1)
1. Make HTTP-Request with Method `OPTION` to any API Route (e.g. `nextcloud.local/apps/tables/api/1/tables`) *without* any Authorization (why? - see additional info)

### Expected behavior

Response has a Status of 200 and has set the required CORS-Headers from Nextclouds `CORSMiddleware`

### Actual behavior

Response is a Status of 500 (Internal Server Error) without the expected CORS-Headers.

Internally, the server logs the following Error:  
`
TypeError: OCA\Tables\Controller\Api1Controller::__construct(): Argument #11 ($userId) must be of type string, null given
`

### Tables app version

0.6.0-beta.1

### Browser

Firefox 116.0.2

### Client operating system

Windows 11

### Operating system

nextcloud:27.0.0-apache (official docker image)

### Web server

Apache

### PHP engine version

PHP 8.2

### Database

MariaDB

### Additional info

Why should it work without any Authorization headers?
According to the [spec](https://fetch.spec.whatwg.org/#cross-origin-request-with-preflight-0):
> For **a [CORS-preflight request](https://fetch.spec.whatwg.org/#cors-preflight-request)**, [request](https://fetch.spec.whatwg.org/#concept-request)’s [credentials mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) is always "same-origin", i.e., it **excludes credentials**

"A CORS-preflight request [...] excludes credentials" !

Now I wonder, if this is will be a problem in more Nextcloud-APIs implementing CORSMiddleware or if it is just a problem in this plugin (or just a problem of my setup), because `OCA\Tables\Controller\Api1Controller` expects to be constructed with a `$userId` wich is non-present if we call the preflight without any credentials.

logianer:
I just did a test install of nextcloud on a different server and it worked flawless...maybe my setup is buggy somehow

datenangebot:
done by #466

PR #466, opened 2023-08-12
Avoid error on CORS requests

logianer:
Made constructor parameter $userId of type `?string` (= be optional)
If no userId is passed (aka. is null),
$this->userId is set to an empty string to not cause any problems or errors.

See #465 for more info

### DIFF
--- a/lib/Controller/Api1Controller.php
+++ b/lib/Controller/Api1Controller.php
@@ -27,7 +27,7 @@ class Api1Controller extends ApiController {
 
 	private V1Api $v1Api;
 
-	private string $userId;
+	private ?string $userId;
 
 	protected LoggerInterface $logger;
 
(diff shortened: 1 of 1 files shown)


lib/Controller/Api1Controller.php
```diff
@@ -55,7 +55,7 @@ public function __construct(
 		$this->importService = $importService;
 		$this->viewService = $viewService;
 		$this->viewMapper = $viewMapper;
-		$this->userId = $userId;
+		$this->userId = $userId ?? '';
```
juliushaertl, line 58:
```suggestion
		$this->userId = $userId;
```

juliushaertl (changes requested):
Thanks a lot for your pull request. Your change mages sense, though the fallback to empty string is not required and we should probably just keep the null value then. Could you adjust that? Then we can get this merged for the next release

datenangebot (approved):

Issue #468, opened 2023-08-14
Do not show toast on successful actions that are visibile in the UI


juliushaertl:
Some examples:
- Share 
- Update permission
- delete share

Instead we should try to indicate the change inline, e.g. by disabling the checkboxes and showing a loading spinner until the operation is finished.

PR #470, opened 2023-08-16
fix: No need to show toast messages on success

juliushaertl:
As requested by @jancborchardt 

The checkbox indicating the value is enough and we should only show a toast in an error case.

Fix https://github.com/nextcloud/tables/issues/468


luka-nextcloud:
I think it would be better to show a loading spinner until the operation is finished as suggested at https://github.com/nextcloud/tables/issues/468

datenangebot (approved):

PR #492, opened 2023-08-28
ci: add renovate json config

### LINKED ISSUE #484: Move to renovate for dependency updates
As we do it in text, renovate would allow grouping dependency bumps of multiple dependencies (like all tiptap) into one PR.

Example config: https://github.com/nextcloud/text/blob/main/renovate.json

datenangebot:
closes #484

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,71 @@
+{
+	"$schema": "https://docs.renovatebot.com/renovate-schema.json",
+	"extends": [
+		"config:base"
+	],
+	"timezone": "Europe/Berlin",
+	"schedule": [
+		"every weekend"
+	],
+	"reviewers": ["team:office"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -0,0 +1,71 @@
...
+		{
+			"groupName": "tiptap",
+			"matchPackagePrefixes": [
+				"@tiptap/",
+				"prosemirror-"
+			]
+		},
+		{
+			"matchUpdateTypes": ["major"],
+			"matchBaseBranches": ["stable27", "stable26", "stable25"],
```
juliushaertl, line 46: Remove those and other stable* references as long as we don't actively maintain those stable branches?

datenangebot: 👍

PR #505, opened 2023-08-31
fix: Select tile on enter

### LINKED ISSUE #248: Template selector is not keyboard accessible

juliushaertl:
Follow up fix to close #248 

Tab navigation was possible but selecting a template using enter not


datenangebot (approved):
Thanks for adding this!

PR #508, opened 2023-08-31
fix: make the userId configurable for the analytics data provider to use

### LINKED ISSUE #464: AnalyticsDatasource error on 0.6.0-beta.1
```
{
  "reqId": "vXWpGRi4KlEUiw5xVSBA",
  "level": 3,
  "time": "2023-08-11T14:45:35+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "analytics",
  "method": "",
  "url": "--",
  "message": "Can not initialize data source: \"OCA\\\\Tables\\\\Datasource\\\\AnalyticsDatasource\"",
  "userAgent": "--",
  "version": "27.1.0.0",
  "data": {
    "app": "analytics"
  }
}
```

As we had some large refactoring and feature merged with views, this likely needs some adjustment. Having a proper exception trace would be quite nice in addition as well as test coverage for the data source. ;)

FYI @Rello

datenangebot:
closes #464

### DIFF
--- a/lib/Api/V1Api.php
+++ b/lib/Api/V1Api.php
@@ -10,7 +10,7 @@ use OCA\Tables\Service\RowService;
 class V1Api {
 	private RowService $rowService;
 	private ColumnService $columnService;
-	private string $userId;
+	private ?string $userId;
 
 	public function __construct(ColumnService $columnService, RowService $rowService, string $userId) {
 		$this->columnService = $columnService;
--- a/lib/Api/V1Api.php
+++ b/lib/Api/V1Api.php
@@ -12,7 +12,7 @@ class V1Api {
 	private ColumnService $columnService;
 	private ?string $userId;
 
-	public function __construct(ColumnService $columnService, RowService $rowService, string $userId) {
+	public function __construct(ColumnService $columnService, RowService $rowService, ?string $userId) {
 		$this->columnService = $columnService;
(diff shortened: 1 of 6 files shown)


lib/Api/V1Api.php
```diff
@@ -9,12 +9,13 @@
 
 class V1Api {
 	private RowService $rowService;
-
 	private ColumnService $columnService;
+	private string $userId;
 
-	public function __construct(ColumnService $columnService, RowService $rowService) {
+	public function __construct(ColumnService $columnService, RowService $rowService, string $userId) {
```
juliushaertl, line 15:
```suggestion
	public function __construct(ColumnService $columnService, RowService $rowService, ?string $userId) {
```

Dependency injection might pass in null, so this should be allowed to avoid errors in the logs

juliushaertl: You might want to instead check and throw an error in places where it is used

juliushaertl (changes requested):
Small comment regarding the constructor, otherwise seems good

lib/Api/V1Api.php
datenangebot: > You might want to instead check and throw an error in places where it is used

That should be handled by the permissions service.

juliushaertl (approved):

PR #510, opened 2023-09-01
fix(TextEditor): reset the editor, if an empty value is set

### LINKED ISSUE #459: Add more rows with rich text
When having a rich text in create row and if add more is enables, the text in the rich text field does not get deleted when saving the first row.

datenangebot:
- for example if the modal was resetted

closes #459


src/shared/components/ncEditor/NcEditor.vue
```diff
@@ -72,8 +72,15 @@ export default {
...
 			if (value.trim() !== this.localValue.trim()) {
-				this.editor.setContent(value, false)
 				this.localValue = value
+
+				// reset editor if content is empty
+				// otherwise the (empty) content will not be updated
+				if (value === '') {
+					this.setupEditor()
+				} else {
+					this.editor.setContent(value, true)
```
juliushaertl, line 82: Wondering about the second parameter now, there should only be one https://github.com/nextcloud/text/blob/a0f9aa64c10200479c6c1643a0ce33de221b5a76/src/editor.js#L76

src/shared/components/ncEditor/NcEditor.vue
```diff
@@ -72,8 +72,15 @@ export default {
 	watch: {
 		text(value) {
 			if (value.trim() !== this.localValue.trim()) {
-				this.editor.setContent(value, false)
 				this.localValue = value
+
+				// reset editor if content is empty
+				// otherwise the (empty) content will not be updated
+				if (value === '') {
+					this.setupEditor()
```
juliushaertl, line 80: Make sure to destroy possibly existing instances using this before:

```
this?.editor?.destroy()
```

src/shared/components/ncEditor/NcEditor.vue
datenangebot: Comes from the original interface?! https://github.com/ueberdosis/tiptap/blob/a1164fc03b7b08b6a17ed35476f0459ef4b9b106/packages/core/src/commands/setContent.ts#L14

But it was false and I just set true for testing. So we can remove it anyway...

src/shared/components/ncEditor/NcEditor.vue
datenangebot: this is what happens first in method this.setupEditor(), why do I have to do it before twice?

juliushaertl: Ah sorry, then I misread the diff somehow

juliushaertl (approved):

Issue #441, opened 2023-08-03
occ tables:owner does not change shares


Rello:
### Steps to reproduce

1. create a table with user A
2. share table with user B
3. transfer table to user C

### Expected behavior

User C is seeing existing shares - or all shares get removed when transferring the ownership?

### Actual behavior

User B is seeing the table as shared by user C (owner of the table).
User C does not see any share on the table, because in `tables_shares`, the `sender` is still user A.

### Web server

None

### PHP engine version

None

### Database

None

PR #519, opened 2023-09-07
fix(occ): change sender if a table owner changes

datenangebot:
closes: #441

### DIFF
--- a/lib/Service/PermissionsService.php
+++ b/lib/Service/PermissionsService.php
@@ -53,7 +53,7 @@ class PermissionsService {
 		if ($userId === null) {
 			$e = new \Exception();
 			$error = 'PreCheck for userId failed, requested in '. get_class($this) .'.';
-			$this->logger->debug($error, $e->getTrace());
+			$this->logger->debug($error, ['exception' => new \Exception()]);
 			throw new InternalError($error);
 		}
 
--- a/lib/Service/PermissionsService.php
+++ b/lib/Service/PermissionsService.php
@@ -51,8 +51,9 @@ class PermissionsService {
 		}
 
 		if ($userId === null) {
+			$e = new \Exception();
 			$error = 'PreCheck for userId failed, requested in '. get_class($this) .'.';
-			$this->logger->debug($error);
+			$this->logger->debug($error, $e->getTrace());
 			throw new InternalError($error);
 		}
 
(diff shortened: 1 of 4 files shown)


lib/Service/PermissionsService.php
```diff
@@ -51,8 +51,9 @@ public function preCheckUserId(string $userId = null, bool $canBeEmpty = true):
 		}
 
 		if ($userId === null) {
+			$e = new \Exception();
 			$error = 'PreCheck for userId failed, requested in '. get_class($this) .'.';
-			$this->logger->debug($error);
+			$this->logger->debug($error, $e->getTrace());
```
juliushaertl, line 56:
```suggestion
			$this->logger->debug($error, ['exception' => new \Exception()]);
```

lib/Service/PermissionsService.php
```diff
@@ -480,7 +481,7 @@ private function basisCheckById(int $elementId, string $nodeType, ?string &$user
 		return false;
 	}
 
-	private function userIsElementOwner(string $userId, $element): bool {
+	private function userIsElementOwner($element, string $userId = null): bool {
```
juliushaertl, line 484: Would be nice to type hint, maybe by introducing a base class or interface for all element classes

lib/Service/TableService.php
```diff
@@ -274,8 +275,27 @@ public function setOwner(int $id, string $newOwnerUserId, ?string $userId = null
 
 			$table->setOwnership($newOwnerUserId);
 			$table = $this->mapper->update($table);
-			$this->enhanceTable($table, $userId);
-			return $table;
+
+			# also change owners of related shares
```
juliushaertl, line 279:
```suggestion
			// also change owners of related shares
```

lib/Service/TableService.php
```diff
@@ -274,8 +275,27 @@ public function setOwner(int $id, string $newOwnerUserId, ?string $userId = null
...
+			$updatedShares = $this->shareService->changeSenderForNode('table', $id, $newOwnerUserId, $userId);
+			$updatesSharesOutput = [];
+			foreach ($updatedShares as $share) {
+				/* @var Share $share */
```
juliushaertl, line 283: Maybe better add a `@return Share[]` to the method generating that array?

juliushaertl (commented):
Just minor comments

lib/Service/TableService.php
datenangebot: Yeah, much better.

lib/Service/PermissionsService.php
datenangebot: We planned to review the permissions service at all. Lets handle that in there. #522

juliushaertl (approved):

PR #521, opened 2023-09-08
fix(design): edit columns

datenangebot:
- add column info popover
- don't show -1 as maxTextLength, which just means unlimited
- visual fixes

### DIFF
--- a/src/shared/components/ncTable/partials/columnTypePartials/forms/TextLineForm.vue
+++ b/src/shared/components/ncTable/partials/columnTypePartials/forms/TextLineForm.vue
@@ -28,7 +28,7 @@
 				{{ t('tables', 'Maximum text length') }}
 			</div>
 			<div class="fix-col-4">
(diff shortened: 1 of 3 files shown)


src/shared/components/ncTable/partials/columnTypePartials/forms/TextLineForm.vue
```diff
@@ -42,23 +42,32 @@
...
+
 	data() {
 		return {
 			mutableColumn: this.column,
 		}
 	},
+
+	computed: {
+		getMaxLength() {
+			return this.mutableColumn.textMaxLength >= 0 ? this.mutableColumn.textMaxLength : null
```
juliushaertl, line 61: A max text length of 0 doesn't make sense, right? Should it then just also be unlimited?

datenangebot: Mh, no.

juliushaertl (approved):

PR #533, opened 2023-09-11
Fix/529 Search for tables and views

### LINKED ISSUE #528: Can not find shared tables/views in widget config
Tables and views that are shared via a group can not be found in the widget configuration.

SCR-20230911-hirg

### LINKED ISSUE #529: Can not search for tables

datenangebot:
closes #529 
closes #528 

searching for tables was missing 🙄

### DIFF
--- a/lib/Search/SearchTablesProvider.php
+++ b/lib/Search/SearchTablesProvider.php
@@ -110,7 +110,7 @@ class SearchTablesProvider implements IProvider {
 				$appIconUrl,
 				$table->getEmoji() .' '. $table->getTitle(),
 				($table->getOwnerDisplayName() ?? $table->getOwnership()) . ', ' . $this->l10n->n('%n row', '%n rows', $table->getRowsCount()).', '.$this->l10n->t('table'),
-				$this->getInternalLink('table', $table->getId()),
+				$this->getInternalLink($table->getId(), 'table'),
 				'',
 				false
 			);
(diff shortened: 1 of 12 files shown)


lib/Search/SearchTablesProvider.php
```diff
@@ -96,17 +96,21 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
 		$offset = $query->getCursor();
 		$offset = $offset ? (int) $offset : 0;
 
+		// TODO how to read if dark mode is set to use light icons?
```
juliushaertl, line 99: There doesn't seem to be a proper way to do this as of right now.

The only working one is to use a css class as the icon parameter but this would require to load a stylesheet on every page. https://github.com/nextcloud/server/blob/4748626aa4a06a2c5f7d8e3c0bb0b1ddfd4e710e/lib/public/Search/SearchResultEntry.php#L104

datenangebot: Mh, okay.

juliushaertl:
Ah seems you found it already. Just to have it mentioned (maybe you figured it out that way as well). The logs + screenshots are uploaded as assets on the cypress run https://github.com/nextcloud/tables/actions/runs/6156012270?pr=533

The logs then also show the composer autoload error

juliushaertl (approved):

PR #543, opened 2023-09-12
fix(design): don't hide last column in ncTable used in widgets

### LINKED ISSUE #542: hidden last column in widget content views
SCR-20230912-mrri

datenangebot:
- this is needed because in widgets we don't have the edit-option as last column

closes #542


max-nextcloud (approved):
Assuming you're actually setting the `.sticky` class somewhere else.

PR #550, opened 2023-09-13
make use of dev setup default user

### LINKED ISSUE #549: adjust cypress ci setup to match local and ci needs

datenangebot:
- this allows to run the tests locally and in CI with the same setup
closes #549

### DIFF
--- a/cypress/e2e/tables-unified-search.cy.js
+++ b/cypress/e2e/tables-unified-search.cy.js
@@ -1,5 +1,7 @@
 let localUser
 let localUser2
+const alice = { userId: 'alice', password: 'alice' }
+const bob = { userId: 'bob', password: 'bob' }
 
 describe('The Home Page', () => {
 
@@ -75,7 +77,7 @@ describe('The Home Page', () => {
 	})
 
 	it('Search for shared table via group share', () => {
-		cy.login({ userId: 'user1', password: '1234561' })
+		cy.login(alice)
 		cy.visit('apps/tables')
 
 		// create table to share
@@ -92,13 +94,13 @@ describe('The Home Page', () => {
 
 		cy.get('h3').contains('Shares').parent().find('ul').contains('shareTestGroup').should('exist')
 
-		cy.login({ userId: 'user2', password: '1234561' })
+		cy.login(bob)
 		cy.visit('apps/tables')
 		cy.unifiedSearch('Share for group')
 	})
 
 	it('Search for shared view via group share', () => {
-		cy.login({ userId: 'user1', password: '1234561' })
+		cy.login(alice)
 		cy.visit('apps/tables')
 
 		// create table to share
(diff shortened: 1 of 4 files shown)


cypress/e2e/tables-unified-search.cy.js
```diff
@@ -75,7 +77,7 @@ describe('The Home Page', () => {
 	})
 
 	it('Search for shared table via group share', () => {
-		cy.login({ userId: 'user1', password: '1234561' })
+		cy.login(alice)
```
juliushaertl, line 80: It would be even better to just create those users with different non-conflicting user ids during the before hook

That way there is no dependency on existing setup or configuration on the server used for testing

Examples using the cypress helper library:
https://github.com/nextcloud/richdocuments/blob/a76639cb09f39c7566c1ced6c3e330eb261e77bc/cypress/e2e/share-link.js#L28-L30
https://github.com/nextcloud/richdocuments/blob/a76639cb09f39c7566c1ced6c3e330eb261e77bc/cypress/support/commands.js#L24C14-L24C14

cypress/support/local-setup.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+occ group:add shareTestGroup
+occ group:adduser shareTestGroup alice
+occ group:adduser shareTestGroup bob
```
juliushaertl, line 5: Would be nice to have this as a cypress command:

I should be fairly easy to add an ocs api call similar to https://github.com/nextcloud/text/blob/main/cypress/support/commands.js#L132-L145 calling this API:

https://docs.nextcloud.com/server/latest/developer_manual/_static/openapi.html#/operations/provisioning_api-users-add-to-group

datenangebot: much better, thanks!

cypress/e2e/tables-unified-search.cy.js
datenangebot: much better, thanks!

cypress/e2e/tables-unified-search.cy.js
```diff
@@ -1,14 +1,21 @@
...
+				localUser = user
+			}).then(user => {
+				cy.addUserToGroup(user.userId, groupId)
+			})
+			cy.createRandomUser().then(user => {
+				localUser2 = user
+			}).then(user => {
+				cy.addUserToGroup(user.userId, groupId)
+			})
 		})
```
max-nextcloud, line 19: I don't think you need the `.then()` after `cy.createGroup(groupId)`. It's implicit as cypress will wait for the command to finish if it returns a promise. If it does not then adding `.then(...)` also won't help.

So here's a little less nested version:
```suggestion
		cy.createGroup(groupId)
		cy.createRandomUser().then(user => {
			localUser = user
		}).then(user => {
			cy.addUserToGroup(user.userId, groupId)
		})
		cy.createRandomUser().then(user => {
			localUser2 = user
		}).then(user => {
			cy.addUserToGroup(user.userId, groupId)
		})
```

max-nextcloud: Bonus exercise ;)... If you make use of `localUser` and `localUser2` you can also get rid of the `then` function around the calls to `cy.addUserToGroup`.

max-nextcloud (approved):

PR #556, opened 2023-09-13
fix(design): fix loading the default column-link providers

datenangebot:
- add configuration to avoid saving incorrect column configs
- add info cards if no providers are chosen

SCR-20230913-traj
SCR-20230913-trce


AndyScherzinger (approved):
🐘💙

juliushaertl (approved):

PR #568, opened 2023-09-14
add initial release script

datenangebot:


juliushaertl (commented):
Maybe move to a subdirectory for helper scripts? Also would be good to add to the .nextcloudexclude file to not ship it

juliushaertl (approved):

PR #615, opened 2023-10-12
API docs and api initial v2

datenangebot:
- [x] add annotations for API routed methods
- [x] refactor to ensure type safety
- [x] add psalm types and usage
- [x] introduce new api v2 structure
- [x] refactor integration tests / add for api v2


provokateurin:
> Here is a APIignore because of an psalm/api-docs error. It's known and in progress. https://github.com/nextcloud/tables/pull/615/files#diff-70e5309fcd3311d771b3db9c93490ea270fd5894769093765ec37edb68e5dd9bR24

I'm not sure what you mean? You shouldn't have `@NoCSRFRequired` on a method that serves the frontend files. The openapi-extractor detects this as reachable from the outside because it is in fact reachable from the outside.

juliushaertl:
> @juliushaertl I don't need this when using the ocs routes, right? [#615 (files)](https://github.com/nextcloud/tables/pull/615/files#diff-6301937af1cc1575e9e18ecd27c4a51f0d3cf9ce12e59a14709f626dfa3ef952R9-R10)

Yes, this is only needed for CORS annotated controllers, not for OCS ones.



> > Here is a APIignore because of an psalm/api-docs error. It's known and in progress. [#615 (files)](https://github.com/nextcloud/tables/pull/615/files#diff-70e5309fcd3311d771b3db9c93490ea270fd5894769093765ec37edb68e5dd9bR24)
> 
> I'm not sure what you mean? You shouldn't have `@NoCSRFRequired` on a method that serves the frontend files. The openapi-extractor detects this as reachable from the outside because it is in fact reachable from the outside.

Why not? The page controller that serves the page doesn't get a CSRF token passed by the browser. This is one of the rare cases that should have the `@NoCSRFRequired` annotation. ;)

lib/Controller/MyOCSController.php
```diff
@@ -0,0 +1,58 @@
...
+use OCA\Tables\Errors\InternalError;
+use OCA\Tables\Errors\NotFoundError;
+use OCA\Tables\Errors\PermissionError;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\DataResponse;
+use OCP\AppFramework\OCSController;
+use OCP\IRequest;
+use Psr\Log\LoggerInterface;
+
+class MyOCSController extends OCSController {
```
juliushaertl, line 16:
```suggestion
abstract class AOCSController extends OCSController {
```

To make it clear that this is a base class that will not be initialized on its own? The `A*` is a pattern we use in some other code parts in server to indicate abstract classes in the naming. I'd prefer that more over the `My` which doesn't have much meaning from my perspective.

lib/Controller/MyOCSController.php
```diff
@@ -0,0 +1,58 @@
...
+		$this->userId = $userId;
+	}
+
+	/**
+	 * @param Exception|InternalError $e
+	 * @return DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
+	 */
+	protected function handleError(InternalError|Exception $e): DataResponse {
+		$this->logger->warning('An internal error or exception occurred: ' . $e->getMessage());
+		return new DataResponse(['message' => $e->getMessage()], Http::STATUS_INTERNAL_SERVER_ERROR);
```
juliushaertl, line 37: We should avoid passing the exception message to the user as it may contain security related information. 

Best would be to use https://github.com/nextcloud/server/blob/d55a7c619d3733863387cc8c6269050b72cbff0c/lib/public/HintException.php#L76 for custom exception messages that you want to expose.

lib/Controller/ApiTablesController.php
```diff
@@ -0,0 +1,150 @@
...
+		string $userId) {
+		parent::__construct($request, $logger, $userId);
+		$this->service = $service;
+	}
+
+	/**
+	 * [api v2] Returns all Tables
+	 *
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
```
juliushaertl, line 35:
```suggestion
```

Those `NoCSRFRequired` should not be there for the OCS API methods anymore.

lib/Controller/ApiTablesController.php
```diff
@@ -0,0 +1,150 @@
...
+	 * @param int $id Table ID
+	 * @return DataResponse<Http::STATUS_OK, TablesTable, array{}>|DataResponse<Http::STATUS_FORBIDDEN|Http::STATUS_INTERNAL_SERVER_ERROR|Http::STATUS_NOT_FOUND, array{message: string}, array{}>
+	 *
+	 * 200: Table returned
+	 * 403: No permissions
+	 * 404: Not found
+	 */
+	public function show(int $id): DataResponse {
+		try {
+			return new DataResponse($this->service->find($id)->jsonSerialize());
```
juliushaertl, line 64:
```suggestion
			return new DataResponse($this->service->find($id));
```

jsonSerialize should be called automatically once the DataResponse is passed to json_encode, but also fine to keep it explicit here.

lib/Controller/ApiGeneralController.php
```diff
@@ -0,0 +1,71 @@
...
+	 *
+	 * @NoAdminRequired
+	 * @CORS
+	 * @NoCSRFRequired
+	 *
+	 * @return DataResponse<Http::STATUS_OK, TablesIndex, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
+	 *
+	 * 200: Index returned
+	 */
+	public function initIndex(): DataResponse {
```
juliushaertl, line 60:
```suggestion
	public function index(): DataResponse {
```

I found the init a bit confusing as it doesn't initialize?

lib/Controller/MyOCSController.php
datenangebot: Much better, thanks

lib/Controller/ApiTablesController.php
datenangebot: psalm needs this to get the correct return type... ?!

provokateurin: Yep, this is needed for psalm to understand what is going on

provokateurin:
> Wouldn't it be possible to define a custom type as it is actually an object json structure and not an array?

This is possible, but there is no way to tell psalm that a string is actually a json encoded object. So the structure itself can be properly documented, just not that it is used here (except for a comment).

appinfo/routes.php
```diff
@@ -104,5 +105,22 @@
...
+		['name' => 'ApiTables#create', 'url' => '/api/2/tables', 'verb' => 'POST'],
+		['name' => 'ApiTables#update', 'url' => '/api/2/tables/{id}', 'verb' => 'PUT'],
+		['name' => 'ApiTables#destroy', 'url' => '/api/2/tables/{id}', 'verb' => 'DELETE'],
+
+		['name' => 'ApiColumns#index', 'url' => '/api/2/columns/{nodeType}/{nodeId}', 'verb' => 'GET'],
+		['name' => 'ApiColumns#show', 'url' => '/api/2/columns/{id}', 'verb' => 'GET'],
+		['name' => 'ApiColumns#createNumberColumn', 'url' => '/api/2/columns/number', 'verb' => 'POST'],
+		['name' => 'ApiColumns#createTextColumn', 'url' => '/api/2/columns/text', 'verb' => 'POST'],
+		['name' => 'ApiColumns#createSelectionColumn', 'url' => '/api/2/columns/selection', 'verb' => 'POST'],
+		['name' => 'ApiColumns#createDatetimeColumn', 'url' => '/api/2/columns/datetime', 'verb' => 'POST'],
```
juliushaertl, line 124: No update/delete for rows in v2?

datenangebot: As we discussed this is just the initial start for api v2, more like the rows endpoints will added by time.

juliushaertl (approved):

PR #621, opened 2023-10-19
Feature pushed to PR

ahmed15hasan:


package.json
```diff
@@ -24,6 +24,7 @@
     "stylelint:fix": "stylelint 'css/*.css' 'css/*.scss' 'src/**/*.scss' 'src/**/*.vue' --fix"
   },
   "dependencies": {
+    "16": "^0.0.2",
```
QazCetelic, line 27: Why is this dependency required?

lib/Service/SuperService.php
QazCetelic, line 1: `SuperService` does not make it clear what the service does.

lib/Service/SuperService.php
```diff
@@ -16,4 +16,39 @@ public function __construct(LoggerInterface $logger, ?string $userId, Permission
 		$this->logger = $logger;
 		$this->userId = $userId;
 	}
+
+	public function AdminNotificationCall($tableId,$viewId) {
```
QazCetelic, line 20: `AdminNotificationCall` is not descriptive.

lib/Service/SuperService.php
```diff
@@ -16,4 +16,39 @@ public function __construct(LoggerInterface $logger, ?string $userId, Permission
...
+	 
+        $url = 'http://admin:admin@localhost/ocs/v2.php/apps/notifications/api/v2/admin_notifications/admin';
+
+		$conditionalTerm = (!is_null($tableId)) ? 'Table Id '.$tableId: 'View Id '.$viewId;
+		
+		$data = array(
+			'shortMessage' => 'Alert !',
```
QazCetelic, line 27: Text is not localized

package.json
ahmed15hasan: Actually, I was facing the npm version errors while running the **npm ci** or **npm build**.  The table view wasn't loading due to these errors. That's the reason for the above change.

lib/Service/SuperService.php
ahmed15hasan: I can't find the feature of language switching the in application. If you want to develop the functionality like that I can do it.

lib/Service/SuperService.php
ahmed15hasan: Let me change it.

lib/Service/SuperService.php
QazCetelic: Language switching is built in into Nextcloud. It can be implemented using the `t` function, which replaces the text with text translated in the chosen language. The language files are located in the `l10n` directory.

An example is line 56 in `src/store/data.js`.
```js
} catch (e) {
  displayError(e, t('tables', 'Could not load columns.'))
  return false
}
```

If the user uses Arabic it will load the text from `l10n/ar.js` at line 431:
```json
"Could not load columns." : "تعذر تحميل الأعمدة.",
```

package.json
QazCetelic: Does that mean the dependency isn't required?

lib/Service/SuperService.php
```diff
@@ -16,4 +16,39 @@ public function __construct(LoggerInterface $logger, ?string $userId, Permission
...
+		curl_setopt($ch, CURLOPT_HTTPHEADER, array('OCS-APIREQUEST: true'));
+		curl_setopt($ch, CURLOPT_POSTFIELDS, http_build_query($data));
+		curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+
+		// Execute cURL session and store the response
+		$response = curl_exec($ch);
+
+		// Check for cURL errors
+		if (curl_errno($ch)) { 
+			echo 'cURL error: ' . curl_error($ch);
```
QazCetelic, line 45: This should use the dedicated logger.

lib/Controller/RowController.php
```diff
@@ -64,13 +66,19 @@ public function show(int $id): DataResponse {
...
-		array $data
-	): DataResponse {
-		return $this->handleError(function () use ($tableId, $viewId, $data) {
-			return $this->service->create(
+		array $data,
+		SuperService $superService
+	): DataResponse { 
+		return $this->handleError(function () use ($tableId, $viewId, $data, $superService) { 
+
+			$create = $this->service->create(
```
juliushaertl, line 74: For some more obvious naming

```suggestion
			$newRow = $this->service->create(
```

lib/Service/SuperService.php
```diff
@@ -16,4 +16,39 @@ public function __construct(LoggerInterface $logger, ?string $userId, Permission
 		$this->logger = $logger;
 		$this->userId = $userId;
 	}
+
+	public function AdminNotificationCall($tableId,$viewId) {
+	 
+        $url = 'http://admin:admin@localhost/ocs/v2.php/apps/notifications/api/v2/admin_notifications/admin';
```
juliushaertl, line 22: This is not the expected API to be used, the reference is about https://github.com/nextcloud/notifications/blob/master/docs/notification-workflow.md which is a PHP API provided by Nextcloud server. There is no need to perform an HTTP request here especially with passing username/password.

lib/Service/SuperService.php
juliushaertl: It would require to inject the IL10N service, further reference can be found at https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#php

lib/Service/SuperService.php
```diff
@@ -16,4 +16,39 @@ public function __construct(LoggerInterface $logger, ?string $userId, Permission
...
+		// Check for cURL errors
+		if (curl_errno($ch)) { 
+			echo 'cURL error: ' . curl_error($ch);
+		}
+
+		// Close cURL session
+		curl_close($ch);
+
+		// Output the response
+		return $response;
```
juliushaertl, line 52: Is there a need to return that?

lib/Service/SuperService.php
juliushaertl, line 20: How about adding types?

package.json
juliushaertl: Yeah this should not be needed, can you drop the changes from package.json/package-lock.json from the PR?

juliushaertl (changes requested):

package.json
ahmed15hasan: Yeah sure, let me do it.

juliushaertl:
Yes, when integrating the notifications app this would also not work as we'd need to use the PHP OCP API for emitting the notifications

PR #637, opened 2023-10-30
chore(github): exclude github datenangebot from pr-feedback

datenangebot:


max-nextcloud:
You might want to add your nick here as well: https://github.com/nextcloud/.github/blob/master/workflow-templates/pr-feedback.yml

juliushaertl (approved):

PR #643, opened 2023-11-02
refactor(design): redesign the filter form 

### LINKED ISSUE #611: Several display problems with the view filter
### Steps to reproduce

Screenshot 2023-10-05 at 20-05-35 Tabellen - Steffen's Cloud

I try to filter all entries with a date from a specified year.

I thought, I set one filter to "larger than 2023-01-01" and another to "smaller than 2023-12-31".

### Expected behavior

I expected to see just the entries with a date from 2023.

### Actual behavior

See screenshot above:
1. The field in the middle is too small for the German words "ist größer als".
2. The last field seems to be too small for "2023-01-01". 
3. I can't add a date that is larger than today it seems. "2023-12-31" at least is wrong and won't be used as a filter.

### Tables app version

0.6.0

### Browser

Firefox 118.0.1 

### Client operating system

Ubuntu

### Operating system

No sure

### Web server

Apache

### PHP engine version

PHP 8.2

### Database

MariaDB

### Additional info

_No response_

datenangebot:
- use grid layout
- split fields into 2 lines
- add background and spaces to separate different filters by design

closes #611 

| Before | After |
| --- | --- |
| SCR-20231102-jcsx | SCR-20231102-izzo |

### DIFF
--- a/src/modules/main/partials/ColumnFormComponent.vue
+++ b/src/modules/main/partials/ColumnFormComponent.vue
@@ -75,7 +75,7 @@ export default {
 				group
 					.toUpperCase()
 					.replace('_', '')
-					.replace('-', '')
+					.replace('-', ''),
 			)
 			return str.charAt(0).toUpperCase() + str.slice(1)
 		},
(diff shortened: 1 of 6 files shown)


juliushaertl:
Wouldn't it be better to only split to two lines if there is not enough space?

juliushaertl:
As briefly discussed with @datenangebot Let's keep this as a quick fix as it is. We can implement dynamic wrapping later on but this is more tricky to cover all cases.

juliushaertl (approved):

PR #654, opened 2023-11-08
fix(column link): fix parsing json instead using regex

datenangebot:
this fix should bring back a green CI...

### DIFF
--- a/lib/Service/ColumnTypes/TextLinkBusiness.php
+++ b/lib/Service/ColumnTypes/TextLinkBusiness.php
@@ -16,10 +16,26 @@ class TextLinkBusiness extends SuperBusiness implements IColumnTypeBusiness {
 			]));
 		}
 
+		// if is json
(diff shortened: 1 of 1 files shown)


blizzz (approved):

lib/Service/ColumnTypes/TextLinkBusiness.php
```diff
@@ -16,10 +16,26 @@ public function parseValue(string $value, ?Column $column = null): string {
 			]));
 		}
 
+		// if is json
+		/** @noinspection PhpComposerExtensionStubsInspection */
```
juliushaertl, line 20: I think you could also just add `"ext-json": "*"` as a dependency to composer.json to generally make phpstorm aware of that. We require the json extension anyways for Nextcloud server

datenangebot: https://github.com/nextcloud/tables/pull/656

PR #668, opened 2023-11-10
fix: Update browser page title depending on current view/table

juliushaertl:
Make sure to update the browser window title with the current table or view title.

Screenshot 2023-11-10 at 13 11 00
Screenshot 2023-11-10 at 13 10 36

Code is a shamless copy from deck ;)


datenangebot (approved):
looks good

PR #679, opened 2023-11-14
Fix/678

### LINKED ISSUE #678: Table with selection choice as mandatory field, no update possible.
### Steps to reproduce

1. An existing table with a selection field as a mandatory field and, for example, a text field.
2. An existing data set, fully recorded.
3. Open the update dialog and click the Save button. Optionally with changes to the text field.
4. An error message appears: "Please fill in the mandatory fields"

### Expected behavior

No error message, as the mandatory field already contains data.

### Actual behavior

An error message, although data is already available.

### Tables app version

0.6.2

### Browser

Firefox 119.0.1

### Client operating system

Win10

### Operating system

Linux

### Web server

Apache

### PHP engine version

PHP 8.1

### Database

MariaDB

### Additional info

_No response_

datenangebot:
closes #678

### DIFF
--- a/lib/Service/ColumnTypes/SelectionBusiness.php
+++ b/lib/Service/ColumnTypes/SelectionBusiness.php
@@ -17,8 +17,8 @@ class SelectionBusiness extends SuperBusiness implements IColumnTypeBusiness {
 			return '';
 		}
 
-		$intValue = (int) $value;
-		if ((string)$intValue === $value) {
+		$intValue = (int)$value;
+		if ((string)$intValue === (string)$value) {
 			// if it seems to be an option ID
 			foreach ($column->getSelectionOptionsArray() as $option) {
 				if($option['id'] === $intValue && $option['label'] !== $value) {
@@ -51,7 +51,7 @@ class SelectionBusiness extends SuperBusiness implements IColumnTypeBusiness {
 		}
 
 		$intValue = (int) $value;
-		if ("".$intValue === $value) {
+		if ((string) $intValue === (string) $value) {
 			// if it seems to be an option ID
 			foreach ($column->getSelectionOptionsArray() as $option) {
 				if($option['id'] === $intValue && $option['label'] !== $value) {
--- a/lib/Service/RowService.php
+++ b/lib/Service/RowService.php
@@ -183,7 +183,7 @@ class RowService extends SuperService {
 	 * @throws InternalError
 	 */
 	private function cleanupData(array $data, array $columns, ?int $tableId, ?int $viewId): ?array {
-		$out = null;
+		$out = [];
 		foreach ($data as $entry) {
 			$column = $this->getColumnFromColumnsArray((int) $entry['columnId'], $columns);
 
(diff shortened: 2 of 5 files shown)


lib/Service/RowService.php
```diff
@@ -183,7 +183,7 @@ public function create(?int $tableId, ?int $viewId, array $data):Row {
 	 * @throws InternalError
 	 */
 	private function cleanupData(array $data, array $columns, ?int $tableId, ?int $viewId): ?array {
```
mejo-, line 185:
```suggestion
	private function cleanupData(array $data, array $columns, ?int $tableId, ?int $viewId): array {
```

Now that you always return an array, you probably can adust the return type. Also needs to be adjusted in the annotations then.

lib/Service/ColumnTypes/SelectionBusiness.php
```diff
@@ -51,7 +51,7 @@ public function canBeParsed($value, ?Column $column = null): bool {
 		}
 
 		$intValue = (int) $value;
-		if ("".$intValue === $value) {
+		if ((string) $intValue === (string) $value) {
```
mejo-, line 54:
```suggestion
		if ((string)(int)$value === (string)$value) {
```

I guess these two lines can be unified into one line.

lib/Service/ColumnTypes/SelectionBusiness.php
```diff
@@ -18,7 +18,7 @@ public function parseValue($value, ?Column $column = null): string {
 		}
 
 		$intValue = (int) $value;
-		if ((string)$intValue === $value) {
+		if ((string) $intValue === (string) $value) {
```
mejo-, line 21:
```suggestion
		if ((string)(int)$value === (string)$value) {
```

I guess these two lines can be unified into one line.

mejo- (approved):
I didn't follow the logic how this will fix the referenced issue, but code changes look sensible to me.

lib/Service/ColumnTypes/SelectionBusiness.php
datenangebot: Mh, I need that value later on and with the variable name it helps me to understand the flow?!

mejo-: Ah, if you need `$intValue` later on, nevermind.

lib/Service/RowService.php
```diff
@@ -223,7 +223,7 @@ private function parseValueByColumnType(Column $column, $value = null) {
 		} catch (NotFoundExceptionInterface|ContainerExceptionInterface $e) {
 			$this->logger->debug('Column type business class not found', ['exception' => $e]);
 		}
-		return "".$value;
+		return (string) $value;
```
blizzz, line 226: Same can be done in line 220. 

In line 222 we can legitimately get mixed return types, right?

datenangebot: About the return types. Technically yes, but I do only expect int, float or strings. 🤷🏻‍♂️

PR #704, opened 2023-11-21
fix(api): add hint for the right format to use

### LINKED ISSUE #384: Updating `mandatory` property of a `column` via REST API always sets `mandatory` to `true`
### Steps to reproduce

Web UI: Define a column with `title` = `Foo` and set `mandatory` to `true`.

Query | Expected state | Actual state
--- | --- | ---
:heavy_check_mark: `curl -X PUT -u 'user:password' https://nextcloud.local/index.php/apps/tables/api/1/columns/30?title=Bar` | `title` = `Bar`, `mandatory` = `true` | `title` = `Bar`, `mandatory` = `true`
:heavy_check_mark: `curl -X PUT -u 'user:password' https://nextcloud.local/index.php/apps/tables/api/1/columns/30?mandatory=true` | `title` = `Bar`, `mandatory` = `true` | `title` = `Bar`, `mandatory` = `true`
:x: `curl -X PUT -u 'user:password' https://nextcloud.local/index.php/apps/tables/api/1/columns/30?mandatory=false` | `title` = `Bar`, `mandatory` = `false` | `title` = `Bar`, `mandatory` = `true`

### Expected behavior

- If the `mandatory` query parameter is not set while updating a column, the resulting state of the `mandatory` property is the same as before the `PUT`
- If the `mandatory` query parameter is set, the resulting state after the `PUT` should match the value of the query parameter.

### Actual behavior

- If the `mandatory` query parameter is not set while updating a column, the resulting state of the `mandatory` property is the same as before the `PUT` (:heavy_check_mark: as expected)
- If the `mandatory` query parameter is set, the resulting state after the `PUT` is *always* `true`, no matter what the actual value of the query parameter was. (:x: Does not work as expected)

### Tables app version

0.5.1

### Browser

_No response_

### Client operating system

_No response_

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

_No response_

datenangebot:
closes #384

### DIFF
--- a/lib/Controller/Api1Controller.php
+++ b/lib/Controller/Api1Controller.php
@@ -477,11 +477,11 @@ class Api1Controller extends ApiController {
 	 * @param 'table'|'view' $nodeType Node type
 	 * @param string $receiver Receiver ID
 	 * @param 'user'|'group' $receiverType Receiver type
-	 * @param bool $permissionRead Permission if receiver can read data
-	 * @param bool $permissionCreate Permission if receiver can create data
-	 * @param bool $permissionUpdate Permission if receiver can update data
-	 * @param bool $permissionDelete Permission if receiver can delete data
-	 * @param bool $permissionManage Permission if receiver can manage node
+	 * @param bool $permissionRead Permission if receiver can read data (use 0 for false or 1 for true as int values)
+	 * @param bool $permissionCreate Permission if receiver can create data (use 0 for false or 1 for true as int values)
+	 * @param bool $permissionUpdate Permission if receiver can update data (use 0 for false or 1 for true as int values)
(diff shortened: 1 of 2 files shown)


provokateurin:
I don't think you need to document this. In the OpenAPI spec it is already correct.

juliushaertl:
I'm also unsure this is actually the case that you need to pass 0, using `true` or `false` as boolean values in requests seems to work quite fine for me.

provokateurin:
0/1 is needed if you use query parameters. If you use a json body then true/false is what you need to use.

juliushaertl:
Ok, I only tested GET where this is handled in https://github.com/nextcloud/server/blob/e7cc7653b885c49b1b3f0a78f91ea05a53e102d8/lib/private/AppFramework/Http/Dispatcher.php#L219

PR #709, opened 2023-11-22
fix: add display names to all column meta data printouts

datenangebot:


lib/Service/ColumnService.php
```diff
@@ -519,4 +523,31 @@ public function getColumnsCount(int $tableId): int {
...
+	private function enhanceColumns(?array $columns): array {
+		if ($columns === null) {
+			return [];
+		}
+
+		$out = [];
+		foreach ($columns as $column) {
+			$out[] = $this->enhanceColumn($column);
+		}
+		return $out;
```
blizzz, line 551: you should not need `$out` here, as objects are always passed by reference. You can return `$columns`.

blizzz (approved):

PR #710, opened 2023-11-22
Fix/692

### LINKED ISSUE #692: parse values for number columns throws errors
1. create a number column
2. set min to 0.5
3. open "create row"
4. console will show up

`Cannot read properties of null (reading 'replace')`

datenangebot:
closes #692 

also fixes:
- handle empty values for number columns correctly
- allow to remove column number config values
- add validation for default value in combination with min and max

SCR-20231122-iwzg

### DIFF
--- a/lib/Service/ColumnService.php
+++ b/lib/Service/ColumnService.php
@@ -362,48 +362,26 @@ class ColumnService extends SuperService {
 			if ($subtype !== null) {
 				$item->setSubtype($subtype);
 			}
-			if ($numberPrefix !== null) {
-				$item->setNumberPrefix($numberPrefix);
-			}
-			if ($numberSuffix !== null) {
-				$item->setNumberSuffix($numberSuffix);
-			}
+			$item->setNumberPrefix($numberPrefix);
+			$item->setNumberSuffix($numberSuffix);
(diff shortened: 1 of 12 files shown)


lib/Service/ColumnService.php
```diff
@@ -362,48 +362,30 @@ public function update(
 			if ($subtype !== null) {
 				$item->setSubtype($subtype);
 			}
-			if ($numberPrefix !== null) {
+			if($numberPrefix !== null) {
```
juliushaertl, line 365:
```suggestion
			if ($numberPrefix !== null) {
```

Not sure why php-cs-fixer doesn't complain or auto fix this, but would be more aligned to the rest of the code base

lib/Service/ColumnService.php
```diff
@@ -362,48 +362,30 @@ public function update(
...
 				$item->setNumberPrefix($numberPrefix);
 			}
-			if ($numberSuffix !== null) {
+			if($numberSuffix !== null) {
```
juliushaertl, line 368:
```suggestion
			if ($numberSuffix !== null) {
```

juliushaertl (approved):

PR #715, opened 2023-11-23
chore: compatible with NC29

datenangebot:


juliushaertl:
Maybe adjust the cypress matrix to stable28 instead of 27 to test min and max stable version, but no strong opinions on that for now.

juliushaertl (approved):

PR #716, opened 2023-11-24
fix(CI): remove search tests because they are incompatible with nc28

datenangebot:
This is needed to bring the CI back to green for further business.
We have to bring the tests back, tracked with: #717


juliushaertl (approved):
You could use describe.skip if you want to keep the test code still for now

PR #733, opened 2023-12-05
fix(test): deactivate the deactivation of the contacts search provider

datenangebot:
- ⁉️have to clarify if the search provider will come back by default or not


juliushaertl:
This is actually an intended change from https://github.com/nextcloud/server/pull/41738 so it only shows up if contacts is enabled as an app in the future

juliushaertl:
Pushed some adjustments to make them all pass

juliushaertl (approved):

enjeck (commented):
This is already merged, but adding a few thoughts that might be worth considering in the future.

juliushaertl:
Yes, very good points. We can try to improve the test cases whenever we adjust them next time 👍

Issue #711, opened 2023-11-22
transfer ownership via webgui


datenangebot:
### Describe the solution you'd like

in modal "edit table" add an option to transfer ownership.
This can be only done by the real owner (not table managers).

PR #741, opened 2023-12-14
feat: transfer ownership via webgui

enjeck:
This addresses https://github.com/nextcloud/tables/issues/711
The goal is to add a way to transfer table ownershi using the webgui


enjeck:
I forgot to sign my commit messages. Can ammend that later

enjeck:
I'm still very confused about what the UI should look like. At https://github.com/nextcloud/tables/issues/711, it suggest adding it to the "edit modal". But that modal already seems crowded, so not sure where to fit the content. Or maybe I'm trying to fit in more content than required. Here's the design I have so far, ready to move things around:
Screenshot from 2023-12-13 19-45-31
Screenshot from 2023-12-13 19-49-17
Screenshot from 2023-12-13 19-49-45
Screenshot from 2023-12-13 19-50-14


Again, still confused about what the UI should look like and planning to move things around or redo after clarification.

appinfo/routes.php
```diff
@@ -18,6 +18,7 @@
 		['name' => 'api1#index', 'url' => '/api/1/tables', 'verb' => 'GET'],
 		['name' => 'api1#createTable',	'url' => '/api/1/tables', 'verb' => 'POST'],
 		['name' => 'api1#updateTable',	'url' => '/api/1/tables/{tableId}', 'verb' => 'PUT'],
+		['name' => 'api1#transferTable',	'url' => '/api/1/tables/{tableId}', 'verb' => 'PUT'],
```
enjeck, line 21: I don't understand how routes work. I assume that every 'url' here has to be unique? Not sure if this feature should be it's own endpoint and how it should look like. I considered using the `updateTable` and having a query string or something, but I don't know how to do that or if that's the best way to go about it.

lib/Controller/Api1Controller.php
```diff
@@ -196,6 +196,33 @@ public function updateTable(int $tableId, string $title = null, string $emoji =
...
+			$this->logger->warning('A permission error occurred: ' . $e->getMessage());
+			$message = ['message' => $e->getMessage()];
+			return new DataResponse($message, Http::STATUS_FORBIDDEN);
+		} catch (InternalError|Exception $e) {
+			$this->logger->warning('An internal error or exception occurred: '.$e->getMessage());
+			$message = ['message' => $e->getMessage()];
+			return new DataResponse($message, Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+	}
+
```
enjeck, line 225: This is how I envision the actual "transfer ownership" functionality to be done. But I have no idea how to connect this to the frontend. Or if this is the right way to go about it.

src/modules/sidebar/partials/TransferForm.vue
```diff
@@ -0,0 +1,187 @@
...
+
+			const request = await axios.get(generateOcsUrl('apps/files_sharing/api/v1/sharees'), {
+				params: {
+					format: 'json',
+					itemType: 'file',
+					search,
+					perPage: this.maxAutocompleteResults,
+					shareType,
+				},
+			})
```
enjeck, line 105: When trying to find users to transfer ownership to, we need to load that from somewhere? I just copied this from the "sharing" functionality. Not sure if we need another api endpoint for this or reusing this one works

enjeck (commented):
I do not yet understand how the backend and frontend connect to each other. I plan to spend more time on more tutorials to understand this. Since the associates issue for this was marked as a "good first issue", I thought I could easily handle this after following the introductory tutorial but I couldn't. So it's either I'm overcomplicating this and there's an easier way to go about it, or it's more complicated than it seems. Would appreciate hints on what's the right direction to follow

src/store/store.js
```diff
@@ -234,6 +234,8 @@ export default new Vuex.Store({
 			commit('setTables', [...tables])
 			return true
 		},
+		async transferTable({ state, commit, dispatch }, { tableId, newOwnerUserId, userId}) {
+		},
```
enjeck, line 238: I'm guessing this is whereI wanna call the backend code that does the ownership transfer. Not sure how to do that.

src/modules/sidebar/partials/TransferForm.vue
datenangebot: Reusing this is fine. If you want to make it better, we could use the "new" endpoint for this. Can be found here: https://docs.nextcloud.com/server/latest/developer_manual/client_apis/OCS/ocs-api-overview.html#auto-complete-and-user-search

Both endpoints ensures that we only see users that we are allowed to see. Taking the config "share only with own groups" into account. Unfortunately this settings is in the _files_ sharing settings... 🙄

src/store/store.js
datenangebot: This is a good place. We have to add the connection with the backend and also can update the data within the store directly.

enjeck:
> > I'm still very confused about what the UI should look like
> 
> The "edit table" modal was meant, that one is not crowded?! Cloud look like this: image
> 
> And if you click on the "change owner" button, your modal gets opened.
> 
> Btw: I guess the word "share" is not correct here, we have sharing option elsewhere for real _sharing_.

Ah, I see what you mean. Makes sense. Should have asked clarified earlier. Thanks

src/modules/modals/TransferTable.vue
```diff
@@ -2,49 +2,56 @@
...
+
+				<!-- We don't need to show this, since the current user will only see this if they're the owner? -->
+				<div v-if="localTable !== undefined" class="col-3 inline space-T-small">
+					<NcUserBubble
+						:margin="4"
+						:size="30"
+						:display-name="localTable.ownerDisplayName"
+						:user="localTable.owner" />
+				</div>
 			</div>
```
enjeck, line 24: I don't think we should show the current owner on this modal. We already show this on the main "Edit Table" modal itself. No reason having it here, imo

src/modules/modals/TransferTable.vue
```diff
@@ -2,49 +2,56 @@
...
-					<NcButton v-if="!prepareTransferTable" type="error" @click="prepareTransferTable = true">
+					<NcButton type="warning" :disabled="newOwnerUserId === ''" @click="transferMe">
 						{{ t('tables', 'Transfer') }}
 					</NcButton>
-					<NcButton v-if="prepareTransferTable"
-						:wide="true"
-						type="error"
-						@click="transferMe">
-						{{ t('tables', 'I really want to transfer this table!') }}
-					</NcButton>
```
enjeck: Not sure if we want to keep this and make the user confirm when transfering a table. I read  that it's good UX to have such a confirmation when a user is doing a destructive or irreversible action, which maybe applies here?

src/modules/modals/TransferTable.vue
```diff
@@ -0,0 +1,118 @@
+<template>
+	<NcModal v-if="showModal"
+		size="normal"
+		@close="actionCancel">
+		<div class="modal__content" data-cy="transferTableModal">
+			<div class="row">
+				<div class="col-4">
+					<h2>{{ t('tables', 'Transfer table') }}</h2>
+				</div>
```
enjeck, line 9: Do we have to add new text like this to the l10n folder?

juliushaertl: No need to, once merged, there is a nightly automation job that will push the newly added strings for translation to transifex which is what we use for translators. Then once the text is translated it will be pushed back to the repo automatically.

Full docs for that are in https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html

cypress/e2e/tables-table.cy.js
```diff
@@ -1,11 +1,15 @@
 let localUser
+let user1
```
juliushaertl, line 2: Maybe call it like that to make it more readable what this is about

```suggestion
let targetUserTransfer
```

cypress/e2e/tables-table.cy.js
```diff
@@ -51,4 +55,27 @@ describe('Manage a table', () => {
...
+		cy.get('.action-button__text').contains('Edit table').click()
+
+		cy.get('[data-cy="editTableModal"]').should('be.visible')
+		cy.get('[data-cy="editTableModal"] button').contains('Change owner').click()
+		cy.get('[data-cy="editTableModal"]').should('not.exist')
+		cy.get('[data-cy="transferTableModal"]').should('be.visible')
+		cy.get('[data-cy="transferTableModal"] input[type="search"]').clear().type(user1.userId)
+		cy.get(`.vs__dropdown-menu [user="${user1.userId}"]`).click()
+		cy.get('[data-cy="transferTableButton"]').should('be.enabled').click()
+		cy.wait(10).get('.toastify.toast-success').should('be.visible')
```
juliushaertl, line 78: There should be no need to wait manually as cy.get should wait for the default timeout for the new element to appear
```suggestion
		cy.get('.toastify.toast-success').should('be.visible')
```

cypress/e2e/tables-table.cy.js
```diff
@@ -51,4 +55,27 @@ describe('Manage a table', () => {
...
+		cy.get('.app-navigation__list').contains('test table').should('not.exist')
```
juliushaertl, line 79: Can we also login as the target user and assert that the table list listed there?

src/modules/modals/TransferTable.vue
```diff
@@ -0,0 +1,118 @@
...
+			}
+		},
+	},
+	methods: {
+		actionCancel() {
+			this.reset()
+			this.$emit('close')
+		},
+		reset() {
+		},
```
juliushaertl, line 97: Can be removed :)

src/shared/components/ncTable/mixins/permissionsMixin.js
```diff
@@ -2,6 +2,9 @@ import { getCurrentUser } from '@nextcloud/auth'
 
 export default {
 	methods: {
+		ownsTable(element) {
+			return element?.ownership === getCurrentUser().uid
```
juliushaertl, line 6: Just from reading the code i think this should be:

```suggestion
			return element?.owner === getCurrentUser().uid
```

src/modules/modals/TransferTable.vue
```diff
@@ -0,0 +1,118 @@
...
+						:display-name="localTable.ownerDisplayName"
+						:user="localTable.owner" />
+				</div>
+			</div>
+			<div class="row">
+				<NcUserAndGroupPicker :select-users="true" :select-groups="false" :new-owner-user-id.sync="newOwnerUserId" />
+			</div>
+			<div class="row">
+				<div class="fix-col-4 space-T justify-between">
+					<NcButton type="warning" :disabled="newOwnerUserId === ''" data-cy="transferTableButton" @click="transferMe">
```
juliushaertl, line 30: Maybe we can right align the button as the primary action is usually on the right.

src/shared/components/ncUserAndGroupPicker/NcUserAndGroupPicker.vue
```diff
@@ -0,0 +1,186 @@
+<template>
+	<div class="row space-B">
+		<h3>{{ t('tables', 'Transfer this table to another user') }}</h3>
```
juliushaertl, line 3: Can we move this part out to the TransferTable component? Then the NcUseraAndGroupPicker is more generic and could be reused in other places

src/shared/components/ncUserAndGroupPicker/NcUserAndGroupPicker.vue
```diff
@@ -0,0 +1,186 @@
...
+			} else if (this.selectGroups) {
+				return [
+					this.SHARE_TYPES.SHARE_TYPE_GROUP,
+				]
+			} else {
+				return []
+			}
+		},
+
+		async asyncFind(query) {
```
juliushaertl, line 120: We could call `asyncFind('')` in the beforeMount hook to show some users already when opening the select input for the first time.

src/shared/components/ncUserAndGroupPicker/NcUserAndGroupPicker.vue
```diff
@@ -0,0 +1,186 @@
...
+					}
+				})
+
+				this.suggestions = this.filterOutCurrentUser(rawSuggestions)
+
+				this.loading = false
+
+			}).catch(err => {
+				console.debug(err)
+			})
```
juliushaertl, line 153: Await is not used currently, might be better to wrap with try/catch and actually await the axios result
```suggestion
			try {
				const res = await axios.get(url)
				const rawSuggestions = res.data.ocs.data.map(autocompleteResult => {
					return {
						user: autocompleteResult.id,
						displayName: autocompleteResult.label,
						icon: autocompleteResult.icon,
						isUser: autocompleteResult.source.startsWith('users'),
						key: autocompleteResult.source + '-' + autocompleteResult.id,
					}
				})

				this.suggestions = this.filterOutCurrentUser(rawSuggestions)

				this.loading = fals
			} catch (err) {
				console.debug(err)
				// TODO: Should at least show some error to the user that the request failed
				// e.g. with https://github.com/nextcloud-libraries/nextcloud-dialogs#toasts
				showError(t('tables', 'Failed to fetch users'))
			}
```

src/shared/components/ncUserAndGroupPicker/NcUserAndGroupPicker.vue
```diff
@@ -0,0 +1,186 @@
...
+				return [
+					this.SHARE_TYPES.SHARE_TYPE_USER,
+				]
+			} else if (this.selectGroups) {
+				return [
+					this.SHARE_TYPES.SHARE_TYPE_GROUP,
+				]
+			} else {
+				return []
+			}
```
juliushaertl, line 117: Might be a bit simpler like this:

```suggestion
			const types = []
			if (this.selectUsers) {
				types.push(this.SHARE_TYPES.SHARE_TYPE_USER)
			}
			if (this.selectGroups) {
				types.push(this.SHARE_TYPES.SHARE_TYPE_GROUP)
			}
			return types
```

src/shared/components/ncUserAndGroupPicker/NcUserAndGroupPicker.vue
```diff
@@ -0,0 +1,186 @@
...
+				try {
+					if (item.isUser && item.user === getCurrentUser().uid) {
+						return arr
+					}
+					arr.push(item)
+				} catch {
+					return arr
+				}
+				return arr
+			}, [])
```
juliushaertl, line 174: I think this could be simplified a lot with just using `list.filter` to remove the current user

juliushaertl (changes requested):
Thanks @enjeck. I left a few inline comments regarding smaller code suggestions, but apart from that this looks quite nice already and works well :)

juliushaertl:
Regarding the CI failure, you might want to check the details at https://github.com/nextcloud/tables/pull/741/checks?check_run_id=19976327121 on how to solve it and make sure to signoff commits in the future:

Screenshot 2023-12-27 at 15 35 36

enjeck:
> Regarding the CI failure, you might want to check the details at https://github.com/nextcloud/tables/pull/741/checks?check_run_id=19976327121 on how to solve it and make sure to signoff commits in the future:

For some reason, trying to sign from VS Code doesn't work. I thought I set this up already, but apparently not :/. I'll try again

src/shared/components/ncTable/mixins/permissionsMixin.js
enjeck: I double cheched and it's `ownership`

src/shared/components/ncUserAndGroupPicker/NcUserAndGroupPicker.vue
enjeck: @juliushaertl I'm confused about how this helps? Querying an empty string doesn't produce any results

src/shared/components/ncTable/mixins/permissionsMixin.js
juliushaertl: All right, thanks for double checking

src/shared/components/ncUserAndGroupPicker/NcUserAndGroupPicker.vue
juliushaertl: Ok, then I remembered that wrong from some other endpoint that used to present suggestions for users. Let's skip that then here.

src/shared/components/ncUserAndGroupPicker/NcUserAndGroupPicker.vue
```diff
@@ -0,0 +1,179 @@
...
+		},
+
+		async getSuggestions(search) {
+			this.loading = true
+			const shareTypes = this.getShareTypes()
+			let shareTypeQueryString = ''
+			shareTypes.forEach(shareType => {
+				shareTypeQueryString += `&shareTypes[]=${shareType}`
+			})
+			const url = generateOcsUrl('core/autocomplete/get?search={searchQuery}&itemType=%20&itemId=%20{shareTypeQueryString}&limit={limit}', { searchQuery: search, shareTypeQueryString, limit: this.maxAutocompleteResults })
```
enjeck, line 140: For some reason, the shareType doesn't work. No matter what I use, it only loads users. @juliushaertl

src/shared/components/ncUserAndGroupPicker/NcUserAndGroupPicker.vue
```diff
@@ -0,0 +1,177 @@
...
+						isUser: autocompleteResult.source.startsWith('users'),
+						key: autocompleteResult.source + '-' + autocompleteResult.id,
+					}
+				})
+
+				this.suggestions = this.filterOutCurrentUser(rawSuggestions)
+				this.loading = false
+			} catch (err) {
+				console.debug(err)
+				showError(t('tables', `Failed to fetch ${this.getShareTypeString().toLowerCase()}`))
```
enjeck, line 156: Reminding myself to separate this for translation purposes

src/shared/components/ncUserAndGroupPicker/NcUserAndGroupPicker.vue
enjeck: My bad. Wasn't concatenating properly. Fixed now

juliushaertl (approved):

juliushaertl:
Could you rebase instead of adding a merge commit? Let me know if you need some guidance on that ;:)

PR #744, opened 2023-12-18
fix: only return values that belong to a view after updating a row fr…

### LINKED ISSUE #743: Toggling a boolean forth and back in a view causes `"Column with id 965 is not part of view with id 12"`
### Steps to reproduce

1. Open view for table with boolean field.
2. Edit one row.
3. Toggle the boolean and save.
4. Edit again, toggle back and save again.
5. See two error messages and a 500 response

grafik

### Expected behavior

Can toggle boolean back and forth

### Actual behavior

Error message displayed

### Tables app version

0.6.4

### Browser

Firefox 120

### Client operating system

Linux

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

Observed on cloud.nextcloud.com

datenangebot:
…om within a view

closes #743

### DIFF
--- a/lib/Db/RowMapper.php
+++ b/lib/Db/RowMapper.php
@@ -402,4 +402,26 @@ class RowMapper extends QBMapper {
 			return 0;
 		}
 	}
+
+	/**
+	 * @param int $id
+	 * @param View $view
+	 * @return Row
+	 * @throws DoesNotExistException
+	 * @throws Exception
+	 * @throws MultipleObjectsReturnedException
+	 */
+	public function findByView(int $id, View $view): Row {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('t1.*')
+			->from($this->table, 't1')
+			->where($qb->expr()->eq('id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT)));
+		$row = $this->findEntity($qb);
+
+		$row->setDataArray(array_filter($row->getDataArray(), function ($item) use ($view) {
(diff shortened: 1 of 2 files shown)


lib/Db/RowMapper.php
```diff
@@ -402,4 +402,26 @@ public function countRows(int $tableId): int {
...
+	 * @return Row
+	 * @throws DoesNotExistException
+	 * @throws Exception
+	 * @throws MultipleObjectsReturnedException
+	 */
+	public function findByView(int $id, View $view): Row {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('t1.*')
+			->from($this->table, 't1')
+			->where($qb->expr()->eq('id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT)));
```
enjeck, line 418: What does `t1` here mean?

datenangebot: this is a named alias for the requested table / data set.
In this combination it is not that important, but sometimes this is needed to reference to a subquery result like this: https://github.com/nextcloud/tables/pull/744/files/7e90043c2961025b8ef6bed31b8ad6f13da4d0e3#diff-f49b633aeda8a873a507dfffbf837e9aa3446bb6c992174749e66827997ab901L335

max-nextcloud (approved):
Only reviewed the code without testing it.

But from the code review it looks like it might fix the issue we discovered.
Maybe add a test to prevent regressions?
This seems like a likely candidate to get optimized away later.

Issue #702, opened 2023-11-21
Error hints for mandatory fields


datenangebot:
### Is your feature request related to a problem? Please describe.

If a mandatory field is empty, you get a notification if you try to save that one.
But the notification is global, so you don't now which mandatory fields (assuming there are more than one) throws error(s).

### Describe the solution you'd like

Mark the related field(s) with a red border etc.

juliushaertl:
UI wise i think it would be nice to just add our NcNoteCard component next to the field that has the error https://nextcloud-vue-components.netlify.app/#/Components/NcNoteCard

PR #755, opened 2023-12-27
feat: add error hints for mandatory fields

enjeck:
Addresses https://github.com/nextcloud/tables/issues/702

### DIFF
--- a/src/modules/modals/CreateRow.vue
+++ b/src/modules/modals/CreateRow.vue
@@ -24,9 +24,9 @@
 							{{ t('tables', 'Add more') }}
 						</NcCheckboxRadioSwitch>
 					</div>
-					<button v-if="!localLoading" class="primary" :aria-label="t('tables', 'Save row')" :disabled="hasEmptyMandatoryRows" @click="actionConfirm()">
+					<NcButton v-if="!localLoading" class="primary" :aria-label="t('tables', 'Save row')" :disabled="hasEmptyMandatoryRows" @click="actionConfirm()">
 						{{ t('tables', 'Save') }}
-					</button>
+					</NcButton>
 				</div>
 			</div>
 		</div>
@@ -34,7 +34,7 @@
 </template>
 
 <script>
-import { NcModal, NcCheckboxRadioSwitch, NcNoteCard } from '@nextcloud/vue'
+import { NcModal, NcCheckboxRadioSwitch, NcNoteCard, NcButton } from '@nextcloud/vue'
 import { showError, showSuccess } from '@nextcloud/dialogs'
 import '@nextcloud/dialogs/dist/index.css'
 import { mapGetters } from 'vuex'
@@ -47,6 +47,7 @@ export default {
 		ColumnFormComponent,
 		NcCheckboxRadioSwitch,
 		NcNoteCard,
+		NcButton,
 	},
 	props: {
 		showModal: {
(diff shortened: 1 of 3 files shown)


enjeck:
Looks like so:
Screenshot from 2023-12-27 17-54-48

src/modules/modals/CreateRow.vue
```diff
@@ -85,25 +100,14 @@ export default {
...
 		},
 		async actionConfirm() {
-			let mandatoryFieldsEmpty = false
-			this.columns.forEach(col => {
-				if (col.mandatory) {
-					const validValue = this.isValueValidForColumn(this.row[col.id], col)
-					mandatoryFieldsEmpty = mandatoryFieldsEmpty || !validValue
-				}
-			})
-			if (!mandatoryFieldsEmpty) {
```
enjeck, line 95: Since we now ensure that the "Save" button is disabled if there are empty mandatory fields, I assume it's safe to remove this check

src/modules/modals/CreateRow.vue
```diff
@@ -12,6 +12,10 @@
 				<ColumnFormComponent
 					:column="column"
 					:value.sync="row[column.id]" />
+				<NcNoteCard v-if="column.mandatory && !isValueValidForColumn(row[column.id], column)"
+					type="error">
+					{{ t('tables', `"${column.title}" should not be empty`) }}
```
juliushaertl, line 17:
```suggestion
					{{ t('tables', '"{columnTitle}" should not be empty', { columnTitle: column.title }) }}
```

Placeholders need to passed in separately for translations so they can be handled by the translators.

src/modules/modals/CreateRow.vue
```diff
@@ -20,7 +24,7 @@
 							{{ t('tables', 'Add more') }}
 						</NcCheckboxRadioSwitch>
 					</div>
-					<button v-if="!localLoading" class="primary" :aria-label="t('tables', 'Save row')" @click="actionConfirm()">
+					<button v-if="!localLoading" class="primary" :aria-label="t('tables', 'Save row')" :disabled="hasEmptyMandatoryRows" @click="actionConfirm()">
```
juliushaertl, line 27: Can we switch to the NcButton component once we're at?

src/modules/modals/EditRow.vue
```diff
@@ -12,6 +12,10 @@
 				<ColumnFormComponent
 					:column="column"
 					:value.sync="localRow[column.id]" />
+				<NcNoteCard v-if="column.mandatory && !isValueValidForColumn(localRow[column.id], column)"
+					type="error">
+					{{ t('tables', `"${column.title}" should not be empty`) }}
```
juliushaertl, line 17: Same here with the placeholder

src/modules/modals/CreateRow.vue
enjeck: I don't know what "passed in separately" means here

juliushaertl: The suggested code change shows how to do it.

The idea is that the `t()` function will do the parameter replacement instead of directly doing it in the string passed over. That way translators will be able to see and translate the text with the placeholder value still preserved and it is only inserted at runtime.

juliushaertl: https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#javascript

enjeck: I see, thanks

juliushaertl (approved):

juliushaertl:
A small test case would be nice, but doing it in cypress e2e tests seems a bit overkill and we do not have frontend unit testing or component testing ready in the tables app yet and not sure how well those components are testable with that in its current state, so maybe still just stick to a simple cypress test.

juliushaertl (approved):

PR #756, opened 2023-12-28
chore: Avoid pulling in conflicting dependencies

juliushaertl:
This allows to run the app (e.g. for running behat tests) without the requirement to only install runtime dependencies. Previously vimeo/psalm pulled in symfony/console in a different version then what the server ships, so using the phar bundle instead isolates their dependencies.

Also removes phan/phan which was not used


juliushaertl:
CI failure is unrelated (ref https://github.com/nextcloud/server/pull/42504)

enjeck (approved):
cool

Issue #751, opened 2023-12-21
Smart picker modal select input is too small


juliushaertl:
<img width="637" alt="Screenshot 2023-12-21 at 19 27 45" src="https://github.com/nextcloud/tables/assets/3404133/b9bcb6bb-affc-4a46-8741-37f453d3c607">

Steps to reproduce:
- Open a text file
- Use the smart picker with typing `/table` to insert a table reference to the document

juliushaertl:
@enjeck Maybe you can check that one?

enjeck:
I can't reproduce this. Looks fine for me. 
![Screenshot from 2023-12-29 15-51-26](https://github.com/nextcloud/tables/assets/32180937/f0887069-3ba0-4726-9a2b-c7538c35b859)

juliushaertl:
The issue seems only after you selected something (the active entry shown in the collapsed select input)

PR #760, opened 2023-12-29
fix: increase smartpicker select height

enjeck:
Addresses: https://github.com/nextcloud/tables/issues/751

### DIFF
--- a/src/views/partials/SearchAndSelectOption.vue
+++ b/src/views/partials/SearchAndSelectOption.vue
@@ -110,12 +110,19 @@ export default {
 		display: block;
 	}
 
+	.content .labels {
+		font-size: 1.3em;
+		font-weight: bold;
+		margin-bottom: 8px;
+	}
+
 	.content .details {
 		display: inline-flex;
 	}
 
 	:deep(.counter-bubble__counter) {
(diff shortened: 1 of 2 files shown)


enjeck:
Before:
Screenshot from 2023-12-29 18-07-37

After:
Screenshot from 2023-12-29 18-02-57

enjeck:
Here's what it looks like now:
Screenshot from 2024-01-03 09-03-45

enjeck:
> Yeah, that looks so much better. 😁 If you want to do it perfect, you should use the default value for spacings: https://github.com/nextcloud/server/blob/89ee85938e2c66e8d1c20f57fa754b72509e524f/apps/theming/css/default.css#L57
> 
> Something like `padding: calc(var(--default-grid-baseline) * 1)`

There are other conflicting styles here, so modifying the padding here doesn't fix this

enjeck:
Looks good. But wondering if we want to only apply the padding after selection (or if this is even possible). Right now, the empty search bar looks massive, and maybe this is awkward?:
image

enjeck:
@datenangebot Good point, thanks.

juliushaertl (approved):

PR #768, opened 2024-01-09
fix(i18n): Fixed grammar

rakekniven:
Reported at Transifex.


juliushaertl:
In server we seem to mostly use administrator. Should this be a general adjustment? Was there some discussion around that?

juliushaertl (approved):

PR #779, opened 2024-01-18
Avoid failures with link column database values from previous versions

### LINKED ISSUE #777: Updating rows removes the github url column value
### Steps to reproduce

1. have a table with rows and a column for GH URLs
2. update a boolean value from false to true
3. save

### Expected behavior

Data is saved as before but with the changed state/value for the boolean column

### Actual behavior

Boolean change is saved, yet GH url is wiped empty

follow up:
1. open row again for editing
2. read url value
3. save
4. see screenshot
840-292-max

### Tables app version

0.6.5

### Browser

Firefox 121.0.1

### Client operating system

WinPro11

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

c.nc.c - supported apps view (🔎 Hub 7 release missing) -> table (🚒 Supported Apps)

juliushaertl:
Fix #777 
- fix(link): Properly sanizite and map database values to what we expect it to be
- fix: Avoid failing on empty strings for URL values

### DIFF
--- a/lib/Service/ColumnTypes/TextLinkBusiness.php
+++ b/lib/Service/ColumnTypes/TextLinkBusiness.php
@@ -12,18 +12,30 @@ class TextLinkBusiness extends SuperBusiness implements IColumnTypeBusiness {
 	 * @return string
 	 */
 	public function parseValue($value, ?Column $column = null): string {
+		if ($value === null) {
+			return '';
+		}
+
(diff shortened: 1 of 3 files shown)


juliushaertl:
Pushed a unit test as well to cover all cases that might occur

Issue #775, opened 2024-01-15
Using "select all" on filtered list also select hidden entries


Art4:
### Steps to reproduce

1. Filter a table or use the search on a table
2. Click the "select all" check box

### Expected behavior

Only the shown entries should be selected.

### Actual behavior

All entries (even the hidden entries) are selected.

### Tables app version

0.6.5

### Browser

Firefox 121

### Client operating system

Windows

### Operating system

Linux

### Web server

Apache

### PHP engine version

PHP 8.2

### Database

MariaDB

### Additional info

scenario: I want to export all vacation requests of user "alice". I use the search typing "alice" but can not use the "select all" check box, because this will select also the hidden entries.

![grafik](https://github.com/nextcloud/tables/assets/2162994/51416f96-7046-430a-bdd9-6b1727212d05)

I have the same problem if I'm using the filter in a column.

![grafik](https://github.com/nextcloud/tables/assets/2162994/8df691f3-06bd-4399-ba9a-6d211750bce0)

juliushaertl:
@nextcloud/designers What's your input on this one? I'd say we should probably limit the selected entires to only match the visible (filtered ones).

szaimen:
> I'd say we should probably limit the selected entires to only match the visible (filtered ones).

yes, agreed :)

PR #782, opened 2024-01-19
fix: select all only selects filtered list

enjeck:
Fixes https://github.com/nextcloud/tables/issues/775


juliushaertl (approved):
Tested and works 👍

PR #783, opened 2024-01-19
Create appstore-build-publish.yml

juliushaertl:
For building releases in the github action

### DIFF
--- a/.github/workflows/appstore-build-publish.yml
+++ b/.github/workflows/appstore-build-publish.yml
@@ -1,0 +1,169 @@
+# This workflow is provided via the organization template repository
+#
+# https://github.com/nextcloud/.github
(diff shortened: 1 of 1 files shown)


.github/workflows/appstore-build-publish.yml
```diff
@@ -0,0 +1,169 @@
+# This workflow is provided via the organization template repository
+#
+# https://github.com/nextcloud/.github
+# https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization
```
enjeck, line 4: Everything here is just an exact copy of what's at the template repo. I expected some lines will be changed to make the template suitable for Tables specifically? :thinking:

Issue #474, opened 2023-08-18
Horizontal scrollbar missing if not scrolled down Verticaly


NoahMarch:
### Steps to reproduce

Have a Table exceed both with and length the screensize

### Expected behavior

If not all Columnes are fitted on current view show the horizontal Scrollbar

### Actual behavior

![Pic1](https://github.com/nextcloud/tables/assets/112391811/9da1a283-cfee-4d6f-ada4-9fb5a70273ce)
As seen in Pic1 horizontal scrollbar isnt shown, even when not all Columnes are displayed

The horizontal scrollbar is shown once scrolled all the way down vertical:
![Pic2](https://github.com/nextcloud/tables/assets/112391811/874dd2a1-f0db-4d31-af41-64b0612eade1)

### Tables app version

0.5.1

### Browser

Chrome 115.0.5790.171

### Client operating system

Windows 10

### Web server

None

### PHP engine version

None

### Database

None

NoahMarch:
If someone wants to Test heres a CSV with the sample data to import (Note: since working from Germany, the import may not work for everyone because of ";" and not "," as seperators, maybe depends on your next cloud settings?) 
[MappeTest.CSV](https://github.com/nextcloud/tables/files/12378208/MappeTest.CSV)

datenangebot:
Thanks for providing teh test file.
Issue fix seems to go technical deep into the main table presentation.
We have to invest some time and a lot of testing.
Move it to next milestone.

Issue #35, opened 2022-03-29
Set/rememeber/save column width


justinlimbo:
If I change the column width it would be nice if this setting would be remembered for the next time I open the table. 
Maybe the column width could be in the column settings.

leex279:
Yes, that would be very nice. First thing has come to my attention after filling in some data.

datenangebot:
thanks for the feedback, I checked the technical details and there should be no problem.
Pushed to the roadmap.

datenangebot:
this is related to the views feature that will come with v1

PR #786, opened 2024-01-19
fix: Make table header sticky

juliushaertl:
Slightly different approach for https://github.com/nextcloud/tables/pull/342 to contribute to #631 making the table column header sticky.

Ideally we would have a combination of absolute positioning and stickyness to let the sticky container only span the outer width of the scroll container, but that seems to be impossible so I went for a resize observer that pins the width of the sticky table header and table options through a css variable which turned out to work reasonably well.

> refactor the table as css div table
> prepare for column resizing https://github.com/nextcloud/tables/issues/35

@datenangebot I don't remember exactly if the points in the old pr are still valid, but considering the refactor would still be valid, this seems like a good intermediate solution for browsing larger tables.

Kapture 2024-02-14 at 09 50 43
Kapture 2024-02-14 at 09 36 58

Also fixes https://github.com/nextcloud/tables/issues/474

## ToDo

- [x] Smart picker content widget scrolling needs fixing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "@tiptap/extension-task-list": "^2.2.2",
         "@tiptap/starter-kit": "^2.2.2",
         "@tiptap/vue-2": "^2.2.2",
+        "@vueuse/core": "^10.7.2",
         "debounce": "^1.2.1",
         "vue": "^2.7.16",
         "vue-material-design-icons": "^5.3.0",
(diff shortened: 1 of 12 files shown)


juliushaertl:
Not it was not directly related. I think this should also work with multiple tables on the same page where sticky headers from lower tables should stack on top of the upper header. Though would not exclude that we need some additional special handling for that (especially since the tables might have different width)

datenangebot (approved):
Reasonable

enjeck:
Can we merge this? @juliushaertl

juliushaertl:
I still need to figure out how to do the scrolling with the link previews 🙈 Will try to get that done the next days

juliushaertl:
Ready from my side, maybe you can have another review @enjeck 

Screen recordings added to the pull request description.

enjeck (approved):
cool :sunglasses:

quentinDupont:
Awesome job, I was looking exactly at this issue, thanks for solving it !

PR #787, opened 2024-01-19
feat: Use simple input field instead of multiselect for plain URLs

juliushaertl:
When no link provider (except URL) is enabled a multiselect field might be confusing to users, so this PR switches over to a simple input field in that case. As there are no suggestions then a multiselect does not make sense.

- In addition I added a copy and open link button to make it easier to access the link again (fixes https://github.com/nextcloud/tables/issues/1186)
- Move clipboard logic into a mixin so it can be reused
- Render a link preview in the edit modal

Styling of the multiselect is still a bit odd, but will check that separately as it seems to come from the component library.

[Kapture 2024-08-22 at 15.54.06.webm](https://github.com/user-attachments/assets/ced0c6dd-5f13-4ca1-9a61-ac76ff63eaa4)

### DIFF
--- a/cypress/e2e/column-text-link.cy.js
+++ b/cypress/e2e/column-text-link.cy.js
@@ -37,8 +37,7 @@ describe('Test column text-link', () => {
 
 		cy.loadTable('Test text-link')
 		cy.get('.NcTable').contains('Create row').click({ force: true })
-		cy.get('.modal__content .slot input').first().type('https://nextcloud.com').tick(500)
-		cy.get('.icon-label-container .labels').contains('https://nextcloud.com').click()
+		cy.get('.modal__content .slot input').first().type('https://nextcloud.com')
 
 		cy.intercept({ method: 'GET', url: '**/search/providers/files/*' }).as('filesResults')
 		cy.get('.modal__content .slot input').eq(1).type('pdf').tick(500)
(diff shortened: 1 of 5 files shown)


enjeck (commented):
Looks good except for the failing Cypress tests. Pushed a commit for this

enjeck:
Weird that the test still fails on some stable branches :thinking:

juliushaertl:
Sorry I should have described this a bit more. The goal of this PR is purely to make the input more user friendly if only URL is enabled as a provider for a column, as in this case the multiselect does not give any benefit as there is only one option to select (which is inserting the plain URL).

I agree we should still fetch the title and store/render it for additional context. Will look into that.

enjeck:
@juliushaertl Do we still need this?

juliushaertl:
Ok that has been a while 🙈 

Finally pushed some more adjustments here all summarized with a screen recording in the first post.

juliushaertl:
I failed to properly adapt the NcSelect to a dynamic height for the selected-option slot, so I decided for now to just stick to a standard height and remove the link type when showing the current selection. Should be not relevant anyways as we have a more detailed link preview:

Screenshot 2024-08-30 at 08 28 14


Screenshot 2024-08-30 at 08 28 08

Issue #648, opened 2023-11-05
Can't create a filter for a multi selection field


schiessle:
### Steps to reproduce

1. Create a table with a "multiple selection" field
2. Create a view and try to set a filter for the field

### Expected behavior

I can set a filter with the two oeprators 
- "equal" filter the exact same selection
-  "contains" filter where the configured value is matched

### Actual behavior

No operator available and therefore I can't set a filter.

tested on c.nc.com

### Tables app version

0.6.0

### Web server

None

### PHP engine version

None

### Database

None

schiessle:
I just notice that there is also no filter on multi select columns possible within a view.

enjeck:
Fixed at https://github.com/nextcloud/tables/pull/798

PR #798, opened 2024-01-24
enh: add filtering in views for multi selection

datenangebot:
With the new db structure it is much easier to filter for selection values inside of multi selection columns. This should make this possible to filter for `contains` and `equal`.  

Not for `empty` for now. Needs to look into the `empty` filter in general...

related to: https://github.com/nextcloud/tables/issues/648

### DIFF
--- a/cypress/e2e/view-filtering-selection.cy.js
+++ b/cypress/e2e/view-filtering-selection.cy.js
@@ -1,0 +1,371 @@
+let localUser
+
+describe('Filtering in a view by selection columns', () => {
+
+	before(function() {
+		cy.createRandomUser().then(user => {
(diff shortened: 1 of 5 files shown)


enjeck (commented):
Cool, I tested it out and it works well :+1: . 

Do we intentionally not have any Cypress tests for View filters?

enjeck (commented):
Looks good to me in general, except that most of the selectors are based on class names that are too generic and highly prone to change. There are so many changes, making it a tedious task to replace the selections with "data-cy" attributes. I can help out with replacing them, if you think it's valuable.

enjeck (approved):

PR #800, opened 2024-01-25
WIP - feat: support multiple tables in store

enjeck:
We want to have multiple tables on the same page

### DIFF
--- a/src/store/data.js
+++ b/src/store/data.js
@@ -4,36 +4,57 @@ import displayError from '../shared/utils/displayError.js'
 import { parseCol } from '../shared/components/ncTable/mixins/columnParser.js'
 import { MetaColumns } from '../shared/components/ncTable/mixins/metaColumns.js'
 import { showError } from '@nextcloud/dialogs'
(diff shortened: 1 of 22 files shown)


src/store/data.js
```diff
@@ -164,11 +193,13 @@ export default {
...
+			if (tableId) {
+				const row = res.data
+				const rows = state.rows[tableId]
+				const index = rows.findIndex(r => r.id === row.id)
+				rows[index] = row
+				commit('setRows', {tableId, rows:[...rows]})
+			}
 			return true
 		},
 		async insertNewRow({ state, commit, dispatch }, { viewId, tableId, data }) {
```
enjeck, line 205: Need to figure out how to handle Views. Using the tableId as a key for state.columns and state.rows doesn't really work

enjeck: We can't just use the View id as a key, since it might collide with a Table id

enjeck:
Will split into multiple PRs. Tracking at https://github.com/nextcloud/tables/issues/807

Issue #807, opened 2024-01-29
Applications: Enable multiple tables instances at the same time


enjeck:
### Is your feature request related to a problem? Please describe.

As part of https://github.com/nextcloud/tables/issues/311, we need to have multiple tables on same page. 

### Describe the solution you'd like

Refactor the global store to hold multiple tables by  using the table ids as a key when storing rows and columns. Also, we need to ensure that modals are targeting the right table/view. 

### Additional context

https://github.com/nextcloud/tables/issues/311#issuecomment-1790569092

enjeck:
Lots of changes required here, so will split this across multiple PRs to make easier to review

PR #808, opened 2024-01-29
enh: Use table id for events

enjeck:
As part of https://github.com/nextcloud/tables/issues/807. We pass table ids around, for use later to support multiple table instances


enjeck:
The reason that I'm passing the table ids around is so that they can be used when emitting events. Currently, modals (such as [CreateColumn.vue](https://github.com/nextcloud/tables/blob/main/src/modules/modals/CreateColumn.vue) need to know exactly what Table to target, and this is gotten from `activeElement` stored in the global store. So by passing the table ids around like I've done, we can always specify an active Table and it will work properly, even when there are multiple tables on screen. 

We're trying to make NcTable independent, so that it may be moved into the component library someday. So passing the table id as prop in NcTable might jeopardize this. 

Not sure if there's a better way (that I understand) to do this.

src/shared/components/ncTable/sections/CustomTable.vue
```diff
@@ -215,10 +219,10 @@ export default {
 	},
 
 	mounted() {
-		subscribe('tables:selected-rows:deselect', this.deselectAllRows)
+		subscribe('tables:selected-rows:deselect', tableId => { this.deselectAllRows(tableId) })
 	},
 	beforeDestroy() {
-		unsubscribe('tables:selected-rows:deselect', this.deselectAllRows)
+		unsubscribe('tables:selected-rows:deselect', tableId => { this.deselectAllRows(tableId) })
```
enjeck, line 225: Realizing that this is inefficient, and will result in too many events if there are multiple tables. Thinking subscribing on the parent instead and using event delegation instead

src/modules/main/sections/MainWrapper.vue
```diff
@@ -84,16 +85,26 @@ export default {
 	},
 
 	methods: {
+		setActiveElement() {
```
juliushaertl, line 88: I'm not sure if having a global state for the current active element is not causing us troubles later if we consider asynchronous operations. Thinking about a request that is performed and in the meantime the user interacts with another view or table already which changes the active id.

What do you think about just using the tableId or viewId as a parameter in the events?

For example:

```
emit('tables:sidebar:sharing', { viewId: this.element.id, open: true, tab: 'sharing' })
```

Or even more flexible something like:

```
emit('tables:sidebar:sharing', { elementType: this.isView ? 'view' : 'table', elementId: this.element.id, open: true, tab: 'sharing' })
```

enjeck: Yep, I didn't consider asynchronous operations. Good catch.

src/shared/components/ncTable/NcTable.vue
```diff
@@ -225,15 +231,17 @@ export default {
 		},
 	},
 	mounted() {
-		subscribe('tables:selected-rows:deselect', this.deselectRows)
+		subscribe('tables:selected-rows:deselect', tableId => this.deselectRows(tableId))
```
juliushaertl, line 234: I think we also need to take `isView` into account here to not run into a conflict when there are views and tables with the same id

enjeck: Good catch, thanks

src/modules/main/partials/TableView.vue
```diff
@@ -110,22 +112,22 @@ export default {
...
-			emit('tables:column:edit', column)
+			emit('tables:column:edit', { column, isView: this.isView, elementId: this.element.id })
 		},
 		deleteColumn(column) {
-			emit('tables:column:delete', column)
+			emit('tables:column:delete', { column, isView: this.isView, elementId: this.element.id })
 		},
 		createRow() {
-			emit('tables:row:create', this.columns)
+			emit('tables:row:create', { column: this.columns, isView: this.isView, elementId: this.element.id })
```
juliushaertl, line 124:
```suggestion
			emit('tables:row:create', { columns: this.columns, isView: this.isView, elementId: this.element.id })
```

src/modules/modals/Modals.vue
```diff
@@ -109,26 +113,30 @@ export default {
...
-		subscribe('tables:column:delete', column => { this.columnToDelete = column })
+		subscribe('tables:column:create', columnInfo => { this.createColumnInfo = columnInfo })
+		subscribe('tables:column:edit', columnInfo => { this.columnToEdit = columnInfo })
+		subscribe('tables:column:delete', columnInfo => { this.columnToDelete = columnInfo })
 
 		// rows
-		subscribe('tables:row:create', columns => { this.columnsForRow = columns })
-		subscribe('tables:row:edit', row => { this.editRow = row })
-		subscribe('tables:row:delete', rows => { this.rowsToDelete = rows })
+		subscribe('tables:row:create', columnsInfo => { this.columnsForRow = columnsInfo })
```
juliushaertl, line 121: With the adjustment where this is emitted it should probably be rather this:
```suggestion
		subscribe('tables:row:create', ({ columns }) => { this.columnsForRow = columns })
```

juliushaertl: Similar adjustments we'd need for the other ones

enjeck: I don't think this works, since we want the entire object, not just the `columns` property

juliushaertl: Sorry I somehow missed the updated usage of it, nevermind about the earlier comment

juliushaertl (approved):

PR #821, opened 2024-02-04
enh: show cropped calendar popup

enjeck:
Calendar was hidden if the modal wasn't long enough. This PR ensures it isn't cropped. 

###  Before
Screenshot from 2024-02-04 11-45-28

### After
Screenshot from 2024-02-04 11-12-31


juliushaertl:
I think this approach would cause the overflow scrolling of modals to break but should not longer be necessary with the migration to the native date time picker with #647

enjeck:
After https://github.com/nextcloud/tables/pull/647, we no longer need this :)

Issue #827, opened 2024-02-08
Row2Mapper::getFilterExpression(): Argument #2 ($column) must be of type Column, null given


ChristophWurst:
### Steps to reproduce

1. Open the app
2. Navigate to a view

### Expected behavior

I see the view

### Actual behavior

> Could not load rows. Unknown error.

error toast shown in the UI.

```json
{
  "reqId": "abcdefg",
  "level": 3,
  "time": "2024-02-08T08:18:44+00:00",
  "remoteAddr": "a:b:c",
  "user": "christoph",
  "app": "index",
  "method": "GET",
  "url": "/apps/tables/row/view/51",
  "message": "OCA\\Tables\\Db\\Row2Mapper::getFilterExpression(): Argument #2 ($column) must be of type OCA\\Tables\\Db\\Column, null given, called in /var/www/nextcloud/apps/tables/lib/Db/Row2Mapper.php on line 275 in file '/var/www/nextcloud/apps/tables/lib/Db/Row2Mapper.php' line 299",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64; rv:120.0) Gecko/20100101 Firefox/120.0",
  "version": "28.0.2.5",
  "exception": {
    "Exception": "Exception",
    "Message": "OCA\\Tables\\Db\\Row2Mapper::getFilterExpression(): Argument #2 ($column) must be of type OCA\\Tables\\Db\\Column, null given, called in /var/www/nextcloud/apps/tables/lib/Db/Row2Mapper.php on line 275 in file '/var/www/nextcloud/apps/tables/lib/Db/Row2Mapper.php' line 299",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/nextcloud/lib/private/AppFramework/App.php",
        "line": 184,
        "function": "dispatch",
        "class": "OC\\AppFramework\\Http\\Dispatcher",
        "type": "->",
        "args": [
          [
            "OCA\\Tables\\Controller\\RowController"
          ],
          "indexView"
        ]
      },
      {
        "file": "/var/www/nextcloud/lib/private/Route/Router.php",
        "line": 315,
        "function": "main",
        "class": "OC\\AppFramework\\App",
        "type": "::",
        "args": [
          "OCA\\Tables\\Controller\\RowController",
          "indexView",
          [
            "OC\\AppFramework\\DependencyInjection\\DIContainer"
          ],
          [
            "51",
            "tables.row.indexView"
          ]
        ]
      },
      {
[...]

PR #828, opened 2024-02-08
fix: Pass all table columns along as we need them for filtering the rows

juliushaertl:
Fixes https://github.com/nextcloud/tables/issues/827

Fix for an internal finding

To reproduce:
- Create a table with a single select and some other fields
- Create a view to filter for the single select but don't include it in the visible columns list
- Load the view

Would be good to cover this case with a test, but pushed to have CI running already to be able to apply a patch as a quick fix

### DIFF
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -26,19 +26,23 @@ class Row2Mapper {
 	private IDBConnection $db;
 	private LoggerInterface $logger;
 	protected UserHelper $userHelper;
+	protected ColumnMapper $columnMapper;
 
 	/* @var Column[] $columns */
 	private array $columns = [];
+	/* @var Column[] $columns */
+	private array $allColumns = [];
 
 	private ColumnsHelper $columnsHelper;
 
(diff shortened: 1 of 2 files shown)


enjeck:
I'm trying to reproduce the error. What does "Create a view to filter for the single select but don't include it in the visible columns list" mean?

juliushaertl:
Uncheck the column from the list of columns but still filter for a value (The Test column is a single select with two options):

Screenshot 2024-02-08 at 14 58 03

enjeck:
Oh, I see

lib/Db/Row2Mapper.php
```diff
@@ -267,12 +271,13 @@ private function getFilter(IQueryBuilder &$qb, array $filterGroup): array {
 		$filterExpressions = [];
 		foreach ($filterGroup as $filter) {
 			$columnId = $filter['columnId'];
+			$column = $this->columns[$filter['columnId']] ?? $this->allColumns[$filter['columnId']];
```
enjeck, line 274:
```suggestion
			$column = $this->columns[$columnId] ?? $this->allColumns[$columnId];
```

enjeck (approved):
Tested and works

PR #830, opened 2024-02-08
enh/vite

juliushaertl:
- fix(deps): update dependency @nextcloud/vue to v8
- chore(deps): rename title prop, update comp
- enh: add aria label to NcSelect
- fix: add label to NcSelect
- fix(tests): fix failing tests
- fix(tests): fix test for select
- enh: improve style
- chore: suppress ResizeObserver exception for Cypress
- enh: rename attribute, modify style
- enh: use NcDateTimePickerNative
- enh: fix lint
- perf: Switch to vite for bundling


enjeck:
@julius You say "based on https://github.com/nextcloud/tables/pull/647", but I don't see how?

juliushaertl:
I based my branch locally on top of yours and then created a pull request towards that one. You can also sett that in the github ui:

Screenshot 2024-02-09 at 09 22 56

Once your PR/branch gets merged github automatically changes this, but that way I can make use of your changes already during development.

enjeck:
Doesn't work for me after running `npm ci` and then `npm run dev`. Am I missing something?

max-nextcloud (approved):
Looks great! Thanks for looking into this.

I have some minor comments only.

Locally I get a warming when building:

grafik

I think the `?? ''` can just be removed in this case - but it would also be nice to catch such warnings in the ci or so. However i was not able to trigger that check on its own.

I also noticed that tables have dark favicons for me now:
grafik
I don't know if that's related. Will try and build with main branch to compare.

juliushaertl:
Likely caused by https://github.com/nextcloud/tables/pull/712 then

juliushaertl:
@susnux I'm having some trouble here with the `npm run watch` command which always exits after the first build. Is that something you've seen already? Any clue why that might happen? We just call `vite --mode development build --watch`

susnux:
> @susnux I'm having some trouble here with the `npm run watch` command which always exits after the first build.

Should be fixed now: https://github.com/vitejs/vite/issues/15951

susnux:
> I think the `?? ''` can just be removed in this case - but it would also be nice to catch such warnings in the ci or so. However i was not able to trigger that check on its own.

The warning means that the right part is never null / undefined because `'string' + variable ?? 'fallback'` reads as:
1. `'string' + variable`
2. `??` Then check the result is null or undefined
3. Then fallback to `'fallback'`.

So just brackets are missing like `'string' + (variable ?? 'fallback')`.

(Otherwise you might get `'stringnull'` as the result)

juliushaertl:
@susnux The two issues i mentioned in person now on this PR:

- The file picker dialog loading fails (a quick patch for the dialogs library is below but I'm unsure if that is the proper approach)
- Once that is patched the dialog does not return any selected file

```diff
diff --git a/lib/composables/isPublic.ts b/lib/composables/isPublic.ts
index ec89e8b..a1000ca 100644
--- a/lib/composables/isPublic.ts
+++ b/lib/composables/isPublic.ts
@@ -10,7 +10,7 @@ import { onBeforeMount, ref } from "vue"
 export const useIsPublic = () => {
        const checkIsPublic = () => (document.getElementById('isPublic') as HTMLInputElement|null)?.value === '1'

-       const isPublic = ref(true)
+       const isPublic = ref(checkIsPublic())
        onBeforeMount(() => { isPublic.value = checkIsPublic() })

        return {
```

blizzz:
What is the current state? There are handful of conflicts at least.

.github/workflows/cypress-custom.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [ 16 ]
+        node-version: [ 18 ]
```
juliusknorr, line 33: As discussed lets adapt https://github.com/nextcloud/.github/blob/master/workflow-templates/cypress.yml#L63-L76 to read them from package.json

juliusknorr (commented):
Two remaining comments, the cypress failure i cannot judge if related right away, but fine to fix separately from my side.

juliusknorr:
Cannot self-approve, but feel free to merge on your own then after resolving the conflict and thanks for picking this up and driving it forward 🥳

.github/workflows/cypress-custom.yml
```diff
@@ -92,12 +92,12 @@ jobs:
...
           path: apps/${{ env.APP_NAME }}
 
-      - name: Set up node 16
+      - name: Set up node 18
         uses: actions/setup-node@v3
         with:
           cache: 'npm'
           cache-dependency-path: apps/${{ env.APP_NAME}}/package-lock.json
-          node-version: 16
+          node-version: 18
```
susnux, line 100: if specified in package.json under `engine` you can also use:
```suggestion
          node-version-file: package.json
```

.github/workflows/cypress-custom.yml
juliusknorr: Oh good to know, thanks for sharing that

.github/workflows/node.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           npm ci
           npm run build --if-present
 
-      - name: Check webpack build changes
+      - name: Check build changes
```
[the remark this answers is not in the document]
enjeck: Done at https://github.com/nextcloud/.github/pull/562

PR #840, opened 2024-02-14
enh: use required name prop for NcAppSettingsSection

enjeck:
Need to use `name` instead of `title` prop, as required by nextcloud/vue v8


src/modules/modals/ViewSettings.vue
```diff
@@ -1,10 +1,10 @@
 <template>
 	<NcAppSettingsDialog :open.sync="open" :show-navigation="true" data-cy="viewSettingsDialog" :title="createView ? t('tables', 'Create view') : t('tables', 'Edit view')">
-		<NcAppSettingsSection v-if="columns === null" id="loading" title="">
+		<NcAppSettingsSection v-if="columns === null" id="loading" name="">
```
juliushaertl, line 3: We could enhance that while at it:

```suggestion
		<NcAppSettingsSection v-if="columns === null" id="loading" :name="t('tables', 'Loading columns …')">
```

To have a label for accessibility

juliushaertl (approved):

Issue #812, opened 2024-01-30
RFC: DB Structure for Contexts


blizzz:
For #311, RC

## Structure

```mermaid
---
title: Tables Contexts Database Structure
---
erDiagram

Context ||--o{ Page : contains
Page ||--o{ Page_Content : contains
Context }|--o{ Rel_Context_Resource : contains
Page_Content }o--o{ Rel_Context_Resource : refs
Context ||--o{ Shares : has
Shares ||--o{ Navigation : has

Context {
    int id PK
    string name
    string icon "Material Design Icon URI"
    text description
    string owner_id
    string owner_type "currently only user, group envisioned in a future version"
}

Rel_Context_Resource {
    int id PK
    int context_id 
    int resource_id
    string resource_type
    int permissions "bitmask"
}

Page {
    int id PK
    int context_id
    string page_type "currently start page, or regular (better naming needed)"
}

Page_Content {
    int id PK
    int page_id
    int resource_rel_id
    int order "to be displayed in ascending order"
}

Shares {
    string node_type "introducing context as type"
}

Navigation {
    int share_id PK
    int display_mode "0 - hide, 1 - recipients, 2 - everybody"
    string user_id PK "NULL for default, an entry with user id acts as override"
}
```

### Resource Mapping

Currently, resources can be Tables or Views. In several places we have the tuple of `resource_id`  and `resource_type`.  These could be places in a another tables `Resources` that would contain the `id`, `type` and `original` id. Then, instead of the tuple, we can just refer to the id. This results in another join however, where necessary.

:heavy_check_mark: applied

### Less Normalization

With the current approach each proposed table has its narrow scope and any duplication is avoided. However, it requires joins in most places to map the required entity with the necessary context (pun not intended). It is possible to combine some tables to avoid joining. One candidate would be `Page` and `Page_Contents`. Then, the page id could not be a primary key anymore is it will be used with every row of content it may have.

The same is an options for `Share` and `Permissions`.

:heavy_multiplication_x: dismissed

## Remarks

- tables are prefixed with `**PREFIX**tables_contexts` (aport of the context context table itself, it will just be `**PREFIX**tables_contexts`)
- We (enjeck, datenangebot and I) discussed a possibility to hide specific table and view IDs from the end user by having a masked/hashed or otherwise disguised ID for resources. `Rel_Context_Resources`actually offers this possibility, by introducing an own id (that would also function as PK, changing the current PK to a unique index). 
[...]

Issue #311, opened 2023-05-23
Applications / Contexts


datenangebot:
Lets create contexts where we can put different resources together to act like an small app (smapp 🙃).
Resources can be views, tables, a description (rich text with all the smart picker magic) and maybe other "related resources" from nextcloud, chat room, files directory, ...

## Subtasks

- [x] #812 
- [x] https://github.com/nextcloud/tables/pull/861
- [x] #807
- [x] #864
- [x] #908
- [x] #863
- [x] #896
- [x] #895
- [x] #897
- [x] #898

### Follow up

- [ ] #933
- [ ] #934
- [ ] #935

## Summary

- The user facing naming would be "Application" but code wise we use Context to avoid confusion with Nextcloud apps of all sorts

This is what we discussed in Berlin in person within the contributor week:
![SCR-20231024-mgmf](https://github.com/nextcloud/tables/assets/55329475/0cfc84fd-fa71-41f3-9a9a-bd0c849ab7d3)
![SCR-20231024-mgno](https://github.com/nextcloud/tables/assets/55329475/e775cbfe-f4be-4de9-b419-5943dee6fed9)

juliushaertl:
Unsorted further related input from the feature planning meeting for future discussions:
- Look more from the user perspetive than tables developer (what to prioritize in the UI, how to present tables/views/contexts to the user)
  - For most users receiving shared views might be the most important
  - Possibly group them by their parent table/context
  - Wording of views
  - Possibly separate table design and usage more
  - Reusable definitions (table schema + views), import/export, sharable, maybe through the app store on the long term

juliushaertl:
@datenangebot Can you provide a summary of what was discussed in Berlin and what the MVP would look like and need in terms of implementation as discussed?

datenangebot:
> Can you provide a summary of what was discussed in Berlin and what the MVP would look like and need in terms of implementation as discussed?

I updated the initial post here with some mockups.
@jancborchardt @nimishavijay did I got that right or do you have something to add?

juliushaertl:
From the discussion with @datenangebot 

- Q1: Contexts will be owned by a user in
  - Out of scope but planned for the future: Make a group the owner
- Permission management
  - Q2: Each user with manage permission on a resource can add it to a context they create
    - The resource will show linked contexts
    - When the user looses manage permission, the context linking will still persist, listing them will give the resource owner control over who has access to data, but the manage permission is considered trusted to extend the base resource permissions to others through contexts
[...]

PR #841, opened 2024-02-14
feat(db): database structure for contexts

blizzz:
Implements the DB structure as proposed in #812 with these two differences:

* for consistency, I renamed "resource" to "node" how it is already named in the existing share table
* for it is not possible to have a column being nullable and part of the primary key, in navigation i changed the user_id to not be nullable and have the default set to an empty string. The intended logic therefore is changed that the default is an empty string, instead of null. Otherwise, same behaviour is envisioned.
* owner_type is now an int, to be used with constants to be introduced, as discussed

Dedicated indices will be added later on once the queries are standing. Ideally prereleases are not shipped without them being done.

:information_source: this PR is opened against a feature branch, `feat/contexts` freshly forked from main. This is because 0.7 is not completed and branched off so far. We can merge it to main either after 0.7 is branched off.

Contributes to #311

### DIFF
--- a/lib/Migration/Version000800Date20240213123743.php
+++ b/lib/Migration/Version000800Date20240213123743.php
@@ -1,0 +1,133 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2024 Arthur Schiwon <blizzz@arthur-schiwon.de>
+ *
+ * @author Arthur Schiwon <blizzz@arthur-schiwon.de>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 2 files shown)


lib/Migration/Version000800Date20240213123743.php
```diff
@@ -0,0 +1,133 @@
...
+	 * @throws SchemaException
+	 */
+	protected function haveContextTable(ISchemaWrapper $schema): void {
+		if ($table = $this->shouldAddTable(self::PREFIX . 'context', $schema)) {
+			$table->addColumn('id', Types::INTEGER, ['notnull' => true]);
+			$table->addColumn('name', Types::STRING, ['notnull' => true, 'length' => 200]);
+			$table->addColumn('icon', Types::STRING, ['notnull' => true, 'length' => 64]);
+			$table->addColumn('description', Types::TEXT);
+			$table->addColumn('owner_id', Types::STRING, ['notnull' => true, 'length' => 64]);
+			$table->addColumn('owner_type', Types::STRING, ['notnull' => true, 'length' => 64]);
```
juliushaertl, line 73: We could probably use an int to store this and use the IShare constant values for user/group/circle

blizzz: Indeed, having an int her makes sense. Technically it is not a share, going with IShare constants nag by naming and could be confusing, but would work. As would adding another set of consts :)

juliushaertl: Yep, own constants sounds good 👌

juliushaertl: Yep, own constants sounds good 👌

blizzz: done

juliushaertl (approved):

PR #843, opened 2024-02-14
feat: hold multiple tables/views in global store

enjeck:
As part of https://github.com/nextcloud/tables/issues/807 , we want to hold multiple Tables' data in the vuex store. We do this because the data is needed for various modals to function properly

### Todo
- [x] Consider whether or not to clear the store. In the normal case where we have just one Table or View per page, no need storing all previously-seen Table/View data


juliushaertl (approved):

juliushaertl:
Restarted the failing ci job as it looked like a network issue on the runner

PR #844, opened 2024-02-14
Proof of Concept: Show that multiple tables can be on same page

enjeck:
This is based on https://github.com/nextcloud/tables/pull/843. This PR won't be merged.  As part of https://github.com/nextcloud/tables/issues/807, we want to confirm that it is possible to have multiple tables/views on a single page and ensure that actions like editing/creating rows and columns still work. 

Too see this in action locally,
- Replace line 53 of Context.vue with existing Table ids
- Navigate to http://nextcloud.local/index.php/apps/tables/#/context/1


enjeck:
This is still a work in progress. 
### Todo
- [ ] Some processes like creating columns and editing rows need work. In https://github.com/nextcloud/tables/pull/808, I had assumed only the element id is needed, but we need the entire element. 
- [ ] Options.vue need to consider element ids in events. I missed this in https://github.com/nextcloud/tables/pull/808
- [ ] Discuss about what to do about sidebars e.g Sharing. Will these be needed for Contexts (https://github.com/nextcloud/tables/issues/311) or not? 
- [ ] Think about Table/View headers, which are sticky when we scroll. This stickiness isn't ideal when there are multiple elements on a page

juliushaertl:
Gave this a quick smoke test and seems to work quite well already (also with sticky headers of tables).

juliushaertl:
Thanks I'll have a first look tomorrow, I discussed with @elzody ealier that she can look into the icon picker which should go into the modal but can probably be worked on without conflicting with ongoing work too much.

We can further align then tomorrow

src/modules/navigation/sections/Navigation.vue
```diff
@@ -62,19 +52,32 @@
...
-						:key="table.id"
-						:filter-string="filterString"
-						:table="table" />
+					<NavigationTableItem v-for="table in getArchivedTables" :key="table.id"
+						:filter-string="filterString" :table="table" />
 				</NcAppNavigationItem>
 			</ul>
+			<div v-if="contextsLoading" class="icon-loading" />
+			<ul v-if="!contextsLoading">
+				<NcAppNavigationCaption :name="t('tables', 'Contexts')">
```
juliushaertl, line 61: User facing naming would be "Applications"
```suggestion
				<NcAppNavigationCaption :name="t('tables', 'Applications')">
```

src/pages/Contextpage.vue
```diff
@@ -0,0 +1,55 @@
+<!-- TODO figure out whether to keep this page, and how to structure it if yes -->
```
juliushaertl, line 1: I think we can drop it, showing contexts in the navigation is enough

src/router.js
```diff
@@ -15,6 +17,26 @@ export default new Router({
...
+		{
+			path: '/context/',
+			component: Contextpage,
+			name: 'context',
+		},
+		{
+			path: '/context/row/:rowId',
+			component: Contextpage,
+			name: 'contextRow',
+		},
```
juliushaertl, line 29: This can then be removed as well

juliushaertl (approved):
Fine to get into the contexts feature branch from my side so we can move to smaller follow up packages.

Can you check the CI failures? Passing Ci would still be required

PR #842, opened 2024-02-14
fix: properly delete tables with filtered view

enjeck:
This fixes an issue with deleting Tables with have filtered Views. 

### To reproduce 
- Create a table
- Create a view with filters
- Try to delete the table

Video showing the problem:
[Screencast from 02-14-2024 12:46:02 PM.webm](https://github.com/nextcloud/tables/assets/32180937/8fbf5596-b8d2-49a5-a592-3ce4dff0deeb)

<details>
<summary>Error traces:</summary>

```
[...]

PR #846, opened 2024-02-15
fix: Only skip failing table data loading instead of full failure

juliushaertl:
@enjeck mentioned that even with https://github.com/nextcloud/tables/pull/842 we could still have cases where the delete failed and we are left in a kind of broken state.

This commit makes sure that we do not throw a type error but a internal one that is logged and handled gracefully and we can just skip the failing table enhancement for listing the table. This way even if a deletion failed on a previous attempt, users can still use tables, list them and delete the orphaned entry with the fix from #842


enjeck (approved):
Thanks for looking into this!

Issue #829, opened 2024-02-08
Hide user status from owner avatar


ChristophWurst:
### Steps to reproduce

1. Have someone share a table with view with you

### Expected behavior

Distraction-free user interface

### Actual behavior

![image](https://github.com/nextcloud/tables/assets/1374172/ddef16ef-63b1-4f0c-abdc-98b409e7999b)

IMO no added value from showing the status there. It just consumes more space and makes the UI look busy.

### Web server

None

### PHP engine version

None

### Database

None

PR #853, opened 2024-02-16
Fix hide user status from owner avatar (issue #829)

elzody:
Status icon removed from navigation sidebar as illustrated in issue #829 

image

Also removed from other places which I thought were not needed, such as next to the table title, the share list when sharing the table, and the views dashboard:

image
image
image

If these additional changes are unnecessary or should not have been changed, it can easily be rolled back to only remove the icon from the navigation sidebar as shown in the original issue.

### DIFF
--- a/src/modules/main/sections/Dashboard.vue
+++ b/src/modules/main/sections/Dashboard.vue
@@ -47,7 +47,8 @@
 										<NcAvatar
 											:display-name="share.receiverDisplayName"
 											:user="share.receiver"
-											:is-no-user="share.receiverType !== 'user'" />
+											:is-no-user="share.receiverType !== 'user'"
+											:show-user-status="false" />
 									</div>
 								</div>
 							</td>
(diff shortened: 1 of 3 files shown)


juliushaertl:
Adjusting other places seems fine, just the share recipients i think it would make sense to keep the status to be consistent with other sharing sidebars (like in files) where we also show the status. Maybe you can just drop that commit.

elzody:
@jancborchardt Curious to know some designer feedback on this

elzody:
> Adjusting other places seems fine, just the share recipients i think it would make sense to keep the status to be consistent with other sharing sidebars (like in files) where we also show the status. Maybe you can just drop that commit.

I think I managed to drop that commit, hopefully I did it correctly without messing anything up :)

juliushaertl (approved):

juliushaertl:
Looks like that worked fine ;)

PR #865, opened 2024-02-22
feat: Add backend for table archive and favorite flag

### LINKED ISSUE #469: Consider different sorting or way of organizing the navigation
A few points we should discuss that came up in the feature review:

- For most users shared views might be the most important ones
- Shared views probably should also be grouped by table name (or have some indication), otherwise they just show "My request" but more useful would be "Vacation requests > My requests"
- Favorites might be useful
- Archiving tables or views might be useful

juliushaertl:
Resolves #469 

- [ ] Backend
  - [x] Adjust database layout
    - [x] Add archived flag to tables ~and views~
    - [x] Add favorites table that stores if a node (table/view) is favorited by a user
  - [x] Expose archived flag as an option to set through existing API endpoints when updating tables/views
  - [x] Add Controller and Favorite service to mark as favorite and remove favorite
  - [x] First working endpoints
  - [x] Cleanup a bit
  - [x] Adapt other api endpoints that have a comment in routes
  - [x] Proper response format for favorites
  - [x] Add capability for favorites api endpoints
  - [x] Check open api spec annotations
  - [x] Write integration tests
- [ ] Frontend
  - [x] Navigation
    - [x] Remove categories from navigation and just show a plain list (sort favorites first, then by last modification time)
    - [x] Hide archived tables/views in a collapsible entry (like in deck)
  - [x] Archiving
    - [x] Add button to archive/unarchive in the edit table and edit view dialogs
  - [x] Favorites
    - [x] Add action to the menu to mark as favorite and remove from favorites 
  - [ ] Cypress tests for adding/removing favorites
  - [ ] Cypress test for archiving/unarchiving

### DIFF
--- a/lib/Service/FavoritesService.php
+++ b/lib/Service/FavoritesService.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2024 Julius Härtl <jus@bitgrid.net>
+ *
+ * @author Julius Härtl <jus@bitgrid.net>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCA\Tables\Service;
+
+use OCP\Cache\CappedMemoryCache;
+use OCP\IDBConnection;
+
+class FavoritesService {
+
+	private CappedMemoryCache $cache;
+
+	public function __construct(private IDBConnection $connection, private ?string $userId) {
+		$this->cache = new CappedMemoryCache();
+	}
+
+	public function isFavorite(int $nodeType, int $id): bool {
+		if ($cached = $this->cache->get($this->userId . '_' . $nodeType . '_' . $id)) {
+			return $cached;
+		}
+
+		// We still might run this multiple times
+
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select('*')
+			->from('tables_favorites')
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($this->userId)));
+
+		$result = $qb->executeQuery();
+		while ($row = $result->fetch()) {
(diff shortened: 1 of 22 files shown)


juliushaertl:
@blizzz A quick backend review would already be appreciated :)

lib/AppInfo/Application.php
```diff
@@ -26,6 +26,9 @@
 class Application extends App implements IBootstrap {
 	public const APP_ID = 'tables';
 
+	public const NODE_TYPE_TABLE = 0;
```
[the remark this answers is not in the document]
blizzz: I have not yet, thanks for the heads up :+1:

lib/Service/FavoritesService.php
```diff
@@ -0,0 +1,138 @@
...
+		?string $userId
+	) {
+		$this->connection = $connection;
+		$this->permissionsService = $permissionsService;
+		$this->userId = $userId;
+		$this->cache = new CappedMemoryCache();
+	}
+
+	public function isFavorite(int $nodeType, int $id): bool {
+		$cacheKey = $this->userId . '_' . $nodeType . '_' . $id;
```
blizzz, line 54: userId does not need to be in the cacheKey. This confused me a bit, for it appears the cache could be in use for different users (and then clearing everything in the next line would be too much). Naming things ;)

Can change, not must change.

blizzz (approved):
Backend bits look good!

lib/Service/FavoritesService.php
juliushaertl: Adjusted but with a comment on the constructor to avoid someone changing this to a shared cache without thinking about side effects.

juliushaertl:
@elzody I pushed one more fix to also expose the favorite flag on views when they are returned through the API.

src/modules/navigation/sections/Navigation.vue
```diff
@@ -35,6 +38,22 @@
 					:view="view" />
 			</ul>
 
+			<NcAppNavigationItem :name="t('tables', 'Archived tables')" :allow-collapse="true" :open="false">
```
juliushaertl, line 41: We probably can hide this entry if no archived tables are there

src/modules/navigation/partials/NavigationTableItem.vue
```diff
@@ -193,6 +219,18 @@ export default {
...
+			const res = await this.$store.dispatch('updateTable', {
+				id: this.table.id,
+				data: { archived },
+			})
+
+			// eslint-disable-next-line no-console
+			if (!res) { console.log('failed to archive/unarchive table') }
+
+			// eslint-disable-next-line no-console
+			console.log('archived/unarchived table')
```
juliushaertl, line 232: Can probably dropped, the vuex store method should already log a console error if this fails.

juliushaertl:
Tested what is already there and works like a charm :)

Maybe you can rebase with your next push to latest main branch to resolve the conflict with the merge of #878

juliushaertl (commented):
Cannot self-approve but for the frontend :+1:

juliushaertl:
Good catch @enjeck 

I'll merge this for now and file follow up tasks for the open items

PR #877, opened 2024-02-27
Extract selection option labels for Analytics

Rello:
- when column type selection is used, the API needs to deliver the labels, and not the internal ids
fixes https://github.com/Rello/analytics/issues/385

- Using lib/Analytics (and not lib/Datasource) for better readability
- Adopting from LegacyRow to Row

### DIFF
--- a/lib/Analytics/AnalyticsDatasource.php
+++ b/lib/Analytics/AnalyticsDatasource.php
@@ -9,13 +9,14 @@
  * @author Marcel Scherello <analytics@scherello.de>
  */
 
-namespace OCA\Tables\Datasource;
+namespace OCA\Tables\Analytics;
 
 use OCA\Analytics\Datasource\IDatasource;
-use OCA\Tables\Api\V1Api;
 use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Errors\PermissionError;
+use OCA\Tables\Service\ColumnService;
+use OCA\Tables\Service\RowService;
 use OCA\Tables\Service\TableService;
 use OCA\Tables\Service\ViewService;
 use OCP\AppFramework\Db\DoesNotExistException;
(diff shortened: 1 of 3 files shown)


enjeck:
@Rello Hi! Are you still working on this?

Rello:
Hi @juliushaertl you think we can merge this?
tests look ok to me if I can judge them correctly?

blizzz:
I am afraid this has potential to be breaking potential existing API use cases. I suppose the IDs of the selection can be fetched already (did not double check). Not that the output is much documented.

There is a comflict btw. If we merge, it's better to have the commits squashed.

Rello:
@blizzz what do you suggest?
The idea is nice dynamic and the same for tables & forms (potentially for others). 
An app can register itself to Analytics.
but all the requirements like tests & co are far more that can be achieved by a hobby project.

I think of skipping this and simply ship the datasources for everything with Analytics itself. Kind of double work because local duplication of the API logic might be required in many places. but not impacting other apps

suggestion?
also applies to 
https://github.com/nextcloud/forms/pull/2195

Rello:
as discussed with @blizzz 
the Analytics data source was decoupled from the API endpoint to not create breaking changes.
Analytics is taking care of any specific logic on its own

lib/Analytics/AnalyticsDatasource.php
```diff
@@ -172,13 +176,69 @@ public function readData($option): array {
...
 
-		return [
-			'header' => $header,
-			'dimensions' => array_slice($header, 0, count($header) - 1),
-			'data' => $data,
-			//'rawdata' => $data,
-			'error' => 0,
-		];
+		return ['header' => $header, 'dimensions' => array_slice($header, 0, count($header) - 1), 'data' => $data, //'rawdata' => $data,
+			'error' => 0,];
```
blizzz, line 180: Nitpick-as-a-Service: indentation looks odd. Would be better in the previous style.

blizzz (approved):
You'll make me extra happy by squashing the commits =)

P.S.: Let me do that quickly

lib/Analytics/AnalyticsDatasource.php
Rello: let me check which code styler setting messed this up

blizzz: saw to late. but a follow up pr is welcomed.

Issue #365, opened 2023-06-13
Add import link to spreadsheets in files


datenangebot:
like analytics did...

PR #890, opened 2024-03-01
Feat: File action to import from Files to Tables

elzody:
## Related Issue
This feature is in response to issue #365 

## Explanation
This feature adds a file action to files in the Files app, which allows you to import the file into Tables, provided it is a supported file format. Upon clicking the "Import into Tables" button in the file action menu, a dialog will be displayed which lets you choose to import the data as a new table, or if you want to import the data into an existing table.

Importing the data as a new table allows you to choose an icon and a name for the table, and if none are provided, will use the 🔧 icon and the file name for the table information. If you choose to import the file into an already existing table, it allows you to select one of your tables and gives you the choice of creating missing columns or not. *

\* Possible bug with this, will look into it

## Screenshots
<details>
<summary>File action</summary>

grafik
</details>

<details>
<summary>Import dialog</summary>

grafik
grafik
</details>

<details>
<summary>Import results dialog</summary>

grafik
</details>

## TODO
- [x] Look into better way of including the Tables icon in the file action than hardcoding the SVG string
- [x] Because 2 new components were added, look into maybe the re-usability of these, particularly the import results dialog?
- [x] No testing yet, so this would probably be a good thing to include at some point
- [x] Instead of passing `t` function for translation to the modal, can probably just import it itself (just realized this was a thing)
- [ ] Implement better error handling / error feedback for the user
- [x] There seems to be a residual commit that is unrelated to this PR; it should be dropped

## Follow up
- [ ] Fix potential issue with importing into an existing table when the "Create missing columns" toggle is disabled
- [ ] Upon mounting the import dialog component, the user's tables are retrieved from the backend. Look into ways to cache this and reduce number of network requests.

### DIFF
--- a/cypress/e2e/tables-import.cy.js
+++ b/cypress/e2e/tables-import.cy.js
@@ -49,55 +49,59 @@ describe('Import csv', () => {
 
 })
 
-describe('Import csv from Files file action', () => {
+// only run the following tests if not testing on v26 or v27
+// the files api in use is only supported on v28+
+if (!['stable26', 'stable27'].includes(Cypress.env('ncVersion'))) {
+	describe('Import csv from Files file action', () => {
 
-	before(function() {
-		cy.createRandomUser().then(user => {
-			localUser = user
-			cy.login(localUser)
-			cy.uploadFile('test-import.csv', 'text/csv')
+		before(function() {
+			cy.createRandomUser().then(user => {
+				localUser = user
+				cy.login(localUser)
+				cy.uploadFile('test-import.csv', 'text/csv')
+			})
 		})
-	})
 
-	beforeEach(function() {
-		cy.login(localUser)
-		cy.visit('apps/files/files')
-	})
+		beforeEach(function() {
+			cy.login(localUser)
+			cy.visit('apps/files/files')
+		})
 
-	it('Import to new table', () => {
-		cy.get('[data-cy-files-list-row-name="test-import.csv"] [data-cy-files-list-row-actions] .action-item button').click()
-		cy.get('[data-cy-files-list-row-action="import-to-tables"]').click()
+		it('Import to new table', () => {
+			cy.get('[data-cy-files-list-row-name="test-import.csv"] [data-cy-files-list-row-actions] .action-item button').click()
+			cy.get('[data-cy-files-list-row-action="import-to-tables"]').click()
 
-		cy.intercept({ method: 'POST', url: '**/apps/tables/import/table/*'}).as('importNewTableReq')
-		cy.get('[data-cy="fileActionImportButton"]').click()
-		cy.wait('@importNewTableReq').its('response.statusCode').should('equal', 200)
+			cy.intercept({ method: 'POST', url: '**/apps/tables/import/table/*'}).as('importNewTableReq')
+			cy.get('[data-cy="fileActionImportButton"]').click()
+			cy.wait('@importNewTableReq').its('response.statusCode').should('equal', 200)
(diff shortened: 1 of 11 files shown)


juliushaertl (commented):
A few comments, only read the code so far, haven't tried it yet

cypress/e2e/tables-import.cy.js
```diff
@@ -30,3 +32,56 @@ describe('Import csv', () => {
...
+	})
+
+	it('Import to new table', () => {
+		cy.reload()
+
+		cy.get('[data-cy-files-list-row-actions] .action-item button').click()
+		cy.get('[data-cy-files-list-row-action="import-to-tables"]').click()
+		cy.get('[data-cy="fileActionImportButton"]').click()
+
+		cy.wait(5200)
```
juliushaertl, line 57: `cy.wait` is an anti-pattern as it makes tests flaky or slower then needed.

best here would be to intercept the HTTP request and wait for it to finish.

See https://github.com/nextcloud/tables/blob/f918c273decd57b23988fe99e16258f4c6ae208c/cypress/e2e/view.cy.js#L69-L73 for examples

elzody: Yep, gotcha. Didn't figure out it was a thing until working on the tests for archive/favorites. Will fix

juliushaertl:
When wrapping up, would also be nice to squash commits in a meaningful way a bit if possible or just all in one feature commit which makes the history once merged a bit clearer. Also I haven't mentioned before and we don't enforce this, but most of us try to somewhat stick to conventional commits https://www.conventionalcommits.org/en/v1.0.0/ Might be nice to adapt that as well ;)

src/file-actions.js
```diff
@@ -0,0 +1,30 @@
...
+	enabled: (files) => {
+		const file = files[0]
+
+		return file.type === 'file' && validMimeTypes.includes(file.mime)
+	},
+
+	exec: (file) => {
+		spawnDialog(FileActionImport, { file })
+		return null
+	},
```
juliushaertl, line 27: We could possibly use [dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports) to reduce the size of the entry point that is loaded and parsed on the files app. That way webpack can split the javascript chunks and this helps keeping the initial loading time sane:

Haven't tested but something like this should work (and removing the original import):
```suggestion
	async exec: (file) => {
		const { default: FileActionImport } = await import('./modules/modals/FileActionImport.vue')
		spawnDialog(FileActionImport, { file })
		return null
	},
```

juliushaertl (commented):
Just two comments, otherwise this looks good 👍

juliushaertl:
Just pushed two small fixups when trying out locally. First one for cypress, second adding the missing dependencies

juliushaertl:
The new files api only works with 28 so we just just skip the cypress tests on earlier versions.

juliushaertl:
We can just pass in the server version as collectives does in the cypress.yml github workflow:

	CYPRESS_ncVersion: ${{ matrix.server-versions }}

Then we can skip the tests in the spec files

	102:			if (['stable26', 'stable27'].includes(Cypress.env('ncVersion'))) {

juliushaertl:
Rebased, squashed and pushed a small follow up fix. I noticed that the tests seemed to fail only sometimes where the file action was not showing up. This might happen if the files app code initializes before the action was registered, so I moved the script loading to be earlier

juliushaertl (approved):

PR #900, opened 2024-03-04
Match default values when querying views

juliushaertl:
When gathering rows that are visible in views we need to take rows into
account that have no corresponding entry in the columns_TYPE tables. For
those we can match the filter in PHP and include row ids from a left
join where there is no entry.

Taking an example of a yes/no selection:

If the default value of a column doesn't match we end up using the same
query as before for collecting row ids

```
...
	`id` IN (SELECT `row_id` FROM `oc_tables_row_cells_selection` WHERE (`column_id` = 7) AND (`value` LIKE '"false"'))
...
```

If the default value matches, we need to include those that do not have
an entry

```
...
	`id` IN (SELECT `row_id` FROM `oc_tables_row_cells_selection` WHERE (`column_id` = 7) AND (`value` LIKE '"false"'))
OR
	`id` IN (select sl.id from oc_tables_row_sleeves sl LEFT JOIN oc_tables_row_cells_selection se ON sl.id = se.row_id AND se.column_id = 7 WHERE se.id IS NULL)
...
```

## Steps to reproduce

- Create a table with one text field
- Add a row
- Add a new column to it with some default value
- Create a filter that filters for the default value

### DIFF
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -309,46 +309,86 @@ class Row2Mapper {
 		$paramType = $this->getColumnDbParamType($column);
 		$value = $this->formatValue($column, $value, 'in');
 
+		// We try to match the requested value against the default before building the query
+		// so we know if we shall include rows that have no entry in the column_TYPE tables upfront
+		$includeDefault = false;
+		$defaultValue = $this->getFormattedDefaultValue($column);
+
 		$qb2 = $this->db->getQueryBuilder();
-		$qb2->select('row_id');
-		$qb2->where($qb->expr()->eq('column_id', $qb->createNamedParameter($column->getId()), IQueryBuilder::PARAM_INT));
-		$qb2->from('tables_row_cells_' . $column->getType());
+		$qb2->selectAlias('sl.id', 'row_id')
+			->from('tables_row_sleeves', 'sl')
+			->leftJoin('sl', 'tables_row_cells_' . $column->getType(), 'v', $qb->expr()->andX(
+				$qb->expr()->eq('sl.id', 'v.row_id'),
+				$qb->expr()->eq('v.column_id', $qb->createNamedParameter($column->getId(), IQueryBuilder::PARAM_INT)),
+			));
+		$qb2->where(
+			$qb->expr()->eq('sl.table_id', $qb->createNamedParameter($column->getTableId(), IQueryBuilder::PARAM_INT))
+		);
 
 		switch ($operator) {
 			case 'begins-with':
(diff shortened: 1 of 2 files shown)


lib/Db/Row2Mapper.php
```diff
@@ -309,46 +309,84 @@ private function getFilterExpression(IQueryBuilder $qb, Column $column, string $
...
 		$qb2 = $this->db->getQueryBuilder();
-		$qb2->select('row_id');
-		$qb2->where($qb->expr()->eq('column_id', $qb->createNamedParameter($column->getId()), IQueryBuilder::PARAM_INT));
-		$qb2->from('tables_row_cells_' . $column->getType());
+		$qb2->selectAlias('sl.id', 'row_id')
+			->from('tables_row_sleeves', 'sl')
+			->leftJoin('sl', 'tables_row_cells_' . $column->getType(), 'v', $qb->expr()->andX(
+				$qb->expr()->eq('sl.id', 'v.row_id'),
+				$qb->expr()->eq('v.column_id', $qb->createNamedParameter($column->getId(), IQueryBuilder::PARAM_INT)),
+				$qb->expr()->eq('sl.table_id', $qb->createNamedParameter($column->getTableId(), IQueryBuilder::PARAM_INT))
```
blizzz, line 323: Being a left join and since it is not establishing a condition between `sl` and `v`, this is rather to be understood as part of the general where clause, no?

blizzz (approved):
reasonable change, one question.

lib/Db/Row2Mapper.php
juliushaertl: Yes, good catch for the sl.table_id, the v.column_id we need to also include NULL values on the join where the row_id does not match

juliushaertl: Adjusted

juliushaertl:
So I would love to have comprehensive tests for this but lack time at the moment. Also still a bit unsure how to cover that. Behat tests might just be to verbose for test definition, so I might go with a phpunit tests that still writes to the DB to cover the different value conversions more easily in a dataProvider.

Nevertheless I'd follow up this if you're fine with it @blizzz

blizzz:
> So I would love to have comprehensive tests for this but lack time at the moment. Also still a bit unsure how to cover that. Behat tests might just be to verbose for test definition, so I might go with a phpunit tests that still writes to the DB to cover the different value conversions more easily in a dataProvider.

Not a fan of unit tests for DB logic tbh, but would be a way to tackle it.

> Nevertheless I'd follow up this if you're fine with it @blizzz

Welcome to the club :neutral_face: perhaps add an issue for tracking.  You have my blessing ;)

Issue #896, opened 2024-03-03
Applications: Create/edit modal


juliushaertl:
-  [x] Title input
-  [x] Description text input
- [x] Resource list
  -  [x] Add button that shows a multiselect allowing to pick tables/views that the user has to add
  - [x] Remove button for each resource
- [ ] Change ordering of resources in the edit modal

juliushaertl:
Moved the left-over here to a follow up task mentioned in https://github.com/nextcloud/tables/issues/311

Issue #863, opened 2024-02-22
Applications: Expose in app menu


juliushaertl:
- [ ] Implement exposing entries in the navigation manager

juliushaertl:
@blizzz Assigned you as we discussed that as the next backend step

PR #902, opened 2024-03-05
feat: Add material design icons for applications

juliushaertl:
In preparation for #863 and #896 

This PR adds a script that automatically extracts material design icons and metadata based on a jq json filter to tables. For now we can just start with the default google ones stripped down without variants.

We cannot use the npm library as the backend needs to serve the image url for the nativation entries. The json metadata file can be used in the frontend to generate a picker component then.


build/fetch-material-icons.sh
```diff
@@ -0,0 +1,23 @@
...
+if [ ! -d "${CLONE_PATH}" ]; then
+    git clone "${REPO_URL}" --depth 1 "${CLONE_PATH}"
+fi
+
+mkdir -p "$TARGET_PATH"
+
+cp $CLONE_PATH/LICENSE $TARGET_PATH/LICENSE
+
+cat $CLONE_PATH/meta.json | 
+    jq '.[]|select(.author == "Google" and (.name | contains("-") | not))' > $TARGET_PATH/meta.json
```
blizzz, line 19: Fine for simplicity, but want to point out that some icons are unique while having a dash, for instance `cash-100`, `air-purifier` or `contactless-payment-circle`.

blizzz (approved):

build/fetch-material-icons.sh
juliushaertl: Yeah we can extend the list later on

PR #904, opened 2024-03-05
feat(Contexts): add Contexts to navigation when applicable

blizzz:
solves #863 

needs #903 (soft)

~~wants https://github.com/nextcloud/server/pull/44178~~

* [x] generate correct link
* [x] does not appear yet for the owner even if there is a share where display is enabled for all
* [ ] move logic from application to db query (see TODO remark in code)

Screenshot made with frontend from https://github.com/nextcloud/tables/pull/844. On the content area in the UI we don't render the material icons yet, but the icon is in the nav bar. The icon over the mouse pointer is material design, while the two icons on the very right have fallback icons (default app icon). 
 
Screenshot_20240313_161237

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -6,6 +6,7 @@ use Exception;
 use OCA\Analytics\Datasource\DatasourceEvent;
 use OCA\Tables\Capabilities;
 use OCA\Tables\Listener\AnalyticsDatasourceListener;
+use OCA\Tables\Listener\BeforeTemplateRenderedListener;
 use OCA\Tables\Listener\LoadAdditionalListener;
 use OCA\Tables\Listener\TablesReferenceListener;
 use OCA\Tables\Listener\UserDeletedListener;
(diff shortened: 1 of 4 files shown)


lib/AppInfo/Application.php
```diff
@@ -68,5 +77,61 @@ public function register(IRegistrationContext $context): void {
 	}
 
 	public function boot(IBootContext $context): void {
+		$context->injectFn([$this, 'addPersonalNavigationEntries']);
```
juliushaertl, line 80: Im wondering if we can do this lazy in some way, maybe on a BeforeTemplateRenderedEvent so we don't trigger this logic on every request

juliushaertl (commented):
Small comment otherwise looks straight forward

lib/AppInfo/Application.php
blizzz: Yes, this disturbs me, too.

blizzz:
@enjeck what would be a nice addition is that the appropriate nav bar icon is set active, when being on an application that has an entry there (otherwise going back to tables).

juliushaertl:
> @enjeck what would be a nice addition is that the appropriate nav bar icon is set active, when being on an application that has an entry there (otherwise going back to tables).

I think this is tricky to do on the frontend. We may need a separate backend route to still render the TemplateResponse but then set the active entry on the navigation manager. We can then set an initial state that indicates that we're on an application page and to further adjustments on the frontend side like hiding the navigation by default. Something for a follow up I'd say.

juliushaertl (approved):

blizzz:
> > @enjeck what would be a nice addition is that the appropriate nav bar icon is set active, when being on an application that has an entry there (otherwise going back to tables).
> 
> I think this is tricky to do on the frontend. We may need a separate backend route to still render the TemplateResponse but then set the active entry on the navigation manager. We can then set an initial state that indicates that we're on an application page and to further adjustments on the frontend side like hiding the navigation by default. Something for a follow up I'd say.

Ah, I was speculating that such an option exist for fronted, but of course… why would it. There was no similar case before. Follow up is fine +1

Issue #892, opened 2024-03-03
Archive: Indent views in archived board listing


juliushaertl:
Not a big deal, but it'll be nice if we eventually indent the views in **Archived tables** sidebar, so that it's obvious they're part of a specific table:

Here, "view" is a view of "ToDo list", but it's not obvious due to lack of indent:
![image](https://github.com/nextcloud/tables/assets/32180937/c1e23526-aca7-403e-8ae2-debdad40573c)

_Originally posted by @enjeck in https://github.com/nextcloud/tables/issues/865#issuecomment-1974847194_

juliushaertl:
Fixed by #909

PR #909, opened 2024-03-06
fix: properly indent child views of archived tables

elzody:
Fix for issue #892 

Added some CSS which will further indent views which are nested in an archived table to make the hierarchy more visible. If there's a better way, definitely let me know :)

### Screenshots
<details>
<summary>Previous</summary>

grafik
</details>

<details>
<summary>Suggested</summary>

grafik
</details>

### DIFF
--- a/src/modules/navigation/sections/Navigation.vue
+++ b/src/modules/navigation/sections/Navigation.vue
@@ -49,6 +49,7 @@
 
 				<!-- ARCHIVED -->
 				<NcAppNavigationItem v-if="getArchivedTables.length > 0"
+					class="archived-items"
 					:name="t('tables', 'Archived tables')"
 					:allow-collapse="true"
 					:open="false">
@@ -182,7 +183,6 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-
 :deep(.filter-box) {
 	.input-field {
 		padding: 8px;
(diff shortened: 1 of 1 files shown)


src/modules/navigation/sections/Navigation.vue
```diff
@@ -201,5 +201,14 @@ export default {
 		margin-top: 3vh;
 	}
 }
+</style>
 
+<style>
+.archived-items .app-navigation-entry__children {
+	& .app-navigation-entry-wrapper .app-navigation-entry__children {
+		& .app-navigation-entry {
+			padding-left: calc(16px * 2);
```
juliushaertl, line 210: Any specific reason to not just use 32px?

elzody: I figured it would be clearer to use that since I did not have access to the $icon-size variable which is defined in the Vue library and set at 16px, to make it clearer that that's what I was referencing. But yeah this can be changed

juliushaertl: All right, would have been fine either way, in that case a code comment might have made sense explaining the calculation, but fine to keep the 32px as well 👍

juliushaertl (approved):

jancborchardt:
@elzody could you add before/after screenshot so a design review is easier? :) Thanks!

elzody:
> @elzody could you add before/after screenshot so a design review is easier? :) Thanks!

No problem, added it now @jancborchardt

jancborchardt (approved):
Looks good design-wise! Regarding code rather than putting a set number, might be better to use a multiple of the default-grid-baseline variable? (Which is 4px, so 8* that in this case.)

juliushaertl:
> Looks good design-wise! Regarding code rather than putting a set number, might be better to use a multiple of the default-grid-baseline variable? (Which is 4px, so 8* that in this case.)

I'd get this in for now. The offset depends on the icon size which is currently also hard coded in the vue library

Issue #911, opened 2024-03-07
View columns do not respect sorting


enjeck:
### Steps to reproduce

1. Create a table
2. While creating a view for this table, go to the "Sort" tab and "Add a new sorting rule", select a column and order to sort by.
3. Create and open view

### Expected behavior

The View columns should be sorted based on the selected column and order

### Actual behavior

No sorting happens

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

The sorting information is send by the backend, seems to be a frontend issue

PR #912, opened 2024-03-08
fix: views respect sorting

enjeck:
Fixes https://github.com/nextcloud/tables/issues/911


juliushaertl:
Thanks for looking into that.

> Sorts that are created initially when the view is created are highlighted in the UI as a local setting, and prompts the user Reset local adjustments to erase it. Do we want to make this setting fixed and non-removable?

Yes, i think that makes sense. If custom sortings are applied then we should of course still show the reset button to go back to the view defined sorting.

> Some columns like RichText are not sortable. For good UX, we should maybe remove these unsortable options from the list of columns when adding a new sorting rule.

:+1:

Issue #799, opened 2024-01-25
Generate typescript definitions from openapi spec


juliushaertl:
As @enjeck brought up it is quite hard to work with a yet unfamiliar code base that has complex data structures as the tables app, so we thought that the openapi spec could be used to automatically generate typescript definitions.

While rolling out typescript in general is probably worth a general discussion and an iterative process, using https://www.npmjs.com/package/openapi-typescript seems to be able to generate a somewhat reasonable definition file.

We should see if:
- this can be used as jsdoc annotations to get proper IDE type hints
- We need any additional setup to be able to already move parts of the code to typescript
- Probably automate generating and updating the definitions

enjeck:
> this can be used as jsdoc annotations to get proper IDE type hints

I tried doing something like this, but it wasn't very helpful. I created a new file with `typedef`s for Tables, Columns and others. Imported these types into other files, and it was kinda helpful. 

The problem now might be VS Code not integrating well with Vue? For example, suppose there is a component, having a prop named `columns` under props. The `columns` prop doesn't get referenced when when I use it (via `this.columns`) under the `data`, `computed` and `methods` option. Wondering if using a specific IDE or extension helps with this. If I could fix this, I'll be very happy

PR #919, opened 2024-03-11
fix(OpenApi): resolve errors and generate openapi file

blizzz:
composer run openapi was not running through


.github/workflows/openapi.yml
```diff
@@ -0,0 +1,81 @@
...
+          coverage: none
+          ini-file: development
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Check Typescript OpenApi types
+        id: check_typescript_openapi
+        uses: andstor/file-existence-action@076e0072799f4942c8bc574a82233e1e4d13e9d6 # v3.0.0
+        with:
+          files: "src/types/openapi/openapi*.ts"
```
juliushaertl, line 36: We don't have that (yet), but let me quickly check how talk does it, maybe we can adapt it as well for #799

juliushaertl: Pushed a commit

juliushaertl (approved):

.github/workflows/openapi.yml
blizzz: Aye, it was optional for the check however.

blizzz:
[Green smoke](https://github.com/nextcloud/tables/actions/runs/8247790348/job/22556708213?pr=919)!

PR #921, opened 2024-03-11
fix: shared views can be favorited

elzody:
## Current behavior
If a view is shared (the user is not the owner of the view), the user will have the option to favorite/unfavorite it, but this will not be reflected in the UI (it will not be added/removed from the favorites list).

## Suggested behavior
Fix the behavior defined above. Additionally, if a table that contains views is favorited, its views will then not display a favorite/unfavorite button until the table is unfavorited. Any views which are already favorited but are also a part of the table will disappear from the favorite list as an individual view, and instead appear normally under the favorited table. This is done to prevent the view from showing in the favorites list twice.

## Recording

https://github.com/nextcloud/tables/assets/23369449/02a5f820-f888-44d7-8ff9-87d1a33a0b9d

## TODO
- [x] Determine if this behavior is appropriate
- [x] Address any design concerns
- [x] Implement testing

### DIFF
--- a/cypress/e2e/tables-favorite.cy.js
+++ b/cypress/e2e/tables-favorite.cy.js
@@ -67,5 +67,23 @@ describe('Favorite tables/views', () => {
 
     cy.get('@testView').parent().parent().should('contain.text', 'Tutorial')
   })
-  
+
+  it('can (un)favorite views with favorited parent tables', () => {
+    cy.get('[data-cy="navigationViewItem"]').first().as('testView')
+    cy.get('[data-cy="navigationTableItem"]').first().as('tutorialTable')
+
+    cy.get('@testView').parent().parent().should('contain.text', 'Tutorial')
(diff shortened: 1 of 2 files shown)


nimishavijay (commented):
Looks good! I would only suggest that the unfavoriting action for a view still be shown even if it's under a favorited table, in case people want to unfavorite individual views on a favorited table. Then it would also follow a consistent pattern of "everything that's not favorited can be favorited and everything that is favorited can be unfavorited". Does that make sense? :)

elzody:
> Looks good! I would only suggest that the unfavoriting action for a view still be shown even if it's under a favorited table, in case people want to unfavorite individual views on a favorited table. Then it would also follow a consistent pattern of "everything that's not favorited can be favorited and everything that is favorited can be unfavorited". Does that make sense? :)

Yes, absolutely. Thank you!

PR #925, opened 2024-03-12
fix: Focus first input when creating  a new table/row/column/view

juliushaertl:
Make sure that the modals to create new elements always focus the first input by default.


elzody (approved):
Seems fine to me, also gave it a quick test and it works. Gave me the chance to look up `nextTick` and do a bit of learning :)

juliushaertl:
Will check the test failure, i think it is just due to the multiselect being automatically focused now

Issue #895, opened 2024-03-03
Applications: Ship icons and icon list for frontend/backend


juliushaertl:
- [x] Extract icons and metadata that we pick from https://github.com/Templarian/MaterialDesign
- [x] Add icons and json to the repo
- [x] Frontend: Implement icon picker popover instead of an emoji picker based on the list from the json added in #902 
- [x] Backend: Validate icon when saving context, expose url to navigation entries

```
cat meta.json| jq '.[]|select(.author == "Google" and (.name | contains("-") | not)) | .name | "svg/" + . + ".svg"' -r | xargs -I{} cp {} picked/
```

PR #929, opened 2024-03-13
feat: contexts material design icon picker

elzody:
Material design icon picker for Contexts.

Part of #895

### DIFF
--- a/src/pages/Context.vue
+++ b/src/pages/Context.vue
@@ -2,12 +2,14 @@
 	<div class="row">
 		<div v-if="loading" class="icon-loading" />
 		<div v-if="!loading && context">
-			<div class="content first-row">
-				<div class="row">
-					<h1> {{ activeContext.iconName }}&nbsp; {{ activeContext.name }}</h1>
+			<div class="content context">
+				<div class="row first-row">
+					<h1 class="context__title">
+						<NcIconSvgWrapper :svg="icon" :size="32" style="display: inline-block;" />&nbsp; {{ activeContext.name }}
+					</h1>
 				</div>
-				<div class="row">
-					<h3> {{ activeContext.description }}</h3>
+				<div class="row space-L context__description">
+					{{ activeContext.description }}
 				</div>
 			</div>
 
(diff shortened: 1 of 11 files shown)


src/shared/components/ncIconPicker/NcIconPicker.vue
```diff
@@ -0,0 +1,51 @@
...
+export default {
+	name: 'NcIconPicker',
+
+	components: {
+		NcPopover,
+		IconPicker,
+	},
+
+	props: {
+		closeOnSelect: {
```
juliushaertl, line 26: Probably still on your list, just to not forget: would be good to pass in the current icon for the edit modal here as well

src/pages/Context.vue
```diff
@@ -4,7 +4,9 @@
 		<div v-if="!loading && context">
 			<div class="content first-row">
 				<div class="row">
-					<h1> {{ activeContext.iconName }}&nbsp; {{ activeContext.name }}</h1>
+					<h1 style="display: flex; align-items: center;">
+						<NcIconSvgWrapper :svg="icon" :size="40" style="display: inline-block;" />&nbsp; {{ activeContext.name }}
```
juliushaertl, line 8: Not sure why but that icon doesn't seem to render, in the navigation it works fine though.

webpack.js
```diff
@@ -17,4 +17,8 @@ webpackConfig.module = {
 	]
 }
 
+webpackConfig.resolve.alias = {
+  MaterialIcons: path.resolve(__dirname, 'img/material'),
+}
```
juliushaertl, line 22: We could probably make use of https://webpack.js.org/api/module-methods/#webpackchunkname to put all material icons into one chunk so we avoid requesting one js file for each of the icons

juliushaertl:
Pushed a quick commit to address the icon color and rebased

src/shared/components/ncIconPicker/NcIconPicker.vue
juliushaertl: I seen this is handled differently now, fixed it to properly pick up the currently set icon in my latest commit

juliushaertl (approved):

elzody:
> @elzody I still don't see the icons in the navigation. Just me? Screenshot from 2024-03-14 14-33-51

They show up fine for me, and I'm assuming for Julius as well? 🤔

elzody:
> Here's what I see in the console. I tried restarting my NC server, didn't help:

Looks like it's still trying to pull an emoji from the backend instead of an icon? Interesting, I'll dig through it and see what I can find, then make another fix

PR #931, opened 2024-03-14
add print styles

FahrJo:
This PR addresses https://github.com/nextcloud/tables/issues/107.
It was tested with Firefox 123.0.1, Chromium 116.0 and Safari 17.3. There sometimes occur minor glitches on Chromium and Safari but Firefox works well with my sample tables and views.


elzody:
I think the most expected behavior can be subjective, but when it comes to printing I think some UI elements should definitely be clipped. It's hard to say what a user might expect when they click the print button - maybe it's a matter of keeping the printing as-is when clicking the print button in the browser, but adding a print button to Tables could provide a way to print exclusively the table when desired, or tweak the behavior even further.

It may be difficult to come up with a solution for wide tables/columns, but I think the changes so far look okay. May require some further discussion, but I think you're on the right track.

FahrJo:
I agree that there is room for improvement, but so far printing a table has not been possible at all. If I have the time, I can also add a button for this print view (in another PR). On the other hand, that would require consideration of the keyboard shortcut *Ctrl+P*. As a user, I would personally prefer the default print content to be the table/view only.

elzody:
> I agree that there is room for improvement, but so far printing a table has not been possible at all. If I have the time, I can also add a button for this print view (in another PR). On the other hand, that would require consideration of the keyboard shortcut _Ctrl+P_. As a user, I would personally prefer the default print content to be the table/view only.

I say go for it :) If anything, it will bring it to light and can be adjusted as necessary

Issue #54, opened 2022-07-22
Add an introduction textarea


Atn-D:
It would be very useful to be able to write in a textarea field a description of the use of the table currently displayed.

It would also be very useful to give indications on how to fill it up in case of sharing.

If markdown was supported, it would be perfect!

datenangebot:
Hi,

thanks for this great idea, I was thinking about that too.
It's on the roadmap, waiting for a technical solution for texts field.
But I'm on it.

Regards

Atn-D:
Hello,
Would it be simpler and faster to have at first just a simple text field that is edited in the name and emoji popin.

And then think about how to implement a rich text?

datenangebot:
Hi,

the technical solution is with nextcloud 26 served.
So, ready to implement. Just a question on prioritization...

Atn-D:
Ok, thx :)

PR #932, opened 2024-03-15
Feat/54 add an introduction textarea

grnd-alt:
* Resolves: #54 
* Target version: main

### Summary
introduce a description editor at the top similar to Files
#### Screenshots:

Viewing a Table:
image

Creating a Table:
image

Editing a Table:
image

### Checklist

- [ ] Code is properly formatted
- [ ] Sign-off message is added to all commits
- [ ] Tests (unit, integration, api and/or acceptance) are included
- [ ] Documentation (manuals or wiki) has been updated or is not required

### DIFF
--- a/lib/Migration/Version000900Date20240314000000.php
+++ b/lib/Migration/Version000900Date20240314000000.php
@@ -27,13 +27,14 @@ class Version000900Date20240314000000 extends SimpleMigrationStep {
 
 		if ($schema->hasTable('tables_tables')) {
 			$table = $schema->getTable('tables_tables');
-			$table->addColumn('description', Types::TEXT, [
-				'default' => '',
-				'notnull' => false,
-			]);
+			if (!$table->hasColumn('description')) {
+				$table->addColumn('description', Types::TEXT, [
+					'default' => '',
+					'notnull' => false,
+				]);
+			}
 			return $schema;
 		}
 		return null;
 	}
-
 }
--- a/lib/Migration/Version000900Date20240314000000.php
+++ b/lib/Migration/Version000900Date20240314000000.php
@@ -1,0 +1,39 @@
+<?php
+
+/** @noinspection PhpUnused */
+
+declare(strict_types=1);
+
+namespace OCA\Tables\Migration;
+
+use Closure;
+use OCP\DB\Exception;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version000900Date20240314000000 extends SimpleMigrationStep {
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 * @throws Exception
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('tables_tables')) {
+			$table = $schema->getTable('tables_tables');
+			$table->addColumn('description', Types::TEXT, [
+				'default' => '',
+				'notnull' => false,
+			]);
+			return $schema;
+		}
+		return null;
+	}
+
+}
--- a/src/modules/main/sections/TableDescription.vue
+++ b/src/modules/main/sections/TableDescription.vue
@@ -1,0 +1,149 @@
+<template>
+	<div class="element-description">
+		<div class="mode-switch">
+			<div class="buttons-wrap">
+				<NcButton v-if="mode !== 'edit' && canManageElement(activeElement)" @click="() => mode='edit'">
+					<template #icon>
+						<IconPencil :size="15" />
+					</template>
+					<template #default>
+						{{ t('tables','Edit') }}
+					</template>
+				</NcButton>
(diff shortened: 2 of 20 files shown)


juliushaertl:
Very nice, two thoughts that might be worth to discuss with @nextcloud/designers :
- The editor is not collaborative, so maybe we want to make it read only by default with a button to toggle to edit
- From my perspective it would be nicer to left-align as we do it in the files app with the folder description

jancborchardt (commented):
> - The editor is not collaborative, so maybe we want to make it read only by default with a button to toggle to edit
> - From my perspective it would be nicer to left-align as we do it in the files app with the folder description

Agree with both of @juliushaertl’s points. Actually a great example of how this could look is the Forms app:

## View mode
image

## Edit mode
image

jancborchardt (changes requested):
@grnd-alt nice, thanks for both the suggestions! :)

- A little bit of moving the description can already achieve a lot, like below when the text is left-aligned with the text of the table below (see modified screenshot below)
- The radio button is a bit huge there and not super needed. I would say an "Edit"/"Done" button in the top right like we have in Collectives is better in this case.
- Is "Hide" a requirement? If so, we can do that via a simple collapsing arrow like we do in Collectives too for the "Recent pages" section
- Can we somehow hide the "Show outline" button in view mode, it is not needed here. cc @juliushaertl
 
316519952-a2efb4bc-ee82-44fd-aa80-f4408ccef8ed

src/modules/main/sections/TableDescription.vue
```diff
@@ -0,0 +1,149 @@
...
+						return
+					}
+					this.description = markdown
+					this.updateDescription()
+				},
+			})
+			this.editor.setReadOnly(true)
+		},
+		async saveDescription() {
+			if (this.descriptionLastEdited !== 0 || this.description === this.activeElement.description || !this.canManageElement(this.activeElement)) return
```
juliushaertl, line 94: Generally preferred code style around the Nextcloud code base and in this case also more readable I think ;)
```suggestion
			if (this.descriptionLastEdited !== 0 || this.description === this.activeElement.description || !this.canManageElement(this.activeElement)) {
				return
			}
```

src/modules/main/sections/TableDescription.vue
```diff
@@ -0,0 +1,149 @@
...
+			this.descriptionSaving = true
+			await this.$store.dispatch('updateTableProperty', { id: this.activeElement.id, data: { description: this.description }, property: 'description' })
+			this.descriptionLastEdit = 0
+			this.descriptionSaving = false
+		},
+		updateDescription() {
+			this.descriptionLastEdit = Date.now()
+			clearTimeout(this.descriptionSaveTimeout)
+			this.descriptionSaveTimeout = setTimeout(async () => {
```
juliushaertl, line 103: Out of curiosity: is setTimeout actually handling async functions properly?

juliushaertl (changes requested):
Small comments, otherwise this looks good 👍

src/modules/main/sections/TableDescription.vue
grnd-alt: I don't think setTimeout actually cares what type of function is passed. After the timer has expired it just calls it and has no influence on it anyways. Though wrapping the actual function call in a async arrow function was useless in this case, I removed that.

enjeck:
@jancborchardt I find it very confusing to have an edit button just for the single "description" field. Then a separate button that edits the title in a separate place. Then no way to add a description when creating a table. As a user, I would expect to be able to add a description when creating a table. And to either have a single button to edit everything at once or to be able to inline edit everything. Having a mix feels weird. Maybe it's just me

jancborchardt:
> I find it very confusing to have an edit button just for the single "description" field. Then a separate button that edits the title in a separate place. Then no way to add a description when creating a table. As a user, I would expect to be able to add a description when creating a table. And to either have a single button to edit everything at once or to be able to inline edit everything. Having a mix feels weird. Maybe it's just me

@enjeck yeah you are right – so then should we move the description box editing (and adding) into the modal? And on the tables page we show it read-only, just like we do for the title?

lib/Migration/Version000900Date20240314000000.php
```diff
@@ -0,0 +1,39 @@
...
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 * @throws Exception
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('tables_tables')) {
```
blizzz, line 28: It should also check whether it has the column already, and only add it if not present.

Also, in the migration file where this table is added, the column has to be introduced, too.

blizzz (changes requested):

src/store/store.js
```diff
@@ -262,11 +262,27 @@ export default new Vuex.Store({
 			}
 			return true
 		},
+		async updateTableProperty({ state, commit, dispatch }, { id, data, property }) {
```
enjeck, line 265: I can't find where this function is being used? Maybe it is no longer needed?

grnd-alt: You're right that's an artifact from when the description was editable in the tables view.

enjeck:
This looks good to me too. Thanks, @grnd-alt

blizzz:
@juliushaertl your change request is still active

juliushaertl (approved):

Issue #898, opened 2024-03-03
Applications: Transfer ownership


juliushaertl:
- [x] Backend API
- [x] Frontend implementation

blizzz:
Once the Context is created the connected tables and views are untangled from the permissions of the owner.  So that all should continue to work. So the design. If it does not, it's a bug.

blizzz:
> > Once the Context is created the connected tables and views are untangled from the permissions of the owner. So that all should continue to work. So the feature spec. If it does not, it's a bug.
> 
> I checked phpmyadmin, and the ownership is still with the original owner. And I'm unable to access tables from transfered contexts using the `table#index` or `table#show` routes 😕

Untangled in the meaning that it table and view ownership or management permissions turn irrelevant (as long as the resource is connected to the Context)

blizzz:
Oh I am starting to understand. 

Checking out your feature branch, the ownership is transferred. But obviously the resources cannot be loaded :man_facepalming: 

I understand all tables and views (at least their definitions) are loaded up front, for I do not see subsequent requests. 

Options:

- change view and tables endpoint (or permission service) to also return tables and view that are available via Contexts. They would appear in the sidebar than as well, at least if we do not provide further information about their source so the frontend can filter them out.
-lazy load requests when opening a contexts, but would require a different endpoint. would not be very swift experience in the user interface.
- additional endpoint to load those tables and views subsequent (or parallel) to the regular fetching of tables and views. 

What's your take?

blizzz:
> > * lazy load requests when opening a contexts, but would require a different endpoint. would not be very swift experience in the user interface.
> 
> I was trying this out. I couldn't use the `table#show` route, as it fails with permission error. Would I need to create something custom for this?

That's on me, on the backend side,

blizzz:
Please try with this patch as quick fix:

```diff
diff --git a/lib/Service/PermissionsService.php b/lib/Service/PermissionsService.php
index 922bd468..98d0543b 100644
--- a/lib/Service/PermissionsService.php
+++ b/lib/Service/PermissionsService.php
@@ -242,7 +242,20 @@ class PermissionsService {
    public function canReadColumnsByTableId(int $tableId, ?string $userId = null): bool {
        $canReadRows = $this->checkPermissionById($tableId, 'table', 'read', $userId);
        $canCreateRows = $this->checkPermissionById($tableId, 'table', 'create', $userId);
-       return $canCreateRows || $canReadRows;
[...]

PR #947, opened 2024-03-21
enh(Contexts): extend permission check

blizzz:
- checkPermission(ById) takes inherited permissions via Context into account
- adds query to fetch all contexts containing a certain node
- adds permission constants to Application

contributes to #898


enjeck (approved):
Tested and works. Thank you!

PR #952, opened 2024-03-25
feat: edit/create rows from tables interactive content widget

### LINKED ISSUE #879: Interactive widget: Create and edit rows
Currently our interactive widget of tables content is quite limited as it only lists rows and allows filtering. To contribute to the interactive widgets feature for 29 it would be good to have:

- A create row button that opens the row creation modal and allows users to add rows without leaving their current context
- A edit row button on each of the rows to be able to edit the data

The existing modal components should be reused, one thing I'm unsure about is if they work independently from the vuex store or if we need to have that setup as well in the reference widget entrypoint.

elzody:
### Issue(s)

Fixes #879 

### Overview
If a user has the ability to create/edit rows on a table/view, the user can also do so from the interactive table widget. The same "create row" button is used, as well as the "edit row" action button when hovering over a row. I made some edits to other components to better support this feature, such as dynamic importing of `store.js` and `data.js` and styling of the `CreateRow.vue` and `EditRow.vue` components (mostly from within `ContentReferenceWidget.vue` so as to not alter the original components elsewhere).


juliushaertl:
@elzody just a small hint, if you prefix the issue if with "fixes" or "resolves" in the description github will automatically add a reference that is also visible on the team board :)

src/modules/modals/CreateRow.vue
```diff
@@ -151,6 +156,38 @@ export default {
 </script>
 <style lang="scss" scoped>
 
+.modal__content {
+	padding: 20px;
+
+	:where(.row .space-T, .row.space-T) {
```
juliushaertl, line 162: I think we don't need the `:where` here as with scss those would automatically be converted to `.modal__content .row .space-T, .modal__content .row.space-T`

src/modules/modals/EditRow.vue
```diff
@@ -154,11 +157,15 @@ export default {
 			this.actionCancel()
 		},
 		async sendRowToBE() {
+			if (!this.$store) {
```
juliushaertl, line 160: Maybe just move this check into the loadStore method?

src/modules/modals/EditRow.vue
```diff
@@ -190,6 +201,44 @@ export default {
 			this.localLoading = false
 			this.actionCancel()
 		},
+		async loadStore() {
+			const { default: store } = await import(/* webpackChunkName: 'store' */ '../../store/store.js')
```
juliushaertl, line 205: In theory we could also make the store available in the widgets vue instance, but also good as it is as using the full store might also require some bigger changes then to make it aware of a tables app context that is not the full app.

src/views/ContentReferenceWidget.vue
```diff
@@ -66,13 +67,98 @@ export default {
...
+				...this.richObject,
+				isView: this.richObject.link.includes('view'),
+			},
+		}
+	},
+
+	computed: {
+		tableId() {
+			const id = this.richObject.link.split('/').at(-2)
+			return parseInt(id)
```
juliushaertl, line 82: We probably should just reuse the parsing that the backend already does. Currently we do not expose this but we can just use a patch like this to have the id and type available within the richObject

```diff
diff --git a/lib/Reference/ContentReferenceHelper.php b/lib/Reference/ContentReferenceHelper.php
index 610282c5..2f070d09 100644
--- a/lib/Reference/ContentReferenceHelper.php
+++ b/lib/Reference/ContentReferenceHelper.php
@@ -5,6 +5,7 @@ namespace OCA\Tables\Reference;
 use Exception;
 use OC\Collaboration\Reference\LinkReferenceProvider;
 use OCA\Tables\AppInfo\Application;
+use OCA\Tables\Db\Table;
 use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Errors\PermissionError;
@@ -95,6 +96,9 @@ class ContentReferenceHelper extends ReferenceHelper {
                        $reference = new Reference($referenceText);
                        $referenceInfo = [];

+                       $referenceInfo['id'] = $elementId;
+                       $referenceInfo['type'] = $element instanceof Table ? Application::NODE_TYPE_TABLE : Application::NODE_TYPE_VIEW;
+
                        if ($element->getEmoji()) {
                                $reference->setDescription($element->getEmoji() . ' ' . $element->getTitle());
                                $referenceInfo['title'] = $element->getTitle();
```

src/views/ContentReferenceWidget.vue
```diff
@@ -66,13 +67,98 @@ export default {
 		},
 	},
 
+	data() {
+		return {
+			element: {
+				...this.richObject,
+				isView: this.richObject.link.includes('view'),
```
juliushaertl, line 74: Also see the patch below for the type

enjeck:
Might be a dumb question, but what is the "interactive content widget"? Is that what pops up when I use the smartpicker? 
I ask because I have no idea how to view what this PR changes :sweat_smile:

src/views/ContentReferenceWidget.vue
```diff
@@ -66,13 +67,98 @@ export default {
...
 		useResizeObserver(this.$el, (entries) => {
 			const entry = entries[0]
 			const { width } = entry.contentRect
 			this.$el.style.setProperty('--widget-content-width', `${width}px`)
 		})
 	},
+
+	methods: {
+		async createRow() {
+			if (!this.$store) {
```
juliushaertl, line 112: Ah i think my earlier comment I misunderstood what you're doing here. 

We can just register the store in https://github.com/nextcloud/tables/blob/5c8f63e7fabdc3b3d34dc73398479dfd87162032/src/reference.js#L27 instead and avoid all the extra logic as we then just have it available

src/views/ContentReferenceWidget.vue
```diff
@@ -66,13 +67,98 @@ export default {
...
+			const { default: CreateRow } = await import('../modules/modals/CreateRow.vue')
+			spawnDialog(CreateRow, {
+				showModal: true,
+				columns: this.element.columns,
+				isView: this.element.isView,
+				elementId: this.tableId,
+			}, async () => {
+				const storeRows = this.$store.data.state.rows
+
+				if (storeRows.length > this.element.rows.length) {
```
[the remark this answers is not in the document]
juliushaertl: However if we load the rows from the backend anyways further down we could also just use those instead of the ones that the richObject holds

lib/Reference/ContentReferenceHelper.php
```diff
@@ -108,6 +108,8 @@ public function resolveReference(string $referenceText): ?IReference {
 
 			$referenceInfo['ownership'] = $element->getOwnership();
 			$referenceInfo['ownerDisplayName'] = $element->getOwnerDisplayName();
+			$referenceInfo['isShared'] = $element->getIsShared();
+			$referenceInfo['onSharePermissions'] = $element->getOnSharePermissions();
```
juliushaertl, line 112: Couldn't see this being used, is it somewhere?

juliushaertl (commented):
See inline :)

juliushaertl:
> Might be a dumb question, but what is the "interactive content widget"? Is that what pops up when I use the smartpicker?
I ask because I have no idea how to view what this PR changes 😅

When you open the smart picker and insert a table you will first see a custom picker element (https://github.com/nextcloud/tables/blob/5c8f63e7fabdc3b3d34dc73398479dfd87162032/src/reference.js#L36-L47) which is unrelated to the PR but just for understanding. The widgets then define the rendering of a preview for a link inserted through the smart picker or manually. We already have [those widgets for a while](https://github.com/nextcloud/tables/blob/5c8f63e7fabdc3b3d34dc73398479dfd87162032/src/reference.js#L20-L34) but now make them more interactive by adding the ability to create/edit rows directly from the link preview widget. Hope that makes it a bit clearer :)

lib/Reference/ContentReferenceHelper.php
elzody: > Couldn't see this being used, is it somewhere?

Yeah, so the `NcTable` component passes some permissions like this to child components, and it turns out in the widget this information isn't actually there, so I added those to the response from the backend so we can pass it to the children from the widget component. Maybe there's a better way?

juliushaertl: No sounds good then

src/views/ContentReferenceWidget.vue
```diff
@@ -21,36 +21,38 @@
 
 <template>
 	<div v-if="richObject" class="tables-content-widget">
-		<h2>{{ richObject.emoji }}&nbsp;{{ richObject.title }}</h2>
+		<h2>{{ element.emoji }}&nbsp;{{ element.title }}</h2>
```
max-nextcloud, line 24: So the `v-if` is still on `richObject` but the content is based on `element`. Is that on purpose?

src/views/ContentReferenceWidget.vue
```diff
@@ -66,13 +68,102 @@ export default {
 		},
 	},
 
+	data() {
+		return {
+			element: {
+				...this.richObject,
+			},
```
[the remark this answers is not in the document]
elzody: It seems like when I don't do this and use `richObject` directly, it's not reactive: it actually updates the properties, but doesn't reflect it in the DOM. Also I thought it's not safe to modify props from child components anyway? So I think what I was doing here is storing the prop in the component's `data`, so that it becomes reactive and it can safely be modified.

This may be a stupid question by now, but I'm pretty sure Tables uses Vue 2 still, yeah? If so, I found the [One-Way Data Flow](https://v2.vuejs.org/v2/guide/components-props#One-Way-Data-Flow) documentation provides some insight into this.

Let me know your thoughts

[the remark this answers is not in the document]
elzody: It seems like `SmartPicker.vue` actually is what is using the `ContentReferenceWidget.vue` component, so maybe we can instead have the widget emit an event or something and passes the new/edited row up to the parent, so it can then be reflected using the `richObject` prop?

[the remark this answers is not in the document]
max-nextcloud: I think my comment was quite vague. Reactivity can be a lot of things. I was referring to `element` not getting updated when `richObject` changes. So the one way data flow down into the component being interrupted. This might be what you want if you want to change the data inside the component so it becomes different from the prop. That's often the case in forms. You seed the form with the initial content and then the user can edit them independently before submitting it. Maybe that's what you are doing here?

> Tables uses Vue 2 still, yeah?
Yes.

I don't really understand the use of the view here. Maybe we can briefly talk about that later. That's probably easier.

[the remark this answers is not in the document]
elzody: > You seed the form with the initial content and then the user can edit them independently before submitting it. Maybe that's what you are doing here?

Yeah basically. I've since changed it to not copy the entire `richObject` prop, but instead just the rows, because that is the only data which changes. The goal was to make it this way so that if you edit or create a row from the widget, the new or edited row will be reflected immediately in the widget without requiring a page reload.

webpack.config.js
enjeck, line 1: Don't forget to update `package.json`'s scripts to point to `webpack.config.js` instead of `webpack.js`

elzody:
I don't think the failing Cypress tests are related, they are also failing on main it seems. Unsure what caused it 😔

mejo- (dismissed):
I tested it in Collectives and faced some issues:

- In document editors with limited width (670px per default in Text and Collectives) the styling looks a bit off when scrolling completely to the right. See screenshot below.
  * There's a big padding right to the table that allows scrolling further than required.
  * the most right column (with the edit buttons) doesn't have a visible border to the right, but the line that splits heading rows from content rows extends over this most right column. grafik

- I'm able to edit the table in place, but when I reload the page afterwards, my changes are gone. Both changed to existing columns and added new columns. I see a `PUT` request to the tables app when saving changes in the widget, but for some reason it's not persisted.

- In read-only mode I'm still able to edit the table in place, this is unexpected. Maybe it's better to display the non-interactive version there?

I don't know the code base well enough to fully understand each line of code change, but changes look sensible to me.

elzody:
@mejo- Thanks for bringing these up. A lot of the styling issues were present before I even started; I tried to fix some up as best I could, but there are definitely still some present. I will look at it some more

elzody:
I made some styling changes which I think make the widget look better than before. Specifically, I made the following changes:

- The table headers for the rich text editor fields were misaligned, because the action button does not have any available actions due to it being a widget and less sophisticated than opening the actual table. I made it so that the spot where the action button goes will always take up a 44px * 44px space, aligning it with the other headers. This was the only way I knew how to do it after thinking on it for several hours; flexbox was NOT working properly there.
- The extra space off to the right of the table was caused by the pagination footer, so I adjusted the space it takes up when rendered in a widget like this. It no longer takes up extra space, that I can see anyway.
- There was some transparent space in the widget header off to the right, which has been fixed now by adding an element to group the table title and options row. They also stick properly now to the top.

The other things that you mentioned were not reproducible on my end. I tried the following:

- When I add or edit a row, it is reflected immediately in the widget. If I reload normally, the changes are still there, as well as in the tables app. This also goes for a hard reload of the page.
- I tested the permissions thing briefly, by sharing it with a non-admin account without write/update permissions for the table. At first, the buttons were still there, but after submitting the new/altered content, an error popped up stating I did not have sufficient permissions to update the table. When I did a hard reload on the page, the buttons disappeared. My guess is the browser caches the JS related to this, but not sure about it yet.

Also, I rebased onto main and resolved a small merge conflict.

@mejo-

juliushaertl (approved):

juliushaertl:
Pushed a small backend commit to ensure that the rich text fields are properly rendered, otherwise 👍

mejo- (commented):
Thanks @elzody.

The CSS/style fixes look great, much better now :ok_hand:

Regarding the problems with editing from Collectives, it's still the same:
* Editing is possible in edit mode, and it alters the table. When I open the table in the Tables app afterwards, the changes are visible. But for some reason, after reloading the collectives page, the old state of the table before the change is displayed.
* Editing in view mode is still possible. When switching to view mode in Collectives, all the buttons to edit the table in place are still there, and when editing the table the changes are pushed to the server.

I recorded a screencast to make more clear what I mean:

| Screencast |
| --- |
| recording1 |

juliushaertl:
> Editing in view mode is still possible. When switching to view mode in Collectives, all the buttons to edit the table in place are still there, and when editing the table the changes are pushed to the server.

I'd not consider this one a blocker for now. We also have this behaviour in other interactive widgets and it would require larger work to make the context of the widget exposed to the rendering provider.

juliushaertl (approved):

Issue #956, opened 2024-03-28
Write critical operations to the audit log


juliusknorr:
### Is your feature request related to a problem? Please describe.

For compliance reasons there should be the possibility to log critical operations into a log file.

### Describe the solution you'd like

The server provides an admin_audit app which can be used to track events that apps emit and write it into the audit log

We should implement audit logging for the tables app for the following actions as a starting point:
- A table or view is deleted
- A row is deleted
- A tables ownership is transferred

### Describe alternatives you've considered

Writing to the regular Nextcloud log is not feasible as those log messages are rarely relevant to regular administrators

### Additional context

Setup documentation: https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/logging_configuration.html#admin-audit-log
Developer documentation: https://docs.nextcloud.com/server/latest/developer_manual/basics/logging.html#admin-audit-logging

PR #959, opened 2024-04-01
feat: Write critical operations to the audit log

hweihwang:
This PR introduces audit logging for critical operations in the Tables app, addressing [issue #956](https://github.com/nextcloud/tables/issues/956). The goal is to enhance compliance measures by logging critical operations into a dedicated log file.

## Changes
* Introduced `AuditLogServiceInterface`, which outlines the contract for audit logging within the application. The `DefaultAuditLogService` is the default implementation that dispatches `CriticalActionPerformedEvent` to interact with the AuditLog App. This service is auto-resolved by registering it in the Dependency Injection Container `Application.php`. This design allows for future extensibility, such as integrating different logging methods or services for AuditLog.

* Developed event dispatchers for corresponding events to trigger audit logs (`TableDeletedEvent` => `WhenRowDeletedAuditLogListener`, `ViewDeletedEvent` => `WhenViewDeletedAuditLogListener`, etc.). This is a standard event-driven approach that decouples auxiliary logic (logging, mail sending, etc.) from the core business logic, effectively handling side effects. Most of these operations can be executed asynchronously via a queue. Each Listener is encapsulated in its own class adhering to the Single Responsibility Principle. The `AuditLogServiceInterface` is injected into each listener to further promote decoupling.

## Testing
Unit tests have been added for the new service and listeners. These tests can be run by executing into the server container and running `composer test`.

## Potential Improvements
* Implementing a separate background queue worker for all listeners could offload non-business logic from the main service logic, improving performance, especially when dealing with external API calls or IO intensive operations.

* Applying Database Transactions: The TableService interacts with multiple tables simultaneously (Table, View, Row, Column). Implementing DB transactions and error handling mechanisms can ensure the integrity and reliability of operations.

Please review and provide your feedback, thank you @juliushaertl!

### DIFF
--- a/lib/Db/LegacyRowMapper.php
+++ b/lib/Db/LegacyRowMapper.php
@@ -116,7 +116,12 @@ class LegacyRowMapper extends QBMapper {
 		return $innerFilterExpressions;
 	}
 
-	private function getFilterGroups($qb, $filters): array {
+	/**
+	 * @param (float|int|string)[][][] $filters
+	 *
+	 * @psalm-param non-empty-list<list<array{columnId: int, operator: 'begins-with'|'contains'|'ends-with'|'is-empty'|'is-equal'|'is-greater-than'|'is-greater-than-or-equal'|'is-lower-than'|'is-lower-than-or-equal', value: float|int|string}>> $filters
+	 */
+	private function getFilterGroups(IQueryBuilder $qb, array $filters): array {
 		$filterGroups = [];
 		foreach ($filters as $groupIndex => $filterGroup) {
 			$filterGroups[] = $qb->expr()->andX(...$this->getInnerFilterExpressions($qb, $filterGroup, $groupIndex));
@@ -149,7 +154,12 @@ class LegacyRowMapper extends QBMapper {
 		}
 	}
 
-	private function addOrderByRules(IQueryBuilder $qb, $sortArray) {
+	/**
+	 * @param (int|string)[][] $sortArray
+	 *
+	 * @psalm-param list<array{columnId: int, mode: 'ASC'|'DESC'}> $sortArray
+	 */
+	private function addOrderByRules(IQueryBuilder $qb, array $sortArray) {
 		foreach ($sortArray as $index => $sortRule) {
 			$sortMode = $sortRule['mode'];
 			if (!in_array($sortMode, ['ASC', 'DESC'])) {
--- a/lib/Db/LegacyRowMapper.php
+++ b/lib/Db/LegacyRowMapper.php
@@ -157,7 +157,7 @@ class LegacyRowMapper extends QBMapper {
 	/**
 	 * @param (int|string)[][] $sortArray
 	 *
-	 * @psalm-param list<array{columnId: int, mode: 'ASC'|'DESC'}> $sortArray
+	 * @psalm-param list<array{columnId?: int, columnType?: string, mode?: 'ASC'|'DESC'}> $sortArray
 	 */
 	private function addOrderByRules(IQueryBuilder $qb, array $sortArray) {
 		foreach ($sortArray as $index => $sortRule) {
--- a/lib/Event/ViewDeletedEvent.php
+++ b/lib/Event/ViewDeletedEvent.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OCA\Tables\Event;
+
+use OCA\Tables\Db\View;
+use OCP\EventDispatcher\Event;
+
+final class ViewDeletedEvent extends Event
+{
+    public function __construct(protected View $view, protected string $userId)
+    {
+        parent::__construct();
+    }
+
+    public function getView(): View
+    {
+        return $this->view;
+    }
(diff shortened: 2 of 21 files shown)


lib/Db/LegacyRowMapper.php
```diff
@@ -149,7 +154,12 @@ private function resolveSearchValue(string $unresolvedSearchValue, string $userI
 		}
 	}
 
-	private function addOrderByRules(IQueryBuilder $qb, $sortArray) {
+	/**
+	 * @param (int|string)[][] $sortArray
+	 *
+	 * @psalm-param list<array{columnId: int, mode: 'ASC'|'DESC'}> $sortArray
```
juliushaertl, line 160: This doesn't seem related to the implementation but fine to annotate here, just the static analysis fails due to the incomplete array description as the array gets enriched with a columnType key

> Error: lib/Db/LegacyRowMapper.php:184:25: InvalidArrayOffset: Cannot access value on variable $sortRule using offset value of 'columnType', expecting 'columnId' or 'mode' (see https://psalm.dev/115)

lib/Event/ViewDeletedEvent.php
```diff
@@ -0,0 +1,22 @@
...
+	public function __construct(protected View $view, protected string $userId) {
+		parent::__construct();
+	}
+
+	public function getView(): View {
+		return $this->view;
+	}
+
+	public function getUserId(): string {
+		return $this->userId;
```
juliushaertl, line 20: We could probably leave passing the user id to every event as it can always be obtained from the user session and enhance the log message with it from the AuditLogService

lib/Listener/WhenViewDeletedAuditLogListener.php
```diff
@@ -0,0 +1,32 @@
...
+		if (!($event instanceof ViewDeletedEvent)) {
+			return;
+		}
+
+		$view = $event->getView();
+		$userId = $event->getUserId();
+
+		$this->auditLogService->log("View with ID: $view->id was deleted by user with ID: $userId", [
+			'view' => $view->jsonSerialize(),
+			'userId' => $userId,
```
juliushaertl, line 29: As commented, we can probably skip this. The userId should be already in place in the log entry for any message (https://github.com/nextcloud/server/blob/d753d0476053fc7859e60fe97670a4299dd8fa32/lib/private/Log/LogDetails.php#L74)

juliushaertl (commented):
Thanks a lot for this and sorry for taking so long to check. This looks very good to me already. The conceptual changes make a lot of sense to have events introduced for those actions and while we might not have a immediate use case for the service contract, this could become quite useful in the future.

I left a few minor comments, maybe you can check those. Looking forward to your updates/comments on that.

lib/Db/LegacyRowMapper.php
hweihwang: @juliushaertl Thank you for spot on! Wondering why the ci/cd checks keeps failing (Not immediately but after few days) 

Changed to this: @psalm-param list<array{columnId?: int, columnType?: string, mode?: 'ASC'|'DESC'}> $sortArray

Because when running psalm at my local there a case only 2 of them present (columnId & mode)

lib/Listener/WhenViewDeletedAuditLogListener.php
hweihwang: Thank you and Totally agreed with this duplication after reading the Log Service in details

lib/Event/ViewDeletedEvent.php
hweihwang: Really good take, initially I thought the Event is more general purpose than just logging and always need to wrap all the necessary data because it knows nothing about the Listener classes (Logging logic) and because in the future there’s some other Listeners observing it (not only logging) so they might need those data (userId) but yes we can always get that from global session one (But yeah with some pattern like distributed event driven system, user might not be accessible globally haha)


Definitely agree with you the Logging Listener no need the duplication of userId there because it’s already on the Logging Service

lib/Db/LegacyRowMapper.php
juliushaertl: The CI requires approval for outside contributors.

I'm also unsure why that may not happen for you locally, just checked and I get the same error. The only thing I could imagine is psalm cache which sometimes messes things up. You could try running with `composer run psalm -- --no-cache --threads=8` and see if that makes a difference.

lib/Listener/WhenViewDeletedAuditLogListener.php
juliushaertl: Thanks for the quick adjustment. Looks good now 👍

lib/Event/ViewDeletedEvent.php
juliushaertl: Yes, something we can probably consider once we come up with use cases that require that. So far we don't have a global event bus system in Nextcloud nor we do have code paths that would require calling operations in scope of other users within tables, but thanks for bringing up that topic and explaining the background.

juliushaertl (approved):

Issue #897, opened 2024-03-03
Applications: Sharing


juliushaertl:
- [x] API endpoints @blizzz https://github.com/nextcloud/tables/pull/962
- [x] Frontend  @enjeck https://github.com/nextcloud/tables/pull/966

blizzz:
all PRs merged

PR #962, opened 2024-04-02
feat(Context): add share logic for contexts

blizzz:
contributes to #311 and #897

* `createShare` endpoints where extended with the optional parameter `displayMode`, which is only effective for contexts. The permissions flags on the other hand are not effective for contexts at the moment.
* `updateShareDisplayMode` (Routes: `api1#updateShareDisplayMode` and `share#updateDisplayMode`) endpoint was added, as the only updatable endpoint was against purely against permissions.
* business logic was adjusted to supply contexts specific needs.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -33,6 +33,7 @@ return [
 		['name' => 'api1#createShare',	'url' => '/api/1/shares', 'verb' => 'POST'],
 		['name' => 'api1#deleteShare',	'url' => '/api/1/shares/{shareId}', 'verb' => 'DELETE'],
 		['name' => 'api1#updateSharePermissions',	'url' => '/api/1/shares/{shareId}', 'verb' => 'PUT'],
+		['name' => 'api1#updateShareDisplayMode',	'url' => '/api/1/shares/{shareId}/display-mode', 'verb' => 'PUT'],
 		['name' => 'api1#createTableShare',	'url' => '/api/1/tables/{tableId}/shares', 'verb' => 'POST'],
 		// -> columns
 		['name' => 'api1#indexTableColumns',	'url' => '/api/1/tables/{tableId}/columns', 'verb' => 'GET'],
@@ -97,6 +98,7 @@ return [
 		['name' => 'share#show', 'url' => '/share/{id}', 'verb' => 'GET'],
 		['name' => 'share#create', 'url' => '/share', 'verb' => 'POST'],
 		['name' => 'share#updatePermission', 'url' => '/share/{id}/permission', 'verb' => 'PUT'],
+		['name' => 'share#updateDisplayMode', 'url' => '/share/{id}/display-mode', 'verb' => 'PUT'],
 		['name' => 'share#destroy', 'url' => '/share/{id}', 'verb' => 'DELETE'],
 
 		// import
--- a/lib/Controller/Api1Controller.php
+++ b/lib/Controller/Api1Controller.php
@@ -34,6 +34,7 @@ use Psr\Log\LoggerInterface;
  * @psalm-import-type TablesColumn from ResponseDefinitions
  * @psalm-import-type TablesRow from ResponseDefinitions
  * @psalm-import-type TablesImportState from ResponseDefinitions
+ * @psalm-import-type TablesContextNavigation from ResponseDefinitions
  */
 class Api1Controller extends ApiController {
 	private TableService $tableService;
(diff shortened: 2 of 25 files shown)


enjeck (commented):
@blizzz I guess you need to add `sharing` to the create and update functions at `ContextController.php`?

blizzz:
> @blizzz I guess you need to add `sharing` to the create and update functions at `ContextController.php`?

The existing share endpoints were extended instead.

enjeck:
> > @blizzz I guess you need to add `sharing` to the create and update functions at `ContextController.php`?
> 
> The existing share endpoints were extended instead.

Sorry, I don't understand how it works. The way I understand it, I can use the existing endpoint by making a POST request to `generateUrl('/apps/tables/share')` and sending data of a format similar to this:
```
{
  nodeType: 'context',
  nodeId: <context id>,
  receiver: <user id>,
  receiverType: 'user',
  permissionRead: true,
  permissionCreate: true,
  permissionUpdate: true,
  permissionDelete: false,
  permissionManage: false,
}
```

But I can't see how the above approach would fit a context share.

I took this screenshot from https://github.com/nextcloud/tables/issues/311:
Screenshot from 2024-04-09 15-27-30

As shown image above, I expect a context share to support multiple receivers, and can have different permissions for each resource. So I expect `receiver` in the object above to be an array or similar. And to somehow have each resource mapped to separate permissions. 

Am I thinking about this incorrectly?

blizzz:
The permissions are (at least for now) irrelevant, so leave the at false, or do not specify them at all – they are optional. Additionally you can set the default displayMode to 0 (default), 1 (navigation entry for recipients) or 2 (navigation entry for all, i.e. recipients and owner). For example:

```json
{
  nodeType: 'context',
  nodeId: <context id>,
  receiver: <user id>,
  receiverType: 'user',
  displayMode: 2
}
```

The permissions from the draft are node-specific and unrelated to the shares! So "My open request" and "Approved" are tables or views and their permissions are set when you are creating the Context (or updating it). You remember the payload from the create Context API 

```json
{
  "name": "Enchanting Pancake",
  "iconName": "airline_seat_recline_extra",
  "description": "comparison of pancake receipes, ingredients, cafés and cooks.",
  "nodes": [
    {
      "id": 4,
      "type": "0",
      "permissions": 7
    },
    {
      "id": 2,
      "type": "1",
      "permissions": 4
    }
  ]
}
```

The permissions in the node specification are the bitmask with the values from these constants: https://github.com/nextcloud/tables/blob/main/lib/AppInfo/Application.php#L38-L43

So far the interpretation of the concept is that all nodes are shared with the same permissions to  all recipients. To quote @juliushaertl 

> There is a set of default permissions that are applied to all share recipients for now [source](https://github.com/nextcloud/tables/issues/311#issuecomment-1790569092)

appinfo/routes.php
```diff
@@ -33,6 +33,7 @@
 		['name' => 'api1#createShare',	'url' => '/api/1/shares', 'verb' => 'POST'],
```
enjeck, line 33: @blizzz So, now I can create a share. But how do I get all the shares associated with a particular context? Do we need an endpoint like this one, but for contexts?

Also, since a context has a `sharing` property, I would have expected that all shares would be listed here. But even after adding shares using `axios.post(generateUrl('/apps/tables/share'), share)`, the `sharing` array is still empty. 


```
context = {
	'id'
	'name'
	'iconName'
	'description'
	'owner'
	'ownerType'
	'sharing'
	'nodes' 
	'pages' 
}
```

blizzz: Thanks for pointing this out. The reason for an empty sharing value was that the user initiated shares were not queried. Only those where the current  users is a receiver either directly, or by group. 

I just pushed a commit that would fetch share information when the user is also the sender.

For only Context owners can share the context, this should actually changed to fetch all shares when the acting user is the owner. That's a follow up then.

lib/Controller/Api1Controller.php
```diff
@@ -573,6 +586,55 @@ public function updateSharePermissions(int $shareId, string $permissionType, boo
...
+	 * 200: Display mode updated
+	 * 400: Invalid parameter
+	 * 403: No permissions
+	 * 404: Share not found
+	 *
+	 * @psalm-param int<0, 2> $displayMode
+	 * @psalm-param ("default"|"self") $target
+	 */
+	public function updateShareDisplayMode(int $shareId, int $displayMode, string $target = 'default'): DataResponse {
+		if ($target === 'default') {
```
juliushaertl, line 610: Could become a constant, same as self

juliushaertl (approved):
Nitpick, can be a follow up

PR #972, opened 2024-04-10
Enhancement/279 pagination for tables ui 

grnd-alt:
## Design
image

### Loading Time for 3k Rows with pagination
https://github.com/nextcloud/tables/assets/56235737/740f9864-50f6-45f6-9a94-fb48ac33d7fa

### 3k Rows Without pagination
https://github.com/nextcloud/tables/assets/56235737/f9b98ff4-0f25-4ba5-901e-3416a5d8f2ef

### DIFF
--- a/src/shared/components/ncTable/sections/CustomTable.vue
+++ b/src/shared/components/ncTable/sections/CustomTable.vue
@@ -19,7 +19,7 @@
 				</TableHeader>
 			</thead>
 			<tbody>
-				<TableRow v-for="(row, index) in getSearchedAndFilteredAndSortedRows"
+				<TableRow v-for="(row, index) in currentPageRows"
 					:key="index"
 					data-cy="customTableRow"
 					:row="row"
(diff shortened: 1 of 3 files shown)


enjeck:
I realize this is still a draft, but just wondering if this PR adds pagination **or** virtual scrolling? There seems to be a mismatch between the title and the commit message :sweat_smile:

grnd-alt:
> I realize this is still a draft, but just wondering if this PR adds pagination **or** virtual scrolling? There seems to be a mismatch between the title and the commit message 😅

Yeah it's because the initial issue planned to implement v-scrolling but that's harder to do on html tables than expected hence we decided to look into pagination

jancborchardt (changes requested):
Looks nice! :)

An improvement on the style would be to use the default color-main-background and instead use a drop shadow like the action menus have.
This would correspond better with it being a floating element above others, and also look a bit more modern than a greyish element.

grnd-alt:
@jancborchardt  Updated With New Screenshot

jancborchardt:
> @jancborchardt Updated With New Screenshot

It doesn’t look like it has the drop shadow, but that looks fine to me, as long as the background is opaque color-main-background – so the element is still readable when there is a long table below it. :)

src/shared/components/ncTable/sections/CustomTable.vue
```diff
@@ -287,7 +358,64 @@ export default {
...
+	background-color: var(--color-main-background);
+	border-radius: var(--border-radius-large);
+	pointer-events: all;
+
+	display: flex;
+	align-items: center;
+}
+
+.pagination-footer{
+	position:absolute;
```
juliushaertl, line 406: When testing with applications where multiple tables are rendered on the same page the pagination items overlap each other, my suggestion would be to use sticky positioning here and give the parent container a relative positioning.

src/shared/components/ncTable/sections/CustomTable.vue
```diff
@@ -287,7 +358,64 @@ export default {
...
+	box-shadow: var(--box-shadow);
+	filter: drop-shadow(0 1px 10px var(--color-box-shadow));
+	bottom: 10px;
```
juliushaertl, line 409: Bottom spacing could be a bit increased to avoid the drop shadow conflicting with the scrollbar from my perspective but nit picking:


Screenshot 2024-05-03 at 14 47 46

vs.

Screenshot 2024-05-03 at 14 48 16

juliushaertl (commented):
Two minor comments, otherwise very nice 🚀

juliushaertl (approved):

Issue #994, opened 2024-04-24
Applications: Frontend tests


juliushaertl:
We should cover the common user interactions for contexts with cypress:

- [x] Create a new context
- [x] Add a table to it
- [x] Add a view to it
- [x] Check that both are rendered
- [x] Share with a user
- [x] Check that the user receives the context
- [x] Check that the application menu entry shows up
- [x] Deleting an application (for both owner and shares)
- [x] Transfering an application

PR #1025, opened 2024-04-29
enh: Consolidate user/group search code

### LINKED ISSUE #1003: Consolidate and resue NcSelects
We have several NcSelects that search for users/groups duplicated across other pages. We should consolidate into just one and reuse

enjeck:
Fixes #1003


elzody (dismissed):

enjeck:
TODO
- Do Applications Cypress test first so we can use it to ensure this doesn't break Applications shares: https://github.com/nextcloud/tables/issues/994
- Decide whether to rename [NcUserAndGroupPicker.vue](https://github.com/nextcloud/tables/pull/1025/files#diff-cba9fe17aced0fbf267d15b0bf069759d63b04f2ea329aa5bfe85d7d3aa49288) for transfers only
- Refactor [ShareForm.vue](https://github.com/nextcloud/tables/pull/1025/files#diff-acecdd02d44cce35f35c8e95bbf543470c035f9cfcb3faf6ef2395ae3b92be3b)

elzody (commented):
I noticed that when I initially create a context, I can add users to the user select in the "Share with accounts" / "Share with users" field, but after it is created I cannot add any more users. I can remove users, just not use the select anymore. Same goes for when I try to transfer the context to someone else, I can't use the drop-down select.

I don't get any browser console errors, so if you can't reproduce, may just be a browser thing on my end.

enjeck:
> I noticed that when I initially create a context, I can add users to the user select in the "Share with accounts" / "Share with users" field, but after it is created I cannot add any more users. I can remove users, just not use the select anymore. Same goes for when I try to transfer the context to someone else, I can't use the drop-down select.
> 
> I don't get any browser console errors, so if you can't reproduce, may just be a browser thing on my end.

Thanks for testing. I can't reproduce the problem, it works fine for me :thinking:

elzody (approved):
Works properly for me now, looks great.

juliushaertl (approved):

PR #1026, opened 2024-04-29
enh: Delete Application and its shares

### LINKED ISSUE #991: Applications: Allow users to delete applications
Leftover from #311 

We should offer users an option to delete applications

enjeck:
Fixes #991

### DIFF
--- a/src/modules/modals/DeleteContext.vue
+++ b/src/modules/modals/DeleteContext.vue
@@ -1,0 +1,56 @@
+<template>
+	<div>
+		<DialogConfirmation :description="getTranslatedDescription"
+			:title="t('tables', 'Confirm application deletion')"
+			:cancel-title="t('tables', 'Cancel')"
+			:confirm-title="t('tables', 'Delete')"
+			confirm-class="error"
+			:show-modal="showModal"
+			@confirm="deleteContext"
+			@cancel="$emit('cancel')" />
+	</div>
+</template>
+
+<script>
+
+import DialogConfirmation from '../../shared/modals/DialogConfirmation.vue'
+import { showSuccess } from '@nextcloud/dialogs'
+import '@nextcloud/dialogs/dist/index.css'
+import { mapState } from 'vuex'
+
+export default {
+	components: {
+		DialogConfirmation,
+	},
+	props: {
+		context: {
+			type: Object,
+			default: null,
+		},
+		showModal: {
+			type: Boolean,
+			default: false,
+		},
+	},
+	computed: {
+		...mapState(['activeContextId']),
+		getTranslatedDescription() {
+			// TODO confirm that we actually do delete all the shares
+			return t('tables', 'Do you really want to delete the application "{context}"? This will also delete the shares and unshare the resources that are connected to this application.', { context: this.context?.name })
+		},
+	},
+	methods: {
+		async deleteContext() {
+			const res = await this.$store.dispatch('removeContext', { context: this.context, receivers: this.context.sharing })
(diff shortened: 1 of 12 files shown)


enjeck:
Need to check/fix some edge cases:
- [ ] Delete shares from a newly-created context without reload
- [ ] Delete shares from a context whose shares were updated
- [ ] check that transfer context works
- [ ]  check that shared resources are removed when context is deleted

src/modules/modals/EditContext.vue
```diff
@@ -209,6 +223,37 @@ export default {
 			}
 			return resources
 		},
+		async actionDeleteContext() {
+			this.prepareDeleteContext = false
+			const res = await this.$store.dispatch('removeContext', { context: this.context, receivers: this.context.sharing })
```
juliushaertl, line 228: This throws an error as this.context doesn't seem to be defined in here. The following seems to work:

```
--- a/src/modules/modals/EditContext.vue
+++ b/src/modules/modals/EditContext.vue
@@ -225,7 +225,8 @@ export default {
                },
                async actionDeleteContext() {
                        this.prepareDeleteContext = false
-                       const res = await this.$store.dispatch('removeContext', { context: this.context, receivers: this.context.sharing })
+                       const context = this.getContext(this.contextId)
+                       const res = await this.$store.dispatch('removeContext', { context, receivers: context.sharing })
                        if (res) {
                                showSuccess(t('tables', 'Application "{context}" removed.', { context: this.title }))
                                // if the active context was deleted, go to startpage
```

src/store/store.js
```diff
@@ -493,11 +494,36 @@ export default new Vuex.Store({
...
 			const contexts = state.contexts
-			const index = contexts.findIndex(t => t.id === id)
+			const index = contexts.findIndex(c => c.id === id)
 			contexts.splice(index, 1)
 			commit('setContexts', [...contexts])
 			return true
 		},
+		async removeContext({ dispatch, state, commit }, { context }) {
+			try {
+				await dispatch('unShareContext', { receivers: context.sharing })
```
juliushaertl, line 504: I would actually prefer to just send a delete request and the backend would not only delete the context but also any shares, pages, app menu entries

@blizzz Any objections? Currently the ContextService (called by ContextController::delete) just drops the context entry it seems

enjeck: it works, thanks @blizzz

src/modules/modals/EditContext.vue
enjeck: good catch, thx

enjeck:
@blizzz How can I fix the Static Analysis errors here?

blizzz:
> @blizzz How can I fix the Static Analysis errors here?

Ah, leftovers from my changes. I added a parameter to ContextService, but changed name and type during working on it. Alas not everywhere. So it's fixing, fix-upping the original commit, and rebasing with squashing it.

blizzz:
Done and rebased.

src/modules/modals/DeleteContext.vue
```diff
@@ -0,0 +1,56 @@
...
+		},
+		showModal: {
+			type: Boolean,
+			default: false,
+		},
+	},
+	computed: {
+		...mapState(['activeContextId']),
+		getTranslatedDescription() {
+			// TODO confirm that we actually do delete all the shares
```
blizzz, line 38: This TODO can go now, as it is handled in the backend.

blizzz (approved):

juliushaertl (approved):

Issue #933, opened 2024-03-15
Applications: Backend/frontend - Indicate active context in the navigation


juliushaertl:
Follow up from https://github.com/nextcloud/tables/pull/904#issuecomment-1997421405

We should:
- Have a template response endpoint that is used for applications accessed from the app menu that will set the proper active navigation icon
- We can pass over an initial state value there so the frontend knows when it has been accessed through the app navigation bar instead of just through tables directly
  - Maybe we want to hide the tables navigation by default then to focus on the current application

PR #1037, opened 2024-04-30
enh(Contexts): set active navigation entry

blizzz:
This is sort of a PoC actually. Addresses #933 

I looked at the backend options first. The problem is that we are currently working with a set anchor, i.e. *example.com/index.php/apps/tables/**#/application/28***, so that is no information that we have on the backend. We'd either have to change this behaviour, or to avoid completely reloading the Tables app, having a mechanism that works with both, an URL that can be understood by the server, and the inside-app URL with the anchor.

Nevertheless, I looked at the frontend bits for a quick win. And there is the quick win:

[Screencast_20240430_113348.webm](https://github.com/nextcloud/tables/assets/2184312/3e1bb2b4-1108-4098-8074-e048ce5216eb)

It is not ideal though, because

- the active entry is visible flicking from Tables to the Tables Application when opening the Application directly, as it is initially set by the backend, and later adjusted by the app
- there is no exposed API to manipulated the navigation bar, so I am directly manipulating the JS class. This will easily break. It might be an approach to serve NC 29, but for 30 some API by the server would be necessary. Even if the backend scenario is solved, as laid out up front, this will be needed for inside-app Application switches.

What do you think?

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -65,12 +65,15 @@ export default {
 			if (currentRoute.path.startsWith('/table/')) {
 				this.$store.commit('setActiveTableId', parseInt(currentRoute.params.tableId))
 				this.setPageTitle(this.$store.getters.activeTable.title)
+				this.switchActiveMenuEntry(document.querySelector('header .header-left .app-menu li[data-app-id="tables"]'))
 			} else if (currentRoute.path.startsWith('/view/')) {
 				this.$store.commit('setActiveViewId', parseInt(currentRoute.params.viewId))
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):
While I'd only consider that a workaround, i think this is good enough for now.

blizzz:
Rebased to resolve conflicts.

> While I'd only consider that a workaround, i think this is good enough for now.

Definitely not more than a workaround, that's what i tried to say :)

blizzz:
@juliushaertl keep it for 0.8 or also bring it down to 0.7?

PR #1038, opened 2024-04-30
fix(App): unstuck navbar focus on internal entries

blizzz:
Fixes the focus being stuck on the nav bar after switching from Tables to a Tables application. Without the fix, the labels remain to be seen, only a click fixes the state. 

Buggy behaviour:
[Screencast_20240430_112536.webm](https://github.com/nextcloud/tables/assets/2184312/8fca0601-1aed-48e2-a669-16b224a822ae)

I understand it is OK to refer to *app-content-vue* as it is already done so in a different place in the same file. Setting the tabIndex is necessary for divs to accept the focus.


juliushaertl (approved):
Seems good 👍

Issue #982, opened 2024-04-17
Analyze queries and look for missing indices


juliushaertl:
Placeholder ticket to analyze the queries that the tables app builds for general browsing and especially for view filters to see if we lack any indices or need to optimize certain queries using https://github.com/nextcloud/profiler

Let's collect findings in this ticket:

- [ ] `DELETE FROM oc_tables_row_cells_* WHERE row_id = ?` ([Code](https://github.com/nextcloud/tables/blob/42eb83b563dcdfc714842757e1734b58fa58d2e0/lib/Db/RowCellMapperSuper.php#L58))
- [ ] table_id on oc_tables_views (https://github.com/nextcloud/tables/pull/1062)
- [ ] table_id on oc_tables_columns
- [ ] table_id on oc_tables_row_sleeves
- [ ] get Columns in one query instead of n ([Code](https://github.com/nextcloud/tables/blob/285bb3b40cd4545c30fb08b19d5cfcc6e4a88b23/lib/Service/ColumnService.php#L98))
- [ ] Tables ownership index

PR #1062, opened 2024-05-06
perf: Avoid extra queries to get the view ownership

juliushaertl:
Currently we were doing n+1 queries just to get the ownership for each table.

We should also add a index for the table_id column on oc_tables_views, but tracking that in #982


blizzz (approved):

blizzz:
backport?

PR #1078, opened 2024-05-08
add table_id index to oc_tables_columns

grnd-alt:


lib/Listener/AddMissingIndicesListener.php
```diff
@@ -0,0 +1,19 @@
...
+final class AddMissingIndicesListener implements IEventListener {
+    public function __construct(protected AuditLogServiceInterface $auditLogService) {
+    }
+    public function handle(Event $event): void {
+        if (!$event instanceof AddMissingIndicesEvent) {
+            throw new \Exception('Error for testing only');
+            return;
+        }
+
+        $event->addMissingIndex('oc_tables_columns', 'table_id_index', ['table_id']);
```
juliushaertl, line 17:
```suggestion
        $event->addMissingIndex('tables_columns', 'table_id_index', ['table_id']);
```

It fails because the table prefix (`oc_`) is added by the db abstraction.

Otherwise you may want to make the index name unique across tables (postgres requires that). A common pattern could be tables_columns_t_id. Shortening that makes sense to keep oracle support possible with 30 characters max https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/database.html#supporting-more-databases

juliushaertl:
One thing I missed when discussing this earlier, we should still add the indices to the existing migration files where the tables are created. That way the index is immediately there on new installations.

lib/Migration/Version000000Date20210921000000.php
```diff
@@ -59,6 +59,8 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 				'notnull' => false,
 			]);
 			$table->setPrimaryKey(['id']);
+
+			$table->addIndex(["ownership"]);
```
juliushaertl, line 63: Would be good to also pass the index name as the second parameter

enjeck:
@grnd-alt @juliushaertl Why are we adding an index?

juliushaertl:
To improve the performance of database queries that use those columns to filter for entries.

juliushaertl (approved):

PR #1080, opened 2024-05-08
enh(App): navigation tweaks for Contexts

blizzz:
This is a more finegrained approach to to #933. The PR introduces an endpoint for each Context, so they can have the correct navigation item highlighted from the beginning, solving a known papercut from https://github.com/nextcloud/tables/pull/1037

It also sets an initial state, so that in Vue we can apply the actual route. Alas, the Vue Router adds the same information into the address bar, duplicating "app[lication]" and the context id. This is not nice. I reset it to the original URL (accepting #/), but it comes at the cost of flickering. In Vue the route cannot be applied without changing the address bar, which is a known drawback.

Another tweak: I disable the app navigation, when the Context is opened from its specific URL, making it more "standalone". Questionable if we want it. There are alternatives also: filtering the Tables and Views to only those included in the Context, and not showing applications; or (perhaps in future) a Context-specific app navigation (maybe relevant once we make use of pages).

What is not included in the Controller is a check whether a Context is actually having a navigation bar entry. So far we did not expose the setting, so all Contexts are shown there anyway. But even if not, I am not sure this is an issue (user has access anyways).

Please check the annotated video for a recorded quick tour.

https://github.com/nextcloud/tables/assets/2184312/f99ea308-9a30-4284-a62b-9ce9e8bc8cc6

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -16,6 +16,7 @@ import Navigation from './modules/navigation/sections/Navigation.vue'
 import { mapState } from 'vuex'
 import Sidebar from './modules/sidebar/sections/Sidebar.vue'
 import { useResizeObserver } from '@vueuse/core'
+import { loadState } from '@nextcloud/initial-state'
 
 export default {
 	name: 'App',
@@ -57,6 +58,19 @@ export default {
 	},
 	methods: {
 		routing(currentRoute) {
+			try {
+				if (loadState('tables', 'contextId', undefined)) {
+					// prepare route, when Context is opened from navigation bar
+					const contextId = loadState('tables', 'contextId', undefined)
+					const originalUrl = window.location.href
+					this.$router.replace('/application/' + contextId).catch(() => {})
+					// reverts turning /apps/tables/app/28 into /apps/tables/app/28#/application/28
+					history.replaceState({}, undefined, originalUrl)
+				}
+			} catch (e) {
+				// contextId is not always set, it is fine.
+			}
+
 			if (currentRoute.name === 'tableRow' || currentRoute.name === 'viewRow') {
 				this.$store.commit('setActiveRowId', parseInt(currentRoute.params.rowId))
 			} else {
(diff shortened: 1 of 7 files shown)


src/App.vue
```diff
@@ -16,6 +16,7 @@ import Navigation from './modules/navigation/sections/Navigation.vue'
 import { mapState } from 'vuex'
 import Sidebar from './modules/sidebar/sections/Sidebar.vue'
 import { useResizeObserver } from '@vueuse/core'
+import { loadState } from '@nextcloud/initial-state'
```
juliushaertl, line 19: Needs an `npm install --save @nextcloud/initial-state`

src/App.vue
```diff
@@ -57,6 +58,19 @@ export default {
 	},
 	methods: {
 		routing(currentRoute) {
+			try {
+				if (loadState('tables', 'contextId', undefined)) {
+					// prepare route, when Context is opened from navigation bar
+					const contextId = loadState('tables', 'contextId', undefined);
+					const originalUrl = window.location.href
+					this.$router.replace('/application/' + contextId).catch(() => {})
+					// reverts turning /apps/tables/app/28 into /apps/tables/app/28#/application/28
```
juliushaertl, line 67: Seems like an OK approach for now.

We could also change the vue-router so that it uses the exact same routes as the backend defines, but that might require a few more changes to have the old urls still work.

https://github.com/nextcloud/deck/commit/9f98f0bd5fe85cd6e42688467c5bca9a75090fc4
https://github.com/nextcloud/deck/pull/1977/files

juliushaertl (approved):
Looks good once CI is happy about linting and missing node dependency 🚀

src/App.vue
blizzz: interesting switch! Thanks for the pointer

enjeck (commented):
@blizzz This is magic :mage: :magic_wand:

blizzz:
Do we want this also in 0.7?

PR #1083, opened 2024-05-10
fix(DB): fetch pageId as int

### LINKED ISSUE #1081: Delete application does not work
### Steps to reproduce

The newly introduced function to delete an application does not work. I get a "Cannot delete application: unknown error" or something like that.

### Expected behavior

An app should be deleted

### Actual behavior

It doesn't get deleted.

### Tables app version

0.7.2

### Browser

Safari

### Client operating system

macOS

### Operating system

RHEL 9

### Web server

Nginx

### PHP engine version

PHP 8.0

### Database

SQLite

### Additional info

_No response_

blizzz:
fixes #1081

### DIFF
--- a/lib/Db/PageMapper.php
+++ b/lib/Db/PageMapper.php
@@ -14,6 +14,9 @@ class PageMapper extends QBMapper {
 		parent::__construct($db, $this->table, Page::class);
 	}
 
+	/**
+	 * @return int[]
+	 */
 	public function getPageIdsForContext(int $contextId): array {
 		$qb = $this->db->getQueryBuilder();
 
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):
I always wonder if the query builder could not cast those automatically, but I guess it cause we don't have type information on the php side for the columns and sqlite considering everything as a string 🙈

blizzz:
> I always wonder if the query builder could not cast those automatically, but I guess it cause we don't have type information on the php side for the columns and sqlite considering everything as a string 🙈

The QBMapper usually would do that, but is not used here. A pitfall to be aware about. Having this automatic with the DB abstraction (based on migration?) would be fancy indeed.

Issue #586, opened 2023-09-18
Column type user and groups


datenangebot:
Find user and groups, select multiple or once's

Show as avatar, user buble or link to user page, talk room etc

Think about data privacy who can see who and think about sharing...

### Related PRs
- Frontend: https://github.com/nextcloud/tables/pull/1113
- Backend + tests: https://github.com/nextcloud/tables/pull/1090
- Frontend tests

RTechSn:
Also, there should be an option to default the current user and an option to lock to the current user only. This would allow to track who created a row or assign an owner to a task type table.

juliushaertl:
Copying additional context from https://github.com/nextcloud/tables/issues/721:

New column type user:
- Can select user, multiple users or (multiple) groups
- Show user id or displayname
- Set if status should be showen (too much, just show it?)
- Only offer users that are allowed in your context (controlled by the general share config eg "share only with own groups" etc.)
- assume a table is shared: if a user is not allowed in the receiver context, just show "1 more users"

https://docs.nextcloud.com/server/latest/developer_manual/client_apis/OCS/ocs-api-overview.html#auto-complete-and-user-search

PR #1090, opened 2024-05-13
feat: add backend for new user/group column

enjeck:
Contributes to #586 

This adds to the backend to support adding a new column that receives users or groups. 

We expect the row values of this column to be array of objects with format `{id: string, type: int}`

#### Request
```bash
curl --silent \
    -X POST \
    -u admin:admin \
    -H "OCS-APIRequest: true" \
    -H "Content-Type: application/json" \
    -d '{ "baseNodeId": 1, "title": "New Column", "usergroupDefault": "[{\"id\": \"admin\", \"type\": 0}]" }' \
'http://nextcloud.local/ocs/v2.php/apps/tables/api/2/columns/usergroup?format=json' | jq
```
#### Result
```json
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 200,
      "message": "OK"
    },
    "data": {
      "id": 7,
      "tableId": 1,
      "title": "New Column",
      "createdBy": "admin",
      "createdByDisplayName": "admin",
      "createdAt": "2024-05-13 15:25:29",
      "lastEditBy": "admin",
      "lastEditByDisplayName": "admin",
      "lastEditAt": "2024-05-13 15:25:29",
      "type": "usergroup",
      "subtype": "",
      "mandatory": false,
      "description": "",
      "numberDefault": null,
      "numberMin": null,
      "numberMax": null,
      "numberDecimals": null,
      "numberPrefix": "",
      "numberSuffix": "",
      "textDefault": null,
      "textAllowedPattern": null,
      "textMaxLength": null,
      "selectionOptions": [],
      "selectionDefault": null,
      "datetimeDefault": null,
      "usergroupDefault": [
        {
          "id": "admin",
          "type": 0,
        }
      ],
      "usergroupMultipleItems": null,
      "usergroupSelectUsers": null,
      "usergroupSelectGroups": null,
      "showUserStatus": null
    }
  }
}
```


tests/integration/features/APIv2.feature
enjeck, line 1: Not sure if APIv1.feature needs to incorporate the new column too

lib/Migration/Version000700Date20230916000000.php
```diff
@@ -38,6 +38,10 @@ class Version000700Date20230916000000 extends SimpleMigrationStep {
 			'name' => 'selection',
 			'db_type' => Types::TEXT,
 		],
+		[
+			'name' => 'usergroup',
+			'db_type' => Types::TEXT,
```
juliushaertl, line 43: I think for building more optimal queries it would actually be nice to have a different table format that not stores the entires in json as text but instead something like this

oc_tables_row_cells_usergroups
| id | column_id | row_id | value | value_type |
|---|---|---|---|---|
| 1 | 2 | 9 | alice | 0 |
| 2 | 2 | 9 | bob | 0 |
| 3 | 2 | 10 | engineering | 1 |

This might be more complex to build but we can do performant filtering on large tables by having an index on the value type and its exact match. For the value type using an integer (0 for user, 1 for group) is common from the server sharing API. Probably something to explore further then. Would be curious if @blizzz may have further thoughts on that.

blizzz: So to have a generic solutions that allows value_type in row_cells_xyz tables:

1. A method like  `lib/Migration/Version000700Date20230916000000::createRowValueTable()` or its optional extension to also add a `value_type` column. The `$columns` array could be extended in that class. For new data types of course we need a repair step.
2. Evaluate an abstract subclass of `RowCellSupper` providing support for setting value type (e.g. adding `setValueTypeWrapper()`) or maybe it is sufficient (though not cleaner) to extend the current class, taking the existence of a value_type from the concrete implementation into account.
3.In the same way evaluate extending `RowCellMapperSuper` with `parseValueTypeOutgoing()` and `parseValueTypeIncoming()` although this is unlikely to contain "weird" data. We shall store there constants only (ideally only ints) and it should not be set by direct user input.
4. `Row2Mapper` needs to become aware of possible `value_type` fields, for example in `insert()`, `getFilterExpression()`  (there we already have type-dependent handling) and generally where at least `formatValue()` is being used
5. Insert ability to merge rows in `oc_tables_row_cells_xyz` that share the same column_id and row_id to be able to support multi-values by row. The difficulty lies in having different cell ids per actually set value, so this makes editing a cell difficult, but on a first glance the frontend does not work with cell IDs when updating a value. `RowCellMapperSuper::findByRowAndColumn()` returns a single entity – a starting point would be to have a generic extension for multiple values with a method lile `findMultipleByRowAndColumn()` that can return multiple entities which will later on be merged for the frontend.
6. Where applicable frontend also has to supply value_type information, e.g. when submitting such cell values.

+ Not looked into it yet by me: frontend live filtering, might not be an issue.

A non-generic solutions also ticks almost all items, albeit on concrete implemenations. The value_type *might* come in handy with other future types (could imagine it for columns like file or image picker, ip address, icon/emoji picker).

On a second  look, i just saw the `TextLinkBusiness` thing and it stores already up to three data points in its value (`title`, `value`, `providerId`). If we want to be able to serve scenarios like these, we'd need to look into a more flexible approach.

blizzz: There would be no operation against a single row, that's true. I think the update is already an update-or-insert, so new values should not be problematic, but those that may be delete do require additional checks and delete queries, yes.

enjeck: @blizzz @juliushaertl I updated the table format

lib/Service/ColumnTypes/UsergroupBusiness.php
```diff
@@ -0,0 +1,50 @@
...
+			$this->logger->warning('No column given, but expected on '.__FUNCTION__.' within '.__CLASS__, ['exception' => new \Exception()]);
+			return false;
+		}
+
+		if($value === null) {
+			return true;
+		}
+
+		foreach ($value as $v) {
+			// TODO: maybe check if key exists first
```
blizzz, line 43: i think this todo is done?

lib/Db/ColumnTypes/UsergroupColumnQB.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace OCA\Tables\Db\ColumnTypes;
+
+use OCP\DB\QueryBuilder\IQueryBuilder;
+
+class UsergroupColumnQB extends SuperColumnQB implements IColumnTypeQB {
+	public function passSearchValue(IQueryBuilder $qb, string $unformattedSearchValue, string $operator, string $searchValuePlaceHolder): void {
+		// TODO how to handle searching for multiple users/groups?
```
blizzz, line 9: Do you mean, if multiple users/groups are part of the search string?

blizzz (commented):
First glance looks good, have not tested myself yet

- [ ] Integration test can be extended around inserting, modifying and delete column values
- [ ] a new migration class is needed to add the new columns on existing installations (existing steps are not executed again).

lib/Db/Column.php
```diff
@@ -121,6 +139,26 @@ public function __construct() {
...
+		// // type usergroup
+		$this->addType('usergroupMultipleItems', 'boolean');
+		$this->addType('usergroupSelectUsers', 'boolean');
+		$this->addType('usergroupSelectGroups', 'boolean');
+		$this->addType('showUserStatus', 'boolean');
+	}
+
+	public function getUsergroupDefaultArray():array {
+		$default = $this->getUsergroupDefault();
+		if ($default !== "" && $default !== null && $default !== 'null') {
```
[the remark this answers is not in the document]
enjeck: No :sweat_smile: . I have no idea what I was doing

lib/Db/Row2Mapper.php
```diff
@@ -506,13 +513,32 @@ private function parseEntities(IResult $result, array $sleeves): array {
...
+			$compositeKey = (string)$rowData['row_id'] . ',' . (string)$rowData['column_id'];
+
+			if ($columnTypes[$rowData['column_id']] == 'usergroup') {
+				if (array_key_exists($compositeKey, $rowValues)) {
+					$rowValues[$compositeKey][] = $value;
+				} else {
+					$rowValues[$compositeKey] = [$value];
+				}
+			} else {
+				$rowValues[$compositeKey] = $value;
```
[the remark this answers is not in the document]
enjeck: It would be nice to have one common format. But other column types aren't arrays and I think it would take many changes to make them as such.

lib/Db/ColumnTypes/UsergroupColumnQB.php
enjeck: Yeah. But I guess we can consider this later on

enjeck:
> * Integration test can be extended around inserting, modifying and delete column values

Not sure what this entails. Do we already have this for other column types? @blizzz

lib/Migration/Version001000Date20240712000000.php
```diff
@@ -0,0 +1,62 @@
...
+				]);
+			}
+			if (!$table->hasColumn('show_user_status')) {
+				$table->addColumn('show_user_status', Types::BOOLEAN, [
+					'notnull' => false,
+					'default' => 0,
+				]);
+			}
+			return $schema;
+		}
```
blizzz, line 60:
```suggestion
		}
		return null;
```

blizzz:
> > * Integration test can be extended around inserting, modifying and delete column values
> 
> Not sure what this entails. Do we already have this for other column types? @blizzz

I think you already covert inserting with a user. 

Then there are some tests in Apiv1 on modification, e.g. https://github.com/nextcloud/tables/blob/6793858e77ab97c143f02079d849c46a48348bc7/tests/integration/features/APIv1.feature#L117

blizzz (commented):
Nitpick, otherwise the recent changes look good, have to test yet.

enjeck:
> > > * Integration test can be extended around inserting, modifying and delete column values
> > 
> > 
> > Not sure what this entails. Do we already have this for other column types? @blizzz
> 
> I think you already covert inserting with a user.
> 
> Then there are some tests in Apiv1 on modification, e.g.
> 
> https://github.com/nextcloud/tables/blob/6793858e77ab97c143f02079d849c46a48348bc7/tests/integration/features/APIv1.feature#L117

@blizzz Maybe I don't understand how it all works, but I don't think we can easily add these tests into APIv2 like we have in APIv1 because there are no v2 routes for modifying columns/rows set up in `routes.php`???

lib/Controller/Api1Controller.php
```diff
@@ -759,6 +764,13 @@ public function createColumn(
 		?string $selectionDefault = '',
 
 		?string $datetimeDefault = '',
+
+		?string $usergroupDefault = '',
+		?bool $usergroupMultipleItems,
+		?bool $usergroupSelectUsers,
+		?bool $usergroupSelectGroups,
+		?bool $showUserStatus,
```
juliushaertl, line 772:
```suggestion
		?bool $usergroupMultipleItems = null,
		?bool $usergroupSelectUsers = null,
		?bool $usergroupSelectGroups = null,
		?bool $showUserStatus = null,
```

Need a default value as otherwise we have an error thrown:

> Optional parameter $usergroupDefault declared before required parameter $showUserStatus is implicitly treated as a required parameter

lib/Controller/Api1Controller.php
```diff
@@ -1440,6 +1482,13 @@ public function createTableColumn(
...
```
juliushaertl, line 1490:
```suggestion
		?bool $usergroupMultipleItems = null,
		?bool $usergroupSelectUsers = null,
		?bool $usergroupSelectGroups = null,
		?bool $showUserStatus = null,
```

lib/Migration/Version001000Date20240712000000.php
```diff
@@ -0,0 +1,63 @@
...
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 * @throws Exception
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('tables_tables')) {
+			$table = $schema->getTable('tables_tables');
```
juliushaertl, line 29:
```suggestion
		if ($schema->hasTable('tables_columns')) {
			$table = $schema->getTable('tables_columns');
```

Seems like the wrong table

lib/Migration/Version000700Date20230916000000.php
```diff
@@ -38,6 +38,10 @@ class Version000700Date20230916000000 extends SimpleMigrationStep {
 			'name' => 'selection',
 			'db_type' => Types::TEXT,
 		],
+		[
```
juliushaertl, line 41: This needs to happen in a separate migration as well, otherwise the usergroup table is not created

It is probably the best to copy the logic of this migration to a new one to create it. if it does not exist.

lib/Db/RowCellUsergroupMapper.php
```diff
@@ -0,0 +1,29 @@
...
+	public function __construct(IDBConnection $db) {
+		parent::__construct($db, $this->table, RowCellUsergroup::class);
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function parseValueIncoming(Column $column, $value): array {
+		return json_decode(json_encode($value), true);
+		//return json_encode($value);
```
juliushaertl, line 20:
```suggestion
```

lib/Migration/Version000700Date20230916000000.php
```diff
@@ -72,6 +76,7 @@ private function createRowValueTable(ISchemaWrapper $schema, string $name, strin
 			$table->addColumn('column_id', Types::INTEGER, ['notnull' => true]);
 			$table->addColumn('row_id', Types::INTEGER, ['notnull' => true]);
 			$table->addColumn('value', $type, ['notnull' => false]);
+			$table->addColumn('value_type', Types::INTEGER, ['notnull' => false]);
```
juliushaertl, line 79: Same for this, should be done in a new migration as well.

juliushaertl (changes requested):
A few comments with some odd things I've seen on the migrations and when trying to run it locally

blizzz:
> > > > * Integration test can be extended around inserting, modifying and delete column values
> > > 
> > > 
> > > Not sure what this entails. Do we already have this for other column types? @blizzz
> > 
> > 
> > I think you already covert inserting with a user.
> > Then there are some tests in Apiv1 on modification, e.g.
> > https://github.com/nextcloud/tables/blob/6793858e77ab97c143f02079d849c46a48348bc7/tests/integration/features/APIv1.feature#L117
> 
> @blizzz Maybe I don't understand how it all works, but I don't think we can easily add these tests into APIv2 like we have in APIv1 because there are no v2 routes for modifying columns/rows set up in `routes.php`???

You can call Api1 endpoints also from the Api2 tests. Just logically it makes send to have the test subject in APIv2, but auxiliary calls can go whenever. They share the same feature context.

lib/Migration/Version001000Date20240712000000.php
```diff
@@ -0,0 +1,82 @@
...
+namespace OCA\Tables\Migration;
+
+use Closure;
+use OCP\DB\Exception;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version001000Date20240712000000 extends SimpleMigrationStep {
```
juliushaertl, line 16:
```suggestion
class Version000800Date20240712000000 extends SimpleMigrationStep {
```

and also name the filename like that The version there is following the app version where we currently target 0.8

blizzz:
@enjeck I pushed two fixup commits for your to review. The failing tests are green now for me, locally. 

There is something I found inconsistent: the default values for the rows is formatted as a json string in the reply to the API call, while the row values are returned as array. I did work around that in the `FeatureContext.php`. I did not double check how the options column type is doing it, but perhaps it is also just lacking tests. But a consistent output format is appreciated.

juliushaertl:
All right, from the journey through the rabbit hole of value transformation/parsing, i came up with a few changes:

Related to this PR:
- The value_type column is now only added to the usergroup cell data table
- I cleaned up the migrations a bit further
- I added the to filter by one value in views for the user group column

More general as the current approach to handle different value formats was really confusing and not feasible for the usergroup column:
- Improve mapper methods to be more clear on what they do
  - `parseValueIncoming`: This had two use cases, one to map from the request data to the value to insert cell data into the database, but also to map from filter values to the value that is used in the query builder to construct view filter queries. It was split into
    - `filterValueToQueryParam`: For a stored value of a filter rule get the value to put in the database query, this could in the future be refactored again to actually build queries more dynamically, but left that out to keep the changes at a reasonable level
    - For the case of inserting/updating values we no longer pass encode/decode json values, but an `applyDataToEntity` method is added for each case to just fill the Entity object depending on the type.
  - `parseValueOutgoing`
    - Changed to a `formatEntity` and no longer use a mixed value type but the actual database entity so we know what to expect and how to transform that to the actual response data format
  - Use `hasMultipleValues` to indicate that multiple values are stored in multiple entities compared to a hardcoded column type
- Removed a few duplicate code bits around getting the mapper class dynamically

Still checking in on the test failures

lib/Db/ColumnTypes/UsergroupColumnQB.php
juliushaertl: Yes, I think this also needs more advanced capabilities of building view filters to not only be able to select a string value but also be able to differentiate between the different value types. Something to file an issue for later

juliushaertl (approved):
👍  But would appreciate @blizzz and @enjeck to also give their thumbs up for my parts

lib/Controller/Api1Controller.php
```diff
@@ -761,6 +761,11 @@ public function indexViewColumns(int $viewId): DataResponse {
 	 * @param string|null $selectionOptions Options for a selection (json array{id: int, label: string})
 	 * @param string|null $selectionDefault Default option IDs for a selection (json int[])
 	 * @param string|null $datetimeDefault Default value, if column is datetime
+	 * @param string|null $usergroupDefault Default value, if column is usergroup (json array{id: string, type: int})
+	 * @param bool|null $usergroupMultipleItems Can select multiple users or/and groups, if column is usergroup
+	 * @param bool|null $usergroupSelectUsers Can select users, if column type is usergroup
+	 * @param bool|null $usergroupSelectGroups Can select groups, if column type is usergroup
+	 * @param bool|null $showUserStatus Whether to show the user's status, if column type is usergroup
```
blizzz, line 768: this is not prefixed with `usergroup`, thus inconsistent with the prefix flags. Or is this intended?

lib/Db/Row2Mapper.php
```diff
@@ -194,8 +188,19 @@ private function getRows(array $rowIds, array $columnIds): array {
 		foreach ($this->columnsHelper->columns as $columnType) {
 			$qbTmp = $this->db->getQueryBuilder();
 			$qbTmp->select('row_id', 'column_id', 'last_edit_at', 'last_edit_by')
-				->selectAlias($qb->expr()->castColumn('value', IQueryBuilder::PARAM_STR), 'value')
-				->from('tables_row_cells_'.$columnType)
+				->selectAlias($qb->expr()->castColumn('value', IQueryBuilder::PARAM_STR), 'value');
+
+			// This is not ideal but I cannot think of a good way to abstract this away into the mapper right now
```
blizzz, line 193: roger. perhaps opening an issue for this afterwards?

lib/Migration/Version000800Date20240712000000.php
```diff
@@ -0,0 +1,91 @@
...
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version000800Date20240712000000 extends SimpleMigrationStep {
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$hasChanges = $this->createUserGroupTable($schema, 'usergroup', Types::TEXT);
```
blizzz, line 23: nitpick: `$hasChanges` sounds like a bool. maybe just `$changes`?

blizzz (commented):
A question, a hint, a follow up thing, and a nitpick, nothing blocking per se.

PR #1092, opened 2024-05-16
fix(Context): do not show hidden columns

blizzz:
There are two endpoints for loading the columns of a view. When using that one against a table scope, all columns will be returned, while fetching only those against the direct view endpoint, only those are returned that actually should be displayed. When opening the classic view, the later is called. In Contexts/Applications, the first one was called resulting in empty columns. This changes fixes it.

In the screenshot you see that "Last time spotted" is not shown after the change, as it is supposed to given the View configuration (hence no values are shown before either).

| Before  | After |
| --- | --- |
| before |  after |


elzody (approved):
Seems reasonable

PR #1101, opened 2024-05-23
feat: Add events for row added and row updated

come-nc:


come-nc:
> Ah, I guess related to Workflows?

Yes!

lib/Event/RowAddedEvent.php
```diff
@@ -0,0 +1,18 @@
...
+use OCA\Tables\Db\Row2;
+use OCP\EventDispatcher\Event;
+
+final class RowAddedEvent extends Event {
+	public function __construct(protected Row2 $row) {
+		parent::__construct();
+	}
+
+	public function getRow(): Row2 {
+		return $this->row;
```
juliushaertl, line 16: @blizzz What's your thought on just returning the entity directly here?

I'm unsure if we shall expose the internal row entity through events, rather expose specific getters or use a serializable event as https://github.com/nextcloud/tables/issues/1103 mentioned

juliushaertl:
@come-nc To evaluate what we want to pass along in the event, could you share more insights on the use case you're covering with your workflow task?

come-nc:
> @come-nc To evaluate what we want to pass along in the event, could you share more insights on the use case you're covering with your workflow task?

Usecases are multiple:

- Have a workflow react to data inserted in a table (for instance a user is created in another application when a line is added to a table).
- Have a workflow react to a modification of a row (typically a boolean column is checked to approve a vacation request or something, triggers a chain of events).

I feel like putting the whole row content information in the event is too much because it may be a lot of data, while putting only the id allows to fetch the data from the API if needed. Not entirely sure. Maybe content but capped at a given maximum length would make sense?

The event will need to be serialized by implementing upcomming `OCP\EventDispatcher\IWebhookCompatibleEvent` interface.

lib/Event/RowAddedEvent.php
blizzz: I also have reservations about exposing the Entity. Turning this into a minified (and so serializable) model instance might could be a way. The Entity is can actually be json-serialized, that could be a base.

Then, in principle, whatever we are doing, we are de-facto exposing a Model to the outside, so setting standards as in an API, and have to keep them stable. That's generally fine, but would be a firstee at least in this app.

blizzz: Just read the note about `OCP\EventDispatcher\IWebhookCompatibleEvent` which is already in master meanwhile. But this is yet another `jsonSerialize()` style method. In a way i still tend to a public model class that itself is just json serializable and we return the output here. On the other hand, this can be already done with the DB Entity.

I also noticed that we are actually exposing the DB Entities already in other events that are also released already. Including Row2.

blizzz:
Fixed conflicts, and added two changes (cf. my comments https://github.com/nextcloud/tables/pull/1101#discussion_r1679024911 and https://github.com/nextcloud/tables/pull/1101#discussion_r1679256807).

I added now a "Public" read-only Row model that is being attached to the Event. 

I am lacking a concept on how the IWebhookCompatibleEvent is supposed to be used (left a 1:1 message with @marcelklehr), but it should contain the necessary data. The acting user is left out (previously discussed with Côme). The Column(s) can be derived from the change sets. Not the Cell, it's probably not necessary, and we actually don't have that information ourselves. Especially since changed data is already debated.

At the moment I send the row's values, and on update also the previous values. With Côme's statement:

> I feel like putting the whole row content information in the event is too much because it may be a lot of data, while putting only the id allows to fetch the data from the API if needed. Not entirely sure. Maybe content but capped at a given maximum length would make sense?

It can be dropped indeed, return only the values that actually have changed (i.e. nothing on create and delete). Still can be a lot. Only the affected columns ids can be sent, too.

What I am having in mind is if process changes shall be followed. E.g. Something may change from "pending" to "open", and from "closed" to "open" and different things shall happen then. Again, I am missing context down the line, whether this is something reasonable, or not relevant here.

lib/Event/AbstractRowEvent.php
```diff
@@ -0,0 +1,43 @@
...
+			$this->row = $rowEntity->toPublicRow($previousValues);
+		}
+
+		public function getRow(): Row {
+			return $this->row;
+		}
+
+		public function getWebhookSerializable(): array {
+			return $this->row->jsonSerialize();
+		}
```
marcelklehr, line 27: Awesomesauce!

marcelklehr (approved):

juliushaertl (approved):

Issue #1064, opened 2024-05-06
Submit row details with ctrl+enter


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a Tables user I want to quickly save my changes when editing a row. Currently I have to tab to the *Save* button or use the mouse.

### Describe the solution you'd like

Convert the row editor modal to a native HTML form I can submit using ctrl+enter when an input is focused.

juliushaertl:
We should also make sure this works on all modals (table/view/application/row create/edit)

PR #1112, opened 2024-05-31
feat: submit row details with ctrl + enter

luka-nextcloud:
* Resolves: #1064


elzody:
Seems like both additions are the same code, maybe it would be an overkill but I would be in favor of moving it to something like a mixin or something and having them both share it. Maybe it's fine the way it is though, it's only a suggestion. Otherwise looks good. What are your thoughts?

luka-nextcloud:
@elzody I think they are too small to move to a mixin. We can consider about it later when those functions expand more.

elzody (approved):

PR #1119, opened 2024-06-04
test(cypress): fix month names

blizzz:
Should fix the failing cypress tests ending in

```
1) Test column date
       Insert and test rows - default now:
     AssertionError: Timed out retrying after 10000ms: Expected to find content: 'June' within the element: <div> but never did.
      at Context.eval (webpack://tables/./cypress/e2e/column-datetimeDate.cy.js:59:86)
```


enjeck (approved):
cool

PR #1141, opened 2024-06-13
use a built-in JS function `localeCompare()` to compare strings

kirisakow:
This PR is a proposal for using a built-in JavaScript function [`localeCompare()`][1] for strings sorting. If you read the documentation, you'll see it does exactly what was previously implemented by hand.

Pros:
- it is a built-in function of the JavaScript API;

- `s1.localeCompare(s2, 'your-locale')` looks cleaner and more legible than nested ternary operators that used to be there;

- it is more straightforward at handling numeric strings: no need to `parseInt(s)` or `parseFloat(s)` before comparison, just `s1.localeCompare(s2, 'your-locale', {numeric: true})`;

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

### DIFF
--- a/src/shared/components/ncTable/mixins/columnsTypes/number.js
+++ b/src/shared/components/ncTable/mixins/columnsTypes/number.js
@@ -33,7 +33,7 @@ export default class NumberColumn extends AbstractNumberColumn {
 			if (!valueA && !valueB) {
 				return super.getNextSortsResult(nextSorts, rowA, rowB)
 			}
-			return valueA.localeCompare(valueB, undefined, { numeric: true }) * factor || super.getNextSortsResult(nextSorts, rowA, rowB)
+			return ((valueA < valueB) ? -1 : (valueA > valueB) ? 1 : 0) * factor || super.getNextSortsResult(nextSorts, rowA, rowB)
 		}
 	}
 
--- a/src/shared/components/ncTable/mixins/columnsTypes/number.js
+++ b/src/shared/components/ncTable/mixins/columnsTypes/number.js
@@ -33,7 +33,7 @@ export default class NumberColumn extends AbstractNumberColumn {
 			if (!valueA && !valueB) {
 				return super.getNextSortsResult(nextSorts, rowA, rowB)
 			}
-			return ((valueA < valueB) ? -1 : (valueA > valueB) ? 1 : 0) * factor || super.getNextSortsResult(nextSorts, rowA, rowB)
+			return valueA.localeCompare(valueB, undefined, { numeric: true }) * factor || super.getNextSortsResult(nextSorts, rowA, rowB)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


elzody:
Hey there, this looks pretty good, other than for me it isn't working for columns which are not strings, so I think it would need adjusted for other column types is my guess.

<details>
<summary>See screenshot here</summary>

image
</details>

kirisakow:
@elzody Thank you. 

### Changes undone for files

- src/shared/components/ncTable/mixins/columnsTypes/number.js
- src/shared/components/ncTable/mixins/columnsTypes/numberProgress.js
- src/shared/components/ncTable/mixins/columnsTypes/numberStars.js
- src/shared/components/ncTable/mixins/columnsTypes/selection.js

### Side question: What's your NC setup for development and testing?

Lately I've been having annoying misbehavior in my local NC developer instance I had spent a few hours to setup... So I've started thinking I should scrap it and go the docker-compose way. I'd appreciate if you guys @elzody @enjeck could suggest me an all-ready docker-compose.yml file as well as any steps to set it up... if you have such knowledge. (Sure, I've seen there are lots of images in [nextcloud/docker-ci][1] repository, but I can't wrap my head around it all on my own...)

[1]: https://github.com/nextcloud/docker-ci?tab=readme-ov-file#ci-images-for-nextcloud

elzody:
@kirisakow My setup, believe it or not, is also a bit messy. I need to take the time to improve the workflow a bit, but generally I use the standalone container and just bind mount the directories of the apps I am testing/developing for into the container. You can find more info about the standalone containers [here](https://juliushaertl.github.io/nextcloud-docker-dev/basics/standalone/), and a sample of my Docker compose file is like so:

```
services:
  nextcloud:
    image: ghcr.io/juliushaertl/nextcloud-dev-php82:latest
    container_name: nextcloud
    restart: unless-stopped
    ports:
      - "7128:80"
    environment:
        - SERVER_BRANCH=master
    volumes:
        # Here I just bind mount the apps, including server
        # If you don't do this, it's fine, it will auto clone it, etc. based on SERVER_BRANCH above
        - type: bind
          source: server        # Path to where I have server repo cloned
          target: /var/www/html

        - type: bind
          source: tables        # Path to where I have tables repo cloned
          target: /var/www/html/apps-extra/tables
```

Isn't the best, but works for me. The only part of the compose file I left out was that related to the Collabora container, not sure if you are gonna be using that at all.

elzody (commented):
Gave it a test, and it works nicely. Looks much more readable than the weird ternary operators before.

I am not quite sure what is up with the DCO check failing, but I *suspect* it has something to do with the `documentation` thing included in your first commit. Try doing an interactive rebase and see if you can alter that message to remove it, and we will see if that fixes it

kirisakow:
> I am not quite sure what is up with the DCO check failing, but I _suspect_ it has something to do with the `documentation` thing included in your first commit. Try doing an interactive rebase and see if you can alter that message to remove it, and we will see if that fixes it

You were right. My original commit message was split in 3 lines (with the URL on the 2nd line), but DCO seemingly wants no more than one newline break, with the sign-off being on the 2nd line. 🤷‍♂

elzody (approved):

PR #1149, opened 2024-06-17
enh: add context e2e tests

### LINKED ISSUE #994: Applications: Frontend tests

enjeck:
Fixes #994 

- [x] Create a new application
- [x] Add a table to it
- [x] Add a view to it
- [x] Check that both are rendered
- [x] Share with a user
- [x] Check that the user receives the context
- [x] Check that the application menu entry shows up
- [x] Deleting an application (for both owner and shares)
- [x] Transfering an application
- [x] Modifying permissions and creating, editing, deleting rows/columns


blizzz (approved):
just one test

PR #1156, opened 2024-06-19
fix: Remove deleted resource from associated applications

### LINKED ISSUE #1027: Remove deleted resource from associated Applications
### Steps to reproduce

1.Create table/view
2. Create Application with view/table
3. Delete table/view
4. Visit Application

### Expected behavior

The deleted table/view should be removed from the Application

### Actual behavior

Application stuck on loading due to Axios error "Could not fetch table"

### Tables app version

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

{
    "message": "Request failed with status code 404",
    "name": "AxiosError",
    "stack": "AxiosError: Request failed with status code 404\n    at settle (http://nextcloud.local/apps/tables/js/tables-main.js:245596:12)\n    at XMLHttpRequest.onloadend (http://nextcloud.local/apps/tables/js/tables-main.js:244087:66)",
    "config": {
        "transitional": {
            "silentJSONParsing": true,
            "forcedJSONParsing": true,
            "clarifyTimeoutError": false
        },
        "adapter": [
            "xhr",
            "http"
        ],
        "transformRequest": [
            null
        ],
        "transformResponse": [
            null
        ],
        "timeout": 0,
        "xsrfCookieName": "XSRF-TOKEN",
        "xsrfHeaderName": "X-XSRF-TOKEN",
        "maxContentLength": -1,
        "maxBodyLength": -1,
        "env": {},
        "headers": {
            "Accept": "application/json, text/plain, */*",
            "requesttoken": "4V9idrIBqNsy2XOn3yTi4fHRo46uEs/EmREV+buEc4Y=:ygUQT/t43JoF6DielmeU1baT9tuBIv2Kw1Qmo/XRPMM="
        },
        "method": "get",
        "url": "http://nextcloud.local/ocs/v2.php/apps/tables/api/2/tables/7"
    },
    "code": "ERR_BAD_REQUEST",
    "status": 404
}

enjeck:
Fixes #1027


blizzz:
:thinking: I thought we/I did it in the past :confused:

lib/Db/ContextNodeRelationMapper.php
```diff
@@ -28,6 +28,26 @@ public function deleteAllByContextId(int $contextId): void {
 		$qb->executeStatement();
 	}
 
+	public function getRelIdsForNode(int $nodeId, int $nodeType): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('id')->from($this->table)->where($qb->expr()->eq('node_id', $qb->createNamedParameter($nodeId)))
+		->andWhere($qb->expr()->eq('node_type', $qb->createNamedParameter($nodeType)));
```
juliushaertl, line 34: Just for easier reading:

```suggestion
		$qb->select('id')->from($this->table)
			->where($qb->expr()->eq('node_id', $qb->createNamedParameter($nodeId)))
			->andWhere($qb->expr()->eq('node_type', $qb->createNamedParameter($nodeType)));
```

lib/Service/ContextService.php
```diff
@@ -252,6 +252,19 @@ public function delete(int $contextId, string $userId): Context {
 		return $context;
 	}
 
+	public function deleteNodeRel(int $nodeId, int $nodeType): void {
+		try {
+			$nodeRelIds = $this->contextNodeRelMapper->getRelIdsForNode($nodeId, $nodeType);
+			foreach ($nodeRelIds as $nodeRelId) {
+				$this->pageContentMapper->deleteByNodeRelId($nodeRelId);
+				$this->contextNodeRelMapper->deleteByNodeRelId($nodeRelId);
+			}
```
juliushaertl, line 261: We can save the n+1 queries here by rather having methods that support multiple ids in an array and use an `IN()` sql statement. Might not be crucial here but generally a good pattern to follow for performance instead of doing queries in a loop

lib/Service/ContextService.php
```diff
@@ -252,6 +252,19 @@ public function delete(int $contextId, string $userId): Context {
...
+			
+		} catch (Exception $e) {
+			$this->logger->error('something went wrong while deleting node relation for node id: '.(string)$nodeId. ' and node type '.(string)$nodeType);
```
juliushaertl, line 264:
```suggestion
			$this->logger->error('Something went wrong while deleting node relation for node id: ' . (string)$nodeId . ' and node type ' . (string)$nodeType, ['exception' => $e]);
```

Usually it is good to also pass the exception in the log so that the reason and trace can be found. Also I would expect that method to throw the error instead of silencing with just a log message.

juliushaertl (changes requested):

lib/Service/ContextService.php
enjeck: Not sure how to go about this :confused: :confused:

juliushaertl: Let's check that together next week then :)

juliushaertl (approved):

blizzz:
> > 🤔 I thought we/I did it in the past 😕
> 
> @blizzz Anything you'd want to still check in that regard?

I would fancy a database clean up (could be a run-once expensive/maintenance background job), but this can be a follow-up PR.  Which can then also be dropped in a future version, still good practice to keep the DB neat and tidy.

src/pages/Context.vue
```diff
@@ -9,7 +9,8 @@
 			<div class="content context">
 				<div class="row first-row">
 					<h1 class="context__title">
-						<NcIconSvgWrapper :svg="icon" :size="32" style="display: inline-block;" />&nbsp; {{ activeContext.name }}
+						<NcIconSvgWrapper :svg="icon" :size="32" style="display: inline-block;" />&nbsp; {{
+							activeContext.name }}
```
blizzz, line 13: is the line break intentional?

blizzz (approved):

Issue #1105, opened 2024-05-24
Provide OCS endpoints for creating rows in Tables or Views


blizzz:
### Is your feature request related to a problem? Please describe.

Turn the legacy `api1#createRowInView` and `api1#createRowInTable` endpoints into OCS endpoints, so that they are easier usable for third parties.

### Describe the solution you'd like

Evaluate whether it is possible to combine them into one endpoint. 

### Additional context

This is required for the upcoming workflow engine efforts.

juliushaertl:
@blizzz Could you line this up for the next two weeks? I would expect implementation being rather just reusing existing services and writing tests.

PR #1161, opened 2024-06-21
enh(API): add OCS API to create rows

blizzz:
solves #1105 

also features
:sparkles: completed and flexible PermissionMiddleware
:sparkles: integration tests of new endpoint against both tables and views, owners and sharees, negative tests
:sparkles: the advertised new OCS endpoint for creating rows
:sparkles: helper model to avoid a loose array with set or not set keys flying around
:sparkles: simple conversion helper to translate stringy node types to ints and other way around

Open question: should I create endpoints for all other row-related methods as well? Would make sense, this feels incomplete. Can do a follow up as well of course.

### DIFF
--- a/lib/Middleware/PermissionMiddleware.php
+++ b/lib/Middleware/PermissionMiddleware.php
@@ -2,10 +2,12 @@
 
 namespace OCA\Tables\Middleware;
 
+use InvalidArgumentException;
 use OCA\Tables\AppInfo\Application;
 use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Errors\PermissionError;
+use OCA\Tables\Helper\ConversionHelper;
 use OCA\Tables\Middleware\Attribute\RequirePermission;
 use OCA\Tables\Service\PermissionsService;
 use OCP\AppFramework\Http;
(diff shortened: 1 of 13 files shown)


blizzz:
One set of failing tests were about gracefullness – fixed.

The other, because in 26 there is no `NoAdminRequired` attribute, only the annotation. Fixed the compatibility thing by adding it. We may consider nevertheless to drop 26 from 0.8.

blizzz:
btw,

> Open question: should I create endpoints for all other row-related methods as well? Would make sense, this feels incomplete. Can do a follow up as well of course.

would only do as follow up, if only to not make this PR bigger

lib/Middleware/PermissionMiddleware.php
```diff
@@ -98,21 +101,78 @@ protected function checkPermission(RequirePermission $attribute): void {
...
+				));
+			}
+		} else {
+			if (!$this->permissionsService->canManageContextById($nodeId, $this->userId)) {
+				throw new PermissionError(sprintf('User %s cannot manage context %d',
+					$this->userId, $nodeId
+				));
 			}
 		}
+		return true;
```
juliushaertl, line 175: Is there a use case for returning true in the assert-Methods if we throw anways in case of no permission?

juliushaertl (approved):
Cool, just a non-blocking question 👍

lib/Middleware/PermissionMiddleware.php
blizzz: Otherwise I could not chain them in the match block: https://github.com/nextcloud/tables/pull/1161/files#diff-58a7c8ba568eef76405b8471729c4b735def0967160958d5ab0105ddb7afb34eR109-R112

could turn it into switch construction instead of course. I find the match more compact and readable.

juliushaertl: Ah ok, that makes sense then 👍

PR #1169, opened 2024-06-26
fix(API): declared array shape was not correct

### LINKED ISSUE #1157: OpenAPI annotation for createRowInTable is likely wrong
### Steps to reproduce

See https://github.com/nextcloud/tables/blob/main/lib/Controller/Api1Controller.php#L1100

### Expected behavior

The code seems to expect

```php
/** @var array<int,mixed> */
```

### Actual behavior

Actually declared is

```php
/** @var array{columnId: int, value: mixed} */
```

### Tables app version

main

### Browser

_No response_

### Client operating system

_No response_

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

_No response_

blizzz:
fixes #1157

### DIFF
--- a/lib/Controller/Api1Controller.php
+++ b/lib/Controller/Api1Controller.php
@@ -1097,7 +1097,7 @@ class Api1Controller extends ApiController {
 	 * @NoCSRFRequired
 	 *
 	 * @param int $tableId Table ID
-	 * @param string|array{columnId: int, value: mixed} $data Data as key - value store
+	 * @param string|array{int, mixed} $data Data as key - value store
 	 * @return DataResponse<Http::STATUS_OK, TablesRow, array{}>|DataResponse<Http::STATUS_FORBIDDEN|Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
 	 *
 	 * 200: Row returned
--- a/openapi.json
+++ b/openapi.json
@@ -3852,15 +3852,11 @@
                                             {
                                                 "type": "object",
                                                 "required": [
-                                                    "columnId",
-                                                    "value"
+                                                    null,
+                                                    null
                                                 ],
                                                 "properties": {
-                                                    "columnId": {
-                                                        "type": "integer",
-                                                        "format": "int64"
-                                                    },
-                                                    "value": {
+                                                    "": {
                                                         "type": "object"
                                                     }
                                                 }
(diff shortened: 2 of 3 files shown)


juliushaertl (approved):

provokateurin:
This fix is not correct, it now requires a list with and two elements of int and mixed https://psalm.dev/r/74104d149c. You can also see that the generated spec is totally wrong and invalid. It should have been `array<int, mixed>` to achieve what you want.

blizzz:
> This fix is not correct, it now requires a list with and two elements of int and mixed https://psalm.dev/r/74104d149c. You can also see that the generated spec is totally wrong and invalid. It should have been `array<int, mixed>` to achieve what you want.

Thank you for your direct reply. You are right of course.

The odd thing is, that OpenAPI will not build with the correct pairs of brackets:

```
PHP Fatal error:  Uncaught Error: api1#createRowInTable: Unable to resolve OpenAPI type:
\PHPStan\PhpDocParser\Ast\Type\GenericTypeNode::__set_state(array(
   'type' => 
  \PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode::__set_state(array(
     'name' => 'array',
     'attributes' => 
    array (
    ),
  )),
   'genericTypes' => 
  array (
    0 => 
    \PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode::__set_state(array(
       'name' => 'int',
       'attributes' => 
      array (
      ),
    )),
    1 => 
    \PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode::__set_state(array(
       'name' => 'mixed',
       'attributes' => 
      array (
      ),
    )),
  ),
   'variances' => 
  array (
    0 => 'invariant',
    1 => 'invariant',
  ),
   'attributes' => 
  array (
  ),
))
```

a plain `array<mixed>` will do, apart of a general error (*✘  Globbing has been deprecated in favor of redocly.yaml’s `apis` keys.*) that also appears without a change. The IDE complains, however, and suggests to replace with `array`, which will openapi enrage, because it might be empty :roll_eyes:

provokateurin:
`array<mixed>` shouldn't be used either. I will look into the error and then hopefully provide a fix here as well.

blizzz:
> `array<mixed>` shouldn't be used either. I will look into the error and then hopefully provide a fix here as well.

No, that was a test just for scientific reasons.

provokateurin:
It doesn't work because this is impossible in JSON. There is no way to index an object by non-strings, so it has to be a list or the keys need to be changed to strings. Since you cast to int later anyway, changing to string seems fine?
I'll add a better warning to openapi-extractor for these cases.

PR #1170, opened 2024-06-26
add scheme import and export

### LINKED ISSUE #1122: Template/schema export/import
In order to properly model and share procedures (especially when integrating with workflow engines), we'd need to have a proper export and import to table/views/applications schema.

Example use cases would be to model a structure on a test system and then be able to "deploy" it to a production one or in a later step have way to distribute "templates" of tables.

- Export of a table will dump the table schema
  - Should we just include all views as well?
- Not sure if exporting an individual view is relevant
- Exporting an application likely should include the schema of all linked tables/views
- Import can be generic

grnd-alt:
resolves #1122

### DIFF
--- a/lib/Controller/ApiTablesController.php
+++ b/lib/Controller/ApiTablesController.php
@@ -3,8 +3,6 @@
 namespace OCA\Tables\Controller;
 
 use Exception;
-use OCA\Tables\Db\Column;
-use OCA\Tables\Db\View;
 use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Errors\PermissionError;
--- a/lib/Controller/ApiTablesController.php
+++ b/lib/Controller/ApiTablesController.php
@@ -7,27 +7,32 @@ use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Errors\PermissionError;
 use OCA\Tables\ResponseDefinitions;
+use OCA\Tables\Service\ColumnService;
 use OCA\Tables\Service\TableService;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\IL10N;
 use OCP\IRequest;
 use Psr\Log\LoggerInterface;
+use OCA\Tables\Controller\TableScheme;
 
 /**
  * @psalm-import-type TablesTable from ResponseDefinitions
  */
 class ApiTablesController extends AOCSController {
 	private TableService $service;
+	private ColumnService $columnService;
 
 	public function __construct(
 		IRequest $request,
 		LoggerInterface $logger,
 		TableService $service,
+		ColumnService $columnService,
 		IL10N $n,
 		string $userId) {
 		parent::__construct($request, $logger, $n, $userId);
 		$this->service = $service;
+		$this->columnService = $columnService;
 	}
 
 	/**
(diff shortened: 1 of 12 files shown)


lib/Controller/TableScheme.php
```diff
@@ -0,0 +1,53 @@
...
+use JsonSerializable;
+use OCA\Tables\Db\Column;
+
+class TableScheme implements JsonSerializable {
+
+	protected ?string $title = null;
+	protected ?string $emoji = null;
+
+	/** @var Column[]|null  */
+	protected ?array $columns = null;
```
juliushaertl, line 14: Can we also include views of a table?

lib/Controller/ApiTablesController.php
```diff
@@ -71,6 +76,73 @@ public function show(int $id): DataResponse {
...
+					$column['numberMin'],
+					$column['numberMax'],
+					$column['numberDecimals'],
+
+					$column['selectionOptions'] == [] ? '' : $column['selectionOptions'],
+					$column['selectionDefault'],
+
+					$column['datetimeDefault'],
+					[],
+				);};
```
juliushaertl, line 139: Would be good to wrap into a transaction. Also maybe we can move this to the service directly so we could also implement an occ command in the future which can reuse the logic?

src/modules/navigation/partials/NavigationTableItem.vue
```diff
@@ -235,6 +244,18 @@ export default {
 		createView() {
 			emit('tables:view:create', { tableId: this.table.id })
 		},
+
+		exportFile() {
+			axios.get(generateOcsUrl(`/apps/tables/api/2/tables/scheme/${this.table.id}`)).then(res => {
+				const blob = new Blob([JSON.stringify(res.data.ocs.data)], { type: 'application/json' })
```
juliushaertl, line 250: Maybe we can have a non-ocs controller that does the same as the API for exporting and use a Content-Disposition header for forcing to download. Putting this in a fake link seems risky with larger data sets

grnd-alt: I've created the non OCS endpoint and used the content-disposition header. Sadly I've not been able to somehow trigger the download without a fake link element. I've also not been able to find any solution on how to automatically download files via the browser without such a workaround. Also there's this slightly angry blog about how it's impossible to trigger downloads via the header using axios: https://medium.com/@drevets/you-cant-prompt-a-file-download-with-the-content-disposition-header-using-axios-xhr-sorry-56577aa706d6

grnd-alt: I've also tried doing the download via a fake form so that the browser handles the request. But when that's done the redirect just returns `csrf failed`. Idk how to send our request tokens etc. via form actions.

juliushaertl: In that case we can just add `?requesttoken=` with https://github.com/nextcloud-libraries/nextcloud-auth/blob/main/lib/requesttoken.ts#L15 to the URL.

lib/Controller/TableScheme.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
```
juliushaertl, line 2: Just rebased but this one should fail due to missing copyright headers. Best would be to always rebase to latest main before doing another push on a pr, so all checks and changes are up to date with the base branch :)

lib/Controller/TableScheme.php
```diff
@@ -0,0 +1,58 @@
...
+namespace OCA\Tables\Controller;
```
juliushaertl, line 3: This class is not really a controller, probably fits best in `OCA\Tables\Model`

lib/Controller/TableScheme.php
```diff
@@ -0,0 +1,58 @@
...
+	/** @var Column[]|null  */
+	protected ?array $columns = null;
+
+	/** @var View[]|null */
+	protected ?array $views = null;
+	protected ?string $description = null;
+	protected ?string $tablesVersion = null;
+
+	public function __construct(string $title, string $emoji, array $columns, array $view, string $description, IAppManager $appManager) {
+		$this->tablesVersion = $appManager->getAppVersion("tables");
```
juliushaertl, line 24: I'd prefer to directly pass the tables version to the model

juliushaertl:
Just noticed that we have one part of the requirement list from the issue missing, but I'd say we can get this in as a first step and then follow up with an additional pull request:

> Exporting an application likely should include the schema of all linked tables/views

juliushaertl (approved):
LGTM, but file follow up tickets for:

- [ ] Schema import/export of applications in tables
- [ ] Integration tests for the API endpoints

PR #1173, opened 2024-06-27
fix(Backend): use object over loose array for permissions

### LINKED ISSUE #1131: Undefined array key "manageTable" at ViewService.php#419
### Steps to reproduce

- `GET /apps/tables/view`

### Expected behavior

Nothing in the logs

### Actual behavior

Undefined array key "manageTable" at …/apps/tables/lib/Service/ViewService.php#419

### Tables app version

0.7.2

blizzz:
Solves issues with undefined array keys, but is said to be more performant but also more clear when coding.

fixes #1131

### DIFF
--- a/lib/Db/Table.php
+++ b/lib/Db/Table.php
@@ -3,6 +3,7 @@
 namespace OCA\Tables\Db;
 
 use JsonSerializable;
+use OCA\Tables\Model\Permissions;
 use OCA\Tables\ResponseDefinitions;
 use OCP\AppFramework\Db\Entity;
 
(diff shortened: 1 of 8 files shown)


lib/Db/Table.php
```diff
@@ -101,10 +102,9 @@ public function jsonSerialize(): array {
 	}
 
 	/**
-	 * @psalm-suppress MismatchingDocblockReturnType
-	 * @return array{read: bool, create: bool, update: bool, delete: bool, manage: bool}|null
+	 * @return ?Permissions
```
juliushaertl, line 105: No need as we have the type hint already.

juliushaertl (approved):

lib/Db/Table.php
blizzz: and removed 2, 3 others of the same

blizzz:
ok, backport does not work without pulling down more things from main.

PR #1176, opened 2024-07-01
fix: ensure default values show as filled

### LINKED ISSUE #993: Leaving default value qualifies as not filled
### Steps to reproduce

1. Create a new column of type numeric, for example "quantity".
2. Set default value to 1.
3. Set required to true
4. Add new row. Leave our field value as default (1).

### Expected behavior

Default value should be acceptable

### Actual behavior

You get an error, saying that you should fill all required fields

### Tables app version

0.6.6

### Browser

Firefox 124.0.2

### Client operating system

Any

### Operating system

Ubuntu 22.04

### Web server

Apache

### PHP engine version

PHP 8.1

### Database

PostgreSQL

### Additional info

_No response_

enjeck:
fixes #993

### DIFF
--- a/src/shared/components/ncTable/mixins/rowHelper.js
+++ b/src/shared/components/ncTable/mixins/rowHelper.js
@@ -11,8 +11,7 @@ export default {
 				) {
 					return true
 				}
-				const hasDefaultValue = columnTypeDefault in column && !(['', 'null'].includes(column[columnTypeDefault]))
-				return false || hasDefaultValue
+				return columnTypeDefault in column && !(['', 'null'].includes(column[columnTypeDefault]))
 			}
 			let hasDefaultValue = columnTypeDefault in column && !(['', null].includes(column[columnTypeDefault]))
 			if (column.type === ColumnTypes.SelectionMulti) {
--- a/src/shared/components/ncTable/mixins/rowHelper.js
+++ b/src/shared/components/ncTable/mixins/rowHelper.js
@@ -1,0 +1,36 @@
+import { ColumnTypes } from './columnHandler.js'
+export default {
+	methods: {
+		isValueValidForColumn(value, column) {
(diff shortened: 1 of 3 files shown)


elzody:
Seems reasonable, just a weird Cypress error. I am assuming it is related since it's affecting the `column-number.cy.js` test, but have not looked into it.

juliushaertl:
I am wondering if instead of filling the default value then we should rather use the placeholder attribute and andjudt the check to allow empty fields for values that have a default? That way a default value would also adapt if changed on the table level.

enjeck:
> I am wondering if instead of filling the default value then we should rather use the placeholder attribute and andjudt the check to allow empty fields for values that have a default? That way a default value would also adapt if changed on the table level.

I just noticed that this problem also affects other column types, not just numbers. Would look into making this approach more generic

src/shared/components/ncTable/mixins/rowHelper.js
```diff
@@ -0,0 +1,36 @@
...
+			const columnTypeDefault = type + 'Default'
+			if (column.type === ColumnTypes.Selection) {
+				if (
+					(value instanceof Array && value.length > 0)
+					|| (value === parseInt(value))
+				) {
+					return true
+				}
+				const hasDefaultValue = columnTypeDefault in column && !(['', 'null'].includes(column[columnTypeDefault]))
+				return false || hasDefaultValue
```
elzody, line 15: What's stopping from just returning `hasDefaultValue`? having the `false ||` seems redundant to me, since if `hasDefaultValue` is false, it'll just return false anyway. Unless I don't understand the logic correctly?

elzody (approved):

juliushaertl:
Don't be too hesitant to click the merge button if it is green 😁

PR #1179, opened 2024-07-03
chore: Drop Nextcloud 26 for the next release as it is EOL

juliushaertl:
Signed-off-by: Julius Härtl <jus@bitgrid.net>


blizzz (approved):
27 is also EOL btw. Could also drop?

juliushaertl:
Let me check that separately, but would generally agree 👍

PR #1187, opened 2024-07-08
chore/bump libraries

juliushaertl:
Splitting out the library bumps and adjustments from #830 

- **chore: Bump @nextcloud/dialogs and @nextcloud/files**
- **chore: Move to new file picker api for esm compatibility**
- **fix: Proper evaluation order when building errors strings**
- **fix: Fix missing actions for empty default slot**
- **fix: Add aria-label to pagination**


juliushaertl:
Will split in smaller PRs

Issue #1177, opened 2024-07-02
Ability to hide applications as a user


juliusknorr:
Leftover from contexts work.

@blizzz Could you document the current logic on what is possible for hiding the applications from the backend side in here so we can have a clear description on what we need to do on the frontend?

blizzz:
So on the backend side we have things in place, but disabled. To turn it on, we simply need:

* [ ] https://github.com/nextcloud/tables/pull/1193

On the frontend side we need

* [ ] a control for the context owner to set the default display mode. 
* [ ] a control per context for a user to set a personal overwrite (e.g. hide or enable)

For the record, we have three modes. The constants are define in `Application`:

```php
	public const NAV_ENTRY_MODE_HIDDEN = 0;
	public const NAV_ENTRY_MODE_RECIPIENTS = 1;
	public const NAV_ENTRY_MODE_ALL = 2;
``` 

- **NAV_ENTRY_MODE_HIDDEN** - no nav bar entry
- **NAV_ENTRY_MODE_RECIPIENTS** - nav bar entry for share recipients, but not the owner
- **NAV_ENTRY_MODE_ALL** - nav bar entry for everybody

Setting the Nav bar entry is possible via following routes:

- `api1#updateShareDisplayMode`
- `share#updateDisplayMode`

as well as when creating a share, via optional `displayMode` mode parameter. Default is hidden!

juliushaertl:
@enjeck As disscussed last week, something to look into after the user/group picker is ready :)

blizzz:
Added two tasks to https://github.com/nextcloud/tables/issues/1177#issuecomment-2225363092

PR #1193, opened 2024-07-12
feat(Contexts): enable nav bar display logic

blizzz:
contributes to #1177

It was temporarily disabled and shown by default as we did not have web UI controls to configure it.

* [x] move some logic into the mapper, [cf. TODO in code](https://github.com/nextcloud/tables/compare/enh/1177/enable-nav-display-logic?expand=1#diff-a4df385ae09b07cf4c026556ce1d2eb17b955736ea7f57e0f925067af9659225R47)
* [ ] :warning:  merge only with the frontend tasks as laid out in https://github.com/nextcloud/tables/issues/1177#issuecomment-2225363092

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -121,6 +121,8 @@ return [
 		['name' => 'search#all', 'url' => '/search/all', 'verb' => 'GET'],
 	],
 	'ocs' => [
+		['name' => 'navigation#getAppsNavigation', 'url' => '/navigation', 'verb' => 'GET'],
+
 		// API v2
 		['name' => 'ApiGeneral#index', 'url' => '/api/2/init', 'verb' => 'GET'],
 		// -> tables
--- a/cypress/e2e/context.cy.js
+++ b/cypress/e2e/context.cy.js
@@ -76,7 +76,6 @@ describe('Manage a context', () => {
         cy.login(nonLocalUser)
         cy.visit('apps/tables')
         cy.loadContext(contextTitle)
-        cy.contains('header .app-menu-entry', contextTitle).should('exist')
         cy.contains('h1', contextTitle).should('exist')
         cy.contains('h1', tableTitle).should('exist')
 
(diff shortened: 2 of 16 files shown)


juliushaertl (approved):
Backend change looks good 👍

juliushaertl (changes requested):
Let me block the merge just to be sure this doesn't get lost from the initial post

⚠️ merge only with the frontend tasks as laid out in https://github.com/nextcloud/tables/issues/1177#issuecomment-2225363092

juliushaertl:
Great, thanks :)

blizzz:
> It seems there's no way for an owner to modify their navigation entry if there are no shares? Creating an application without shares means there's no navigation entry, and no way to change it. @blizzz

This is accurate. We bound the nav bar display state to sharing. Also the share id is part of the primary key.

I do see it is reasonable to have this setting for non-shared applications, but spontaneously a quick solution that is not too terrible does not come to my mind. Is it a blocker?

blizzz:
> > > It seems there's no way for an owner to modify their navigation entry if there are no shares? Creating an application without shares means there's no navigation entry, and no way to change it. @blizzz
> > 
> > 
> > This is accurate. We bound the nav bar display state to sharing. Also the share id is part of the primary key.
> > I do see it is reasonable to have this setting for non-shared applications, but spontaneously a quick solution that is not too terrible does not come to my mind. Is it a blocker?
> 
> Maybe not a blocker, but still inconvenient and not very user-friendly. The UI right now doesn't look great, and I can't seem to fix it due to this: #1295

Yup, I am with you there. Need to think about a good strategy.

blizzz:
Screenshot_20241119_212920

@enjeck I feel a bit disconnected. That toggle there, changing it does not have any effect via GUI.

blizzz:
Remaining one: https://github.com/nextcloud/tables/pull/1250, it should go together with this PR

juliusknorr (approved):

PR #1195, opened 2024-07-12
fix: Navigation table entries UI issues

### LINKED ISSUE #1191: Navigation table entries UI issues
The app navigation has a few small quirks with the new UI adjustments:

- [ ] The counter (3 rows) is cut off, we should probably just use the number as it is done in other apps without the word rows
- [ ] text and icon color is inconsistent in dark mode (should use the proper css var for text on primary)

image

image

enjeck:
fixes #1191 

- [x] The counter (3 rows) is cut off, we should probably just use the number as it is done in other apps without the word rows

- [x]  text and icon color is inconsistent in dark mode (should use the proper css var for text on primary)

#### Before
image

#### After
Screenshot from 2024-07-15 13-38-18

### DIFF
--- a/src/modules/navigation/partials/NavigationTableItem.vue
+++ b/src/modules/navigation/partials/NavigationTableItem.vue
@@ -20,7 +20,7 @@
 
 		<template #counter>
 			<NcCounterBubble v-if="canReadData(table)">
-				{{ n('tables', ' %n ',table.rowsCount) }}
+				{{ table.rowsCount }}
 			</NcCounterBubble>
 			<NcActionButton v-if="table.hasShares" icon="icon-share" :class="{'margin-right': !(activeTable && table.id === activeTable.id)}" @click="actionShowShare" />
 			<div v-if="table.isShared && table.ownership !== userId" class="margin-left">
@@ -279,7 +279,7 @@ export default {
 
 .app-navigation-entry.active {
 	.icon-share  {
-		background-image: var(--icon-share-white);
+		background-image: var(--icon-share-white)
 	}
 
 	.icon-collapse {
--- a/src/modules/navigation/partials/NavigationTableItem.vue
+++ b/src/modules/navigation/partials/NavigationTableItem.vue
@@ -20,7 +20,7 @@
 
 		<template #counter>
 			<NcCounterBubble v-if="canReadData(table)">
-				{{ n('tables', '%n row', '%n rows', table.rowsCount, {}) }}
+				{{ n('tables', ' %n ',table.rowsCount) }}
 			</NcCounterBubble>
 			<NcActionButton v-if="table.hasShares" icon="icon-share" :class="{'margin-right': !(activeTable && table.id === activeTable.id)}" @click="actionShowShare" />
 			<div v-if="table.isShared && table.ownership !== userId" class="margin-left">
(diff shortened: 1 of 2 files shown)


enjeck:
- [ ] Check if the issue affects views too, and fix if so

juliushaertl:
It looks like you still build your local app with either outdated server code or node dependencies on tables, as it still uses the old border radius. Could you check that? Just to make sure the screenshots actually show what we want to go for with Nextcloud 30

src/modules/navigation/partials/NavigationTableItem.vue
```diff
@@ -20,7 +20,7 @@
 
 		<template #counter>
 			<NcCounterBubble v-if="canReadData(table)">
-				{{ n('tables', '%n row', '%n rows', table.rowsCount, {}) }}
+				{{ n('tables', ' %n ',table.rowsCount) }}
```
juliushaertl, line 23:
```suggestion
				{{ table.rowsCount }}
```

As we don't need any translation of that anymore

src/modules/navigation/partials/NavigationTableItem.vue
```diff
@@ -277,6 +277,16 @@ export default {
 </script>
 <style lang="scss">
 
+.app-navigation-entry.active {
+	.icon-share  {
+		background-image: var(--icon-share-white);
+	}
+
+	.icon-collapse {
+		color: var(--color-primary-element-text)
```
juliushaertl, line 286:
```suggestion
		color: var(--color-primary-element-text);
```

src/modules/navigation/partials/NavigationTableItem.vue
```diff
@@ -277,6 +277,16 @@ export default {
 </script>
 <style lang="scss">
 
+.app-navigation-entry.active {
+	.icon-share  {
```
juliushaertl, line 281: Ideally we would migrate away from the icon property and instead use the icon slot to provide an inline material icon. That way it can just pick up the text color as with the collapse icon.

src/modules/navigation/partials/NavigationViewItem.vue
```diff
@@ -15,7 +15,7 @@
 		</template>
 		<template #counter>
 			<NcCounterBubble v-if="canReadData(view)">
-				{{ n('tables', '%n row', '%n rows', view.rowsCount, {}) }}
+				{{ n('tables', ' %n ', view.rowsCount) }}
```
juliushaertl, line 18:
```suggestion
				{{ view.rowsCount }}
```

src/modules/navigation/partials/NavigationTableItem.vue
enjeck: Uh, icon slot? Not sure what that is :confused:

juliushaertl: Checkout like 36 for the next entry in the actions here, NcActionButton has two ways of providing an icon, as a css class name or with the icon slot where we can use material inline svg

juliushaertl (approved):

src/modules/navigation/partials/NavigationTableItem.vue
juliushaertl: Sorry I thought we can get the other fixes in already :) A follow up PR would still be nice

https://pictogrammers.com/library/mdi/icon/account-plus/ should be there as AccountPlus.vue in the library

PR #1205, opened 2024-07-16
fix(api): Fix wrong array type

provokateurin:
Fixes https://github.com/nextcloud/tables/pull/1169


provokateurin:
I found even more instances of the original problem, fixed those as well.

blizzz (approved):

PR #1217, opened 2024-07-18
chore: Proper logs with error level and exception trace

juliushaertl:
While testing a PR i noticed error logging is quite odd with just the message being logged without an exception trace.

Mostly moving from warning to error for 500 status codes in logging and adding the exception trace to it.

This might cause duplicate logging for cases where an InternalError is thrown somewhere deeper but that seems better for now then missing an exception trace.

@blizzz We probably want to cleanup the way exceptions are handled there, I feel the generic InternalError is not very helpful there and could just be a generic Exception. Then we can catch and log all non-specific exceptions in the middleware or in one unified controller method. WDYT?

### DIFF
--- a/lib/Controller/AOCSController.php
+++ b/lib/Controller/AOCSController.php
@@ -38,7 +38,7 @@ abstract class AOCSController extends OCSController {
 	 * @return DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
 	 */
 	protected function handleError($e): DataResponse {
-		$this->logger->warning('An internal error or exception occurred: ['. $e->getCode() . ']' . $e->getMessage());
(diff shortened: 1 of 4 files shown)


blizzz (approved):

blizzz:
> @blizzz We probably want to cleanup the way exceptions are handled there, I feel the generic InternalError is not very helpful there and could just be a generic Exception. Then we can catch and log all non-specific exceptions in the middleware or in one unified controller method. WDYT?

The InternalError is also disturbing me, and I think commonly I experienced it in the past when it was not an internal issue actually. So, yes, I am all for improving the situation. Whether it should be a generic one, I am not sure yet, but having them dealt with in a Middleware sounds nice. Maybe it is a bit hidden, if you are not aware about it, but it is least in your way when working with Controllers and removes a lot of boiler plate.

Issue #1214, opened 2024-07-18
"New table dialog", "Applications create dialog": Icon picker is misaligned


JuliaKirschenheuter:
### Is your feature request related to a problem? Please describe.

![Screenshot from 2024-07-18 11-15-11](https://github.com/user-attachments/assets/b0ac0719-c2f6-48e5-974c-c84c6e5d48b4)
![Screenshot from 2024-07-18 11-15-31](https://github.com/user-attachments/assets/e22bd2b8-5e57-4e56-b3a9-a32146d7fc93)

### Describe the solution you'd like

Align items to center

PR #1219, opened 2024-07-19
fix(files): align icon to center

JuliaKirschenheuter:
Fixes https://github.com/nextcloud/tables/issues/1214

|  Before   | After                                                        |
|-------------|---------------------------------------------------------------|
| Screenshot from 2024-07-18 11-15-11 | Screenshot from 2024-07-26 10-48-10 |-------------|---------------------------------------------------------------|
| Screenshot from 2024-07-18 11-15-31 | Screenshot from 2024-07-26 10-47-57 |-------------|---------------------------------------------------------------|

### DIFF
--- a/src/modules/modals/CreateContext.vue
+++ b/src/modules/modals/CreateContext.vue
@@ -10,7 +10,7 @@
 				<div class="col-4 mandatory">
 					{{ t('tables', 'Title') }}
 				</div>
-				<div class="col-4" style="display: inline-flex;">
+				<div class="col-4 content-emoji">
 					<NcIconPicker :close-on-select="true" @select="setIcon">
 						<NcButton
 							type="tertiary"
@@ -167,5 +167,10 @@ export default {
 <style lang="scss" scoped>
 .modal__content {
 	padding-right: 0 !important;
+
+	.content-emoji {
+		display: inline-flex;
+		align-items: center;
+	}
 }
 </style>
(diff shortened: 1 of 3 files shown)


JuliaKirschenheuter:
thanks @enjeck!

> I think there should be more space between the Title input field and the "Description" label in the Create/Edit table modal

I haven't seen it in requirement list

JuliaKirschenheuter:
> From the docs here https://nextcloud-vue-components.netlify.app/#/Components/NcFields?id=nctextfield, we can't use it? Or maybe I misunderstand.

No sure if i understood you correct, i mean something like this:

Screenshot from 2024-07-24 18-54-32

instead of:

Screenshot from 2024-07-24 18-56-15

juliushaertl:
> > nctextfield
> 
> From the docs here [nextcloud-vue-components.netlify.app#/Components/NcFields?id=nctextfield](https://nextcloud-vue-components.netlify.app/#/Components/NcFields?id=nctextfield), we can't use it? Or maybe I misunderstand.
> 
> Also, if we want to change it here, would be nice to change it everywhere else for consistency

Probably the docs are a bit confusing, this message only referrs to NcInputField, NcTextField is fine to be used (scroll further down for the actual docs of that)

Screenshot 2024-07-25 at 16 03 51

JuliaKirschenheuter:
> I think there should be more space between the Title input field and the "Description" label in the Create/Edit table

done ;)

Issue #1212, opened 2024-07-18
Table view: when scrolling the table is seen through the header


JuliaKirschenheuter:
### Is your feature request related to a problem? Please describe.

![Peek 2024-07-18 09-57](https://github.com/user-attachments/assets/c2682ce2-5db7-45bc-83fe-4e15c0080a81)

### Describe the solution you'd like

Header have to be fixed without overlapping table content

PR #1220, opened 2024-07-19
fix(files): fix width, background of table row and interval between table header

JuliaKirschenheuter:
Fixes https://github.com/nextcloud/tables/issues/1212

|  Before   | After                                                        |
|-------------|---------------------------------------------------------------|
| Screenshot from 2024-07-19 11-24-09 | Screenshot from 2024-07-19 12-21-27 |-------------|---------------------------------------------------------------|
| Screenshot from 2024-07-19 11-18-16

### DIFF
--- a/src/shared/components/ncTable/sections/CustomTable.vue
+++ b/src/shared/components/ncTable/sections/CustomTable.vue
@@ -1,7 +1,7 @@
 <template>
 	<div class="container">
-		<table>
-			<thead>
+		<table class="tables-list__table">
+			<thead class="tables-list__thead">
 				<TableHeader :columns="columns"
 					:selected-rows="selectedRows"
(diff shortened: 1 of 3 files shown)


src/shared/components/ncTable/sections/CustomTable.vue
```diff
@@ -464,7 +464,7 @@ export default {
 
 	thead {
 		position: sticky;
-		top: 118px;
+		top: 108px;
```
JuliaKirschenheuter, line 467: not sure about this one

juliushaertl: I'd say fine as a quick fix, Ideally we could define a var with the height of the table title element and reuse that then

PR #1215, opened 2024-07-18
fix(Capabilities): announce API 2.0

blizzz:

PR #1221, opened 2024-07-19
build(openapi): fix ms typescript generation command

blizzz:
fixes the silent error when generating the openapi assets as described here: https://github.com/nextcloud/tables/pull/1215#issuecomment-2236178507

0.7 is not affected, this happened after upgrading a dev dep here on main

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "watch": "NODE_ENV=development webpack --progress --watch --config webpack.config.js",
     "serve": "NODE_ENV=development webpack serve --progress --config webpack.config.js",
     "typescript:check": "tsc --noEmit",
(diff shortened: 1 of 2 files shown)


provokateurin (approved):

provokateurin:
Can you shortly explain why the previous command wasn't working?

blizzz:
> Can you shortly explain why the previous command wasn't working?

Two things:

1.
>  ✘  Globbing has been deprecated in favor of redocly.yaml’s `apis` keys

Did not like the asterisk with `openapi*.json` as input. Since we only have one input file, we do not need it.

2.
> Error: EISDIR: illegal operation on a directory, open '/srv/http/nextcloud/master/apps-repos/tables/src/types/openapi/'
    at Object.writeFileSync (node:fs:2368:20)

Expects now a file name as  `--output|-o` parameter, previously a folder was fine.

Issue #1232, opened 2024-07-23
Table view: minimum row height should be 50px like in files


JuliaKirschenheuter:
### Is your feature request related to a problem? Please describe.

Current state: 

![Screenshot from 2024-07-23 09-37-49](https://github.com/user-attachments/assets/cb5c88de-0c77-4997-81a0-48d833f930a0)

### Describe the solution you'd like

Table row should be without extra space at the top and bottom

PR #1234, opened 2024-07-23
fix(files): remove unneeded extra place for the rows

JuliaKirschenheuter:
Fixes https://github.com/nextcloud/tables/issues/1232

## Screenshots

Before | After
---|---
Screenshot from 2024-07-23 09-37-49 | Screenshot from 2024-07-23 09-44-23

### DIFF
--- a/src/shared/components/ncTable/sections/CustomTable.vue
+++ b/src/shared/components/ncTable/sections/CustomTable.vue
@@ -389,6 +389,14 @@ export default {
 	}
 }
 
+:deep(.text-editor__wrapper .paragraph-content:last-child) {
+	margin-bottom: 0!important;
+}
+
+:deep(.text-editor__wrapper .ProseMirror > *:first-child) {
+	margin-top: 0!important;
+}
+
 .selected-page{
 	padding-left: 5px;
 


src/shared/components/ncTable/sections/CustomTable.vue
```diff
@@ -464,7 +471,6 @@ export default {
 
 	thead {
 		position: sticky;
-		top: 118px;
```
juliushaertl, line 467: Adjusted in the other PR, let's remove this from here

src/shared/components/ncTable/sections/CustomTable.vue
```diff
@@ -389,6 +389,13 @@ export default {
 	}
 }
 
+:deep(.text-editor__wrapper .paragraph-content) {
+	margin-bottom: 0!important;
+}
+:deep(.text-editor__wrapper div.ProseMirror > *:first-child) {
+	margin-top: 0!important;
+}
```
juliushaertl, line 397: Not sure I fully get those two rules try to achieve, shouldn't we just target something like:

```
:deep(.text-editor__wrapper .paragraph-content:last-child)
```

src/shared/components/ncTable/sections/CustomTable.vue
JuliaKirschenheuter: already removed 🤔

juliushaertl: Sorry was referring remove the change of removing this line from the commit. It actually causes a conflict now which just would need to be resolved.

src/shared/components/ncTable/sections/CustomTable.vue
JuliaKirschenheuter: thank you! Took your suggestion!

Regarding:

```
:deep(.text-editor__wrapper div.ProseMirror > *:first-child) {
	margin-top: 0!important;
}
```

still think this margin-top have to be removed too
Screenshot from 2024-07-23 14-56-35

juliushaertl: Ah right, that makes sense to drop then as well of course 👍

juliushaertl (approved):

PR #1250, opened 2024-07-29
Make Applications visible by default in the app navigation bar

blizzz:
Contributes to #1177 

In 0.7.* we are displaying Tables Applications by default and without further exposed options by default in the navigation bar. The intended default was hidden, but keeping it would break the experience, and may make them less discoverable in general.


lib/Controller/Api1Controller.php
```diff
@@ -1401,7 +1413,20 @@ public function importInView(int $viewId, string $path, bool $createMissingColum
...
+					$tableId,
+					'table',
+					$receiver,
+					$receiverType,
+					$permissionRead,
+					$permissionCreate,
+					$permissionUpdate,
+					$permissionDelete,
+					$permissionManage,
+					Application::NAV_ENTRY_MODE_ALL
```
blizzz, line 1427: FWIW it is not really applicable here (table share…), but have the same default may lead to less confusion in the future us's.

blizzz:
Mh, OpenAPI (not committed yet) has an issue with the constant. Neither does the command deduct the default value, but it also considers it required, instead of optional:

```diff
diff --git a/openapi.json b/openapi.json
index 652d6d5c..ea5b1c45 100644
--- a/openapi.json
+++ b/openapi.json
@@ -2439,7 +2439,8 @@
                                     "nodeId",
                                     "nodeType",
                                     "receiver",
-                                    "receiverType"
+                                    "receiverType",
+                                    "displayMode"
                                 ],
                                 "properties": {
                                     "nodeId": {
@@ -2496,7 +2497,6 @@
                                     "displayMode": {
                                         "type": "integer",
                                         "format": "int64",
-                                        "default": 0,
                                         "description": "context shares only, whether it should appear in nav bar. 0: no, 1: recipients, 2: all"
                                     }
                                 }
```

blizzz:
> Mh, OpenAPI (not committed yet) has an issue with the constant. Neither does the command deduct the default value, but it also considers it required, instead of optional:
> 
> ```diff
> diff --git a/openapi.json b/openapi.json
> index 652d6d5c..ea5b1c45 100644
> --- a/openapi.json
> +++ b/openapi.json
> @@ -2439,7 +2439,8 @@
>                                      "nodeId",
>                                      "nodeType",
>                                      "receiver",
> -                                    "receiverType"
> +                                    "receiverType",
> +                                    "displayMode"
>                                  ],
>                                  "properties": {
>                                      "nodeId": {
> @@ -2496,7 +2497,6 @@
>                                      "displayMode": {
>                                          "type": "integer",
>                                          "format": "int64",
> -                                        "default": 0,
>                                          "description": "context shares only, whether it should appear in nav bar. 0: no, 1: recipients, 2: all"
>                                      }
>                                  }
> ```

I digged down that hole. In the PHP parser used, we can get the class and the name of the constant, but not its value. Reflection is not an option there. So I changed it back to literal value in the Api1Controller.

juliusknorr (approved):

PR #1277, opened 2024-08-05
Add reuse compliance - Part 1

AndyScherzinger:
As a first step convert a larger chunk of files to ship SPDX header where our converter won't be of any help...

---

# SUMMARY

* Bad licenses: 0
* Deprecated licenses: 0
* Licenses without file extension: 0
* Missing licenses: 0
* Unused licenses: 0
* Used licenses: Apache-2.0, AGPL-3.0-or-later, MIT
* Read errors: 0
* Files with copyright information: 604 / 1179
* Files with license information: 580 / 1179

### DIFF
--- a/.devcontainer/setup.sh
+++ b/.devcontainer/setup.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
-
+#
+# SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: AGPL-3.0-or-later
 (
     cd /tmp && /usr/local/bin/bootstrap.sh apache2ctl start
 )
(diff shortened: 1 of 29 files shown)


AndyScherzinger:
one question maybe @elzody can help me with that. Any particular reason to replicate the server repo's image assets in https://github.com/nextcloud/tables/tree/main/cypress/styleguide/assets/img ? Just asking because I will have to replicate its license infos then.

elzody:
@AndyScherzinger That was needed in order to support Cypress component tests for the interactive table widget, since the component tests don't run in a full-fledged e2e set up -- they just run by themselves, meaning we needed to import the styles, etc.

I can see now with the headers how it is problematic and creates a lot of work... not really sure initially if there's a better way to do it, since an e2e test would create a reliance on accessing the widget from another app. I can look into if there's a simpler way to do it so you don't have to re-do all of those, if possible

AndyScherzinger:
@elzody all good, no worries. Adding the license infos for spdx/reuse isn't an issue or a lot of effort since that info is already available from/in the server repo. So don't worry. Fine to keep the files as they are at the moment 👍

Issue #1236, opened 2024-07-24
Refactor column creation function


luka-nextcloud:
### Is your feature request related to a problem? Please describe.

Refactor column creation function to avoid passing many arguments

blizzz:
Stems from this comment https://github.com/nextcloud/tables/pull/944#pullrequestreview-2193607823

First, I agree, there's too many arguments there. Perhaps just reduce this to the very necessary ones, and all optional could be set via dedicated setters on the result object.

But we would have to call all of them anyway, wouldn't we? At least in that case. Otherwise names arguments can be used already (for immediate cure).

Anyways, the methods need an overhaul.

luka-nextcloud:
@juliushaertl @blizzz I can see that function update (https://github.com/nextcloud/tables/blob/main/lib/Service/ColumnService.php#L345) also has many arguments. Should we refactor this function as well?

juliushaertl:
Ideally yes, maybe we can come up with a DTO class to contain the column definition that we can pass along instead in other places as well

PR #1281, opened 2024-08-08
fix: reduce arguments for column creation function

luka-nextcloud:
* Resolves: #1236


lib/Service/ColumnService.php
```diff
@@ -259,36 +212,36 @@ public function create(
...
-		$item->setUsergroupSelectGroups($usergroupSelectGroups);
-		$item->setShowUserStatus($showUserStatus);
+		$item->setSelectionOptions($columnDto->getSelectionOptions());
+		$item->setSelectionDefault($columnDto->getSelectionDefault());
+		$item->setDatetimeDefault($columnDto->getDatetimeDefault());
+		$item->setUsergroupDefault($columnDto->getUsergroupDefault());
+		$item->setUsergroupMultipleItems($columnDto->getUsergroupMultipleItems());
+		$item->setUsergroupSelectUsers($columnDto->getUsergroupSelectUsers());
+		$item->setUsergroupSelectGroups($columnDto->getUsergroupSelectGroups());
+		$item->setShowUserStatus($columnDto->getShowUserStatus());
```
juliushaertl, line 244: Maybe we can also move this assignment logic to the Column class to have something like `$item->setColumnSchema($columnDto)`?

juliushaertl: Or rather have a `ColumnEntity::fromDto` method?

juliushaertl (commented):
Left some comments but would also appreciate some feedback from @blizzz on how to approach this

Issue #1257, opened 2024-08-01
TypeError: this.$el.querySelector is not a function


SystemKeeper:
### Steps to reproduce

1. Create a new table
2. Click on "Create row"
3. Close the dialog again

Tested on daily.

### Expected behavior

No error

### Actual behavior

<img width="1210" alt="image" src="https://github.com/user-attachments/assets/232ad750-95e4-437a-87be-07058f7a6e91">

### Web server

None

### PHP engine version

None

### Database

None

juliushaertl:
Fix added to #1259

PR #1282, opened 2024-08-09
fix: only focus on input if modal is open

enjeck:
fixes https://github.com/nextcloud/tables/issues/1257

### DIFF
--- a/src/modules/modals/CreateColumn.vue
+++ b/src/modules/modals/CreateColumn.vue
@@ -260,9 +260,11 @@ export default {
 			deep: true,
 		},
 		showModal() {
-			this.$nextTick(() => {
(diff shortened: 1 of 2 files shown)


elzody (approved):
Simple enough, couldn't find a better solution myself. Not sure why the optional chaining on the function call doesn't work :D

I think to fix the tests it just needs an `npm run lint:fix`

juliushaertl:
Ideally squash the commits or use `git commit --amend` or [fixup commits](https://blog.sebastian-daschner.com/entries/git-commit-fixup-autosquash) to have a clean git history. That way it is much easier whenever someone needs to refer back to the change which can then be one atomic commit.

PR #1293, opened 2024-08-14
fix: unhide part of single selection NcSelect value

enjeck:
When editing a row with a single selection, part of it is hidden. This unhides it. 

Before:
Screenshot from 2024-08-14 15-21-20

After:
Screenshot from 2024-08-14 15-22-52

### DIFF
--- a/src/modules/modals/EditRow.vue
+++ b/src/modules/modals/EditRow.vue
@@ -242,5 +242,10 @@ export default {
 	:where(.slot.fix-col-4 input, .slot.fix-col-4 .row) {
 		min-width: 100% !important;
 	}
+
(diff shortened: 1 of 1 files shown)


juliushaertl:
The text does not look fully centered vertically and the select input looks a bit too large (considering that the element size has been reduced with nextcloud 30). Could you check that as well? Maybe we are overwriting styles somewhere here as at least the examples in https://nextcloud-vue-components.netlify.app/#/Components/NcSelect?id=ncselect-1 seem fine

juliushaertl (dismissed):
Forgot to request changes for my comment ;)

juliushaertl (approved):

PR #1295, opened 2024-08-16
feat: Add UI elements to modify navigation display

enjeck:
Part of https://github.com/nextcloud/tables/issues/1177 and https://github.com/nextcloud/tables/pull/1193

Todo
- [x] ensure it works when you update a context
- [x] ensure it works for owners and recipients too
- [x] Discuss the UI and copy
- [x] Ensure self-shares to the owner are deleted when share is deleted

### DIFF
--- a/lib/Controller/NavigationController.php
+++ b/lib/Controller/NavigationController.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace OCA\Tables\Controller;
+
+use OCA\Tables\Service\ContextService;
+use OCP\AppFramework\Http\Attribute\NoAdminRequired;
+use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
+use OCP\AppFramework\Http\Attribute\OpenAPI;
+use OCP\AppFramework\Http\DataResponse;
+use OCP\INavigationManager;
+use OCP\IRequest;
+use OCP\IURLGenerator;
+use OCP\IUserSession;
+
+/**
+ * This is a workaround until https://github.com/nextcloud/server/pull/49904 is
+ * settled in all covered NC versions; expected >= 31.
+ */
+class NavigationController extends \OC\Core\Controller\NavigationController {
+	public function __construct(
+		protected ContextService $contextService,
+		protected IUserSession $userSession,
+		string $appName,
+		IRequest $request,
+		INavigationManager $navigationManager,
+		IURLGenerator $urlGenerator
+	) {
+		parent::__construct($appName, $request, $navigationManager, $urlGenerator);
+	}
+
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	#[OpenAPI(scope: OpenAPI::SCOPE_IGNORE)]
+	public function getAppsNavigation(bool $absolute = false): DataResponse {
+		$this->contextService->addToNavigation($this->userSession->getUser()?->getUID());
+		return parent::getAppsNavigation($absolute);
+	}
+}
--- a/lib/Controller/NavigationController.php
+++ b/lib/Controller/NavigationController.php
@@ -1,5 +1,10 @@
 <?php
 
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
 namespace OCA\Tables\Controller;
 
 use OCA\Tables\Service\ContextService;
--- a/src/modules/modals/CreateContext.vue
+++ b/src/modules/modals/CreateContext.vue
@@ -171,7 +171,7 @@ export default {
 					isUser: true,
 					key: 'user-' + getCurrentUser().uid,
 				})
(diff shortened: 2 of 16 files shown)


jancborchardt (changes requested):
@enjeck nice!

So unless I misunderstand something, the settings seem can be simplified. As admins are also just users, they can use the navigation preference like everyone else.

The main settings should only say:
> [ ] Show in app list

As a single checkbox, no subheading needed. It can have a color-text-maxcontrast sentence below saying "This can be overridden by a per-account preference."


For the left navigation entry, same wording and setting
> [ ] Show in app list

blizzz:
One little thing needed: with switching away from the three modes to just two modes, we also have to remove this condition: 

https://github.com/nextcloud/tables/blob/4b7e5a7734558478bbdef447cc16c2a47b41acd6/lib/Db/ContextMapper.php#L189-L190

Otherwise a share owner cannot remove the entry for themselves. Works fine then!

In the backend, we should remove the third option, now that is is unnecessary.

src/shared/constants.js
```diff
@@ -25,3 +25,9 @@ export const TYPE_TEXT = 'text'
 export const TYPE_NUMBER = 'number'
 export const TYPE_DATETIME = 'datetime'
 export const TYPE_USERGROUP = 'usergroup'
+
+export const NAV_ENTRY_MODE = {
+	NAV_ENTRY_MODE_HIDDEN: 0, // no nav bar entry
+	NAV_ENTRY_MODE_RECIPIENTS: 1, // nav bar entry for share recipients, but not the owner
```
blizzz, line 31: since we now only set HIDDEN or ALL, we can remove RECIPIENTS and its usages, right? (same for backend as written in the other comment). :broom:

blizzz:
Related: https://github.com/nextcloud/server/pull/49904

lib/Controller/NavigationController.php
```diff
@@ -0,0 +1,43 @@
...
+use OCP\INavigationManager;
+use OCP\IRequest;
+use OCP\IURLGenerator;
+use OCP\IUserSession;
+
+/**
+ * This is a workaround until https://github.com/nextcloud/server/pull/49904 is
+ * settled in all covered NC versions; expected >= 31.
+ */
+class NavigationController extends \OC\Core\Controller\NavigationController {
```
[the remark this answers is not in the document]
blizzz: By suppression only so far. I pushed a commit that updates the baseline accordingly. The command I ran is `composer run psalm -- --set-baseline=tests/psalm-baseline.xml`.

src/modules/modals/CreateContext.vue
```diff
@@ -42,6 +42,14 @@
 				</div>
 				<NcContextResource :resources.sync="resources" :receivers.sync="receivers" />
 			</div>
+			<div class="row space-T">
+				<NcActionCheckbox :checked="showInNavigationDefault" @change="updateDisplayMode">
+					Show in app list
```
juliusknorr, line 47: Should be translatable

src/modules/modals/CreateContext.vue
```diff
@@ -42,6 +42,14 @@
...
+				</NcActionCheckbox>
+				<p class="nav-display-subtext">
+					This can be overridden by a per-account preference
```
juliusknorr, line 50: Should be translatable

src/modules/modals/EditContext.vue
```diff
@@ -38,7 +38,14 @@
 				</div>
 				<NcContextResource :resources.sync="resources" :receivers.sync="receivers" />
 			</div>
-
+			<div class="row space-T">
+				<NcActionCheckbox :checked="showInNavigationDefault" @change="updateDisplayMode">
+					Show in app list
```
juliusknorr, line 43: Should be translatable

src/modules/modals/EditContext.vue
```diff
@@ -38,7 +38,14 @@
...
+				</NcActionCheckbox>
+				<p class="nav-display-subtext">
+					This can be overridden by a per-account preference
```
juliusknorr, line 46: Should be translatable

src/modules/navigation/partials/NavigationContextItem.vue
```diff
@@ -27,17 +28,21 @@
...
-			<NcActionButton v-if="ownsContext(context)" :close-after-click="true" data-cy="navigationContextDeleteBtn" @click="deleteContext">
+			<NcActionButton v-if="ownsContext(context)" :close-after-click="true" data-cy="navigationContextDeleteBtn"
+				@click="deleteContext">
 				<template #icon>
 					<Delete :size="20" />
 				</template>
 				{{ t('tables', 'Delete application') }}
 			</NcActionButton>
+			<NcActionCheckbox :checked="showInNavigation" @change="updateDisplayMode">
+				Show in app list
```
juliusknorr, line 39: Should be translatable

juliusknorr (commented):
Small nitpicks, but otherwise looks good 👍

juliusknorr (approved):
Tested and works. Also reviewed the backend bits of the base PR.

PR #1321, opened 2024-08-26
repair schema export and import

grnd-alt:
schema exporting and importing is broken on main, this should fix it + add the ability to rename the imported schema directly + give more user feedback on the selected file


juliushaertl:
Can we also have a test covering this? :)

juliushaertl (approved):

PR #1327, opened 2024-08-29
bump version to 0.8.0-beta-3

grnd-alt:


blizzz:
> Relevant test seems to be failing 🙈 But looks like just due to the version bump

Which version bump? Not server, as we declare compat between 27 and 31.

juliushaertl:
Tables bump, the test is about export which contains the version of the app in the json structure and is compared against some text fixture with the old version

grnd-alt:
@blizzz I added a test to check if the export import of schemes work, the json that exports the scheme contains the tablesVersion, so comparing the exported json with the expected failed as the version was different. now the tablesVersion is ignored in the test :D

PR #1330, opened 2024-08-29
ci: Update npm-audit-fix.yml to match tables app branches

juliushaertl:
@blizzz I think we rather want it this way


blizzz (approved):
0.8 is not branched out so far, but does not hurt i guess.

PR #1383, opened 2024-09-25
fix(Import): expect selectionOptions as array from json

blizzz:
currently an import fails silently (no web UI feedback) with `Argument #15 ($selectionOptions) must be of type ?string, array given` in the logs. This fixes handlig of the selectionOptions.


blizzz:
We should check with the usergroup column types, too.

blizzz:
Changed to Luka's suggestion

blizzz:
@luka-nextcloud mind reviewing this one? :see_no_evil:

luka-nextcloud (approved):

Issue #1367, opened 2024-09-18
Invalid openapi specification


Leptopoda:
### Steps to reproduce

The openapi spec states that a request to `/index.php/apps/tables/api/1/tables/{tableId}/rows/simple` returns a List of strings.
https://github.com/nextcloud/tables/blob/5cc780ce4260c262e3b733f5cb9d7959a9262938/openapi.json#L3840-L3853
https://github.com/nextcloud/tables/blob/5cc780ce4260c262e3b733f5cb9d7959a9262938/lib/Controller/Api1Controller.php#L1032-L1060

But making an actual request returns a List of List of String|int.
```json
[
  [
    "What",
    "How to do",
    "Ease of use",
    "Done"
  ],
  [
    "Open the tables app",
    "Click on tables icon in the menu bar.",
    5,
    "true"
  ],
  [
    "Add your first row",
    "Just click on \"new row\" and enter some data inside of the form. At the end click on the bottom \"save\".",
    5,
    "false"
  ],
  [
    "Edit a row",
    "Hover the mouse over a row you want to edit. Click on the pen on the right side. Maybe you want to add a \"done\" status to this row.",
    5,
    "false"
  ],
  [
    "Add a new column",
    "You can add, remove and adjust columns as you need it. Click on the three-dot-menu on the upper right of this table and choose \"create column\". Fill in the data you want, at least a title and column type.",
    4,
    "false"
  ],
  [
    "Read the docs",
    "If you want to go through the documentation, this can be found here: https://github.com/nextcloud/tables/wiki",
    3,
    "false"
  ]
]
```

### Expected behavior

The spec correctly mentions that the endpoint returns a nested list.

### Actual behavior

The spec differs from the actual behavior, making generated api clients fail.

### Tables app version

0.8.0

### Browser

N.A.

### Client operating system

N.A.

### Operating system

Linux (Docker

### Web server

None

### PHP engine version

PHP 8.2

### Database

SQLite

### Additional info

I also had type issues on other endpoints but didn't debug all of them for now.
I suspect that this issue is broader than just the one endpoint.

PR #1386, opened 2024-09-27
fix(OpenAPI): Fix some typings

provokateurin:
Fixes https://github.com/nextcloud/tables/issues/1367

Tried to fix the most problematic things, but there are quite a few more problems that should be fixed.
Fixing everything right now is just too much work.


blizzz:
This breaks integration ans possible frontend tests.

provokateurin:
I know, I will try to fix them. Given the code quality it is hard to judge what parts I can safely refactor and fix and which I should stay away from.

provokateurin:
Yes, although you should eventually upgrade to the newer openapi-extractor versions which require these fixes to be done.

PR #1397, opened 2024-09-29
fix: use unique names for actions upload-artifact

enjeck:
At https://github.com/nextcloud/tables/actions/runs/11089354255?pr=1396, we have the error `Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run`. This is an attempt to fix it.

### DIFF
--- a/.github/workflows/cypress-custom.yml
+++ b/.github/workflows/cypress-custom.yml
@@ -149,24 +149,24 @@ jobs:
           CYPRESS_ncVersion: ${{ matrix.server-versions }}
           npm_package_name: ${{ env.APP_NAME }}
 
-      - name: Print logs
+      - name: Print logs ${{ matrix.node-version }}-${{ matrix.php-versions }}-${{ matrix.server-versions }}-${{ matrix.databases }}
         if: always()
         run: |
           cat /tmp/requestlog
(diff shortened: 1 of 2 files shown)


blizzz:
Seems related: https://github.com/actions/upload-artifact/issues/480

v3 will be expired on Nov 30.

Better way might be as suggested here: https://github.com/actions/upload-artifact/issues/481#issuecomment-1862899981

.github/workflows/cypress-custom.yml
```diff
@@ -136,7 +136,7 @@ jobs:
           cat data/nextcloud.log
 
       - name: Cypress run
-        uses: cypress-io/github-action@v4
+        uses: cypress-io/github-action@v3
```
blizzz, line 139: this is perhaps accidental?

blizzz:
The failing tests fail for me locally as well

blizzz:
there is nothing with a `.paragraph-content`

Screenshot_20241008_125331

blizzz:
Okay, I suppose that was removed in Text (as it is in text wrapper): https://github.com/nextcloud/text/pull/6425

blizzz (approved):

Issue #638, opened 2023-10-31
Adding a row to a filtered view does not set the filtered values


juliusknorr:
- Have a filtered view for a select value
- Create a new row
- Reload the page
- The new row has been added but the value the view is filtered to is not set

Can be seen in our feature tracking table with this filter:
![Screenshot 2023-10-31 at 16 18 14](https://github.com/nextcloud/tables/assets/3404133/b79188d8-12d5-4c27-96d7-fabc204f7171)

Row after adding:
![Screenshot 2023-10-31 at 16 18 29](https://github.com/nextcloud/tables/assets/3404133/f98d92a8-d0b5-4c18-9823-6ca26c689215)

datenangebot:
If I understand that correct, you want to set the filtered value additionally to the new row data, so the match the actual filter.

I would say this is not a bug, more a feature request, because we don't implemented this yet.

But, filter may be more complex than "team=office" and then it's might not be clear which value should be set. Think about filter like "date > today" or "Task contains abc"...

If we provide those fields in addition so they can be set if they are not that clear, that could be a data breach if something gets filtered in the back back secret values?! 😬

juliushaertl:
Agree that this is more tricky than just picking up the filter like for range/contains filters we would not be able to decide automatically on a default. Though I was wondering if for views that have simple equal filters we could still apply those values as defaults, but even figuring that out might be hard. 

> If we provide those fields in addition so they can be set if they are not that clear, that could be a data breach if something gets filtered in the back back secret values?! 😬

True, we should not let users pick if the column is not exposed to the view of course.

Do we have an insight how other solutions handle this? 

A first idea could be to have a dedicated option to pick default values for new entries per view.

datenangebot:
> Do we have an insight how other solutions handle this?

No. Needs research.
In Drupal it is possible to add values to load them for calculation but without presenting in the front end. But Drupal is an expert tool with a different approach...

> A first idea could be to have a dedicated option to pick default values for new entries per view.

Yes, that sounds like a quick solution for that. But that would raise the complexity a lot and users might not know what that means?!
Another idea could be a feature to hide columns in views. It would mean you can edit or create all data including hidden columns, but you don't see them in the table. That would fix the mentioned UC but not as a general solution?

pmk1c:
I think this should be a validation, when adding a new row to a view. At least that was the behaviour I was expecting:

When adding a row to a view with filters, the added row must match the filters of the view, or it can't be created.

[...]

PR #1409, opened 2024-10-08
fix: Properly create new rows in filtered views

juliusknorr:
First attempt to improve #638 by applying exact view filters as a default value for new rows

- [x] Apply filter values as defaults where feasible
- [x] Have some tests

## Follow up
- [ ] Warn during view creation if a new view leads to issues (e.g. a column is hidden but no defaults are present or can be assumed)

### DIFF
--- a/tests/integration/features/bootstrap/FeatureContext.php
+++ b/tests/integration/features/bootstrap/FeatureContext.php
@@ -595,14 +595,6 @@ class FeatureContext implements Context {
 		}
 	}
 
-
-
-
-
-
-
-
-
 	/**
 	 * @Then user :user has the following tables
 	 *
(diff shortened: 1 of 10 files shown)


juliusknorr:
This is good from my side as a first iteration.

Tried to come up with some test cases as well.

blizzz (approved):
Looks good, great with tests!

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -902,6 +894,10 @@ public function applyColumnsToView(string $user, string $columnList, string $vie
 		}, $columns);
 
 		$this->sendUpdateViewRequest($viewAlias, ['columns' => json_encode($columns)]);
+
+		$view['columns'] = $columns;
```
blizzz, line 898: indeed there is no `id` field assigned to it

blizzz:
failing test is unrelated and fixed with https://github.com/nextcloud/tables/pull/1612

PR #1434, opened 2024-10-29
enh: warn during table manager promotion/demotion

enjeck:


juliusknorr (approved):
Looks good and wording also fits from my perspective 👍

marcoambrosini (commented):
I think it would be nice to have a "cancel" button too?
And is this a destructive action? We usually reserve the color error for such actions

marcoambrosini:
@enjeck looks good! let's just right align the cancel button like in other dialogs and we're golden.

Issue #1440, opened 2024-11-05
Date (re-)import shenanigans


blizzz:
### Steps to reproduce

1. Create a table with a `Datetime` field
2. Export the table to csv
3. Create a new table by importing the file. Make sure to set the column type to DateTime during import.

### Expected behavior

New table content is like old table content

### Actual behavior

The import fails, because the date cannot be parsed, for it is in an unexpected format:

```
floor(): Argument #1 ($num) must be of type int|float, string given","Code":0,"Trace":[{"file":"/srv/http/nextcloud/master/apps-repos/tables/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Shared/Date.php","line":224
```

In one place we run `\PhpOffice\PhpSpreadsheet\Shared\Date::excelToDateTimeObject()` without testing the input value. The method assumes it is a float or int. More precisely it expects a Microsoft Excel timestamp like `42495.15902`, but a string in the form of "2024-02-24".

Also I am not sure the detection works, either, for `Date::isDateTime()` is being called and it  does not test for the mentioned timestamp. 

### Tables app version

main

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

Actually, I am not sure how important it is to check for this type. It might be with xlsx formats, I have not tried it out yet. At some point, there migth be collisions between unix timestamps and MS Excel timestamps and it is not possible to tell them apart – though the file type can be taken into account. 

For example,  `42495` is roughly quarter to 1pm on 01.01.1970 in UNIX, but 0.00 o'clock on May 5th 2016 in Excel.

I have local changes trying to tame this to some degree, but not every case is covered yet.

cc @luka-nextcloud if you have insights.

blizzz:
Related to #971

blizzz:
For the record, in xlsx files, the date is stored in the excel format, and import is possible from there.

PR #1463, opened 2024-11-16
enh: display error popup when import preview fails

enjeck:
Right now, at https://github.com/nextcloud/tables/issues/1440, when the import fails, there's no error visible to the user. So, this adds an error popup so the user knows that there was a problem

### DIFF
--- a/src/modules/modals/Import.vue
+++ b/src/modules/modals/Import.vue
@@ -94,11 +94,17 @@
 
 			<!-- show loading -->
 			<div v-if="loading && !waitForReload">
-				<NcEmptyContent :name="t('tables', 'Importing data from ') + importFileName" :description="t('tables', 'This might take a while...')">
-					<template #icon>
-						<NcIconTimerSand />
-					</template>
-				</NcEmptyContent>
+				<div v-if="!importFailed">
+					<NcEmptyContent :name="t('tables', 'Importing data from ') + importFileName"
+						:description="t('tables', 'This might take a while...')">
+						<template #icon>
+							<NcIconTimerSand />
+						</template>
+					</NcEmptyContent>
+				</div>
+				<div v-else>
+					<NcEmptyContent :name="t('tables', 'Failed')" :description="t('tables', 'Unable to import data')" />
+				</div>
 			</div>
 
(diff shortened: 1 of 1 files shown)


blizzz:
Definitely an improvement!

Maybe we still can do update this screen as well? 

Screenshot_20241118_111641

Cause it keep idling and the notification on the top right may escape the eye.

juliusknorr:
Agreed, let's switch to a proper error screen

src/modules/modals/Import.vue
```diff
@@ -281,21 +288,26 @@ export default {
...
+					} else if (res.status === 403) {
+						console.debug('error while importing', res)
+						showError(t('tables', 'Could not import, missing needed permission.'))
+					} else if (res.status === 404) {
+						console.debug('error while importing', res)
+						showError(t('tables', 'Could not import, needed resources were not found.'))
+					} else {
+						showError(t('tables', 'Could not import data due to unknown errors.'))
+						console.debug('error while importing', res)
+					}
```
juliusknorr, line 304: We could actually show the error message as the empty content description instead of another showError toast message.

src/modules/modals/Import.vue
```diff
@@ -281,21 +288,26 @@ export default {
...
+					this.importFailed = true
 				}
 			} catch (e) {
+				showError(t('tables', 'Could not import data due to unknown errors.'))
```
juliusknorr, line 308: Same for this and the ones below

src/modules/modals/Import.vue
```diff
@@ -315,21 +327,26 @@ export default {
...
-				} else if (res.status === 403) {
-					console.debug('error while importing', res)
-					showError(t('tables', 'Could not import, missing needed permission.'))
-				} else if (res.status === 404) {
-					console.debug('error while importing', res)
-					showError(t('tables', 'Could not import, needed resources were not found.'))
 				} else {
-					showError(t('tables', 'Could not import data due to unknown errors.'))
-					console.debug('error while importing', res)
+					if (res.status === 401) {
```
juliusknorr, line 331: Seems like some code duplication, maybe we can move the error handling into a shared method?

PR #1465, opened 2024-11-18
share with teams

### LINKED ISSUE #142: Share with Teams
An important and very helpful function is sharing with circles (which is now called teams).
https://apps.nextcloud.com/apps/circles

The backend should already be kind of generic so that the existing database schema can be reused and we can introduce a different type for existing shares (where sharing to users and groups is already possible).

This will need a backend implementation to support the share types when creating/updating/listing shares and permission checks to be extended to validate access depending on a users team membership. Frontend would need to be extended to support rendering share list entries for teams as well as listing teams to share with in the autocomplete dialog.

- Jonas can probably help with the Teams/Circles details
- Cleopatra for the frontend bits

hweihwang:
Fixes #142

### DIFF
--- a/lib/Helper/CircleHelper.php
+++ b/lib/Helper/CircleHelper.php
@@ -1,0 +1,85 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Tables\Helper;
+
+use OCA\Circles\CirclesManager;
+use OCA\Circles\Model\Member;
+use OCA\Circles\Model\Probes\CircleProbe;
+use OCP\App\IAppManager;
+use OCP\Server;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+/**
+ * @psalm-suppress UndefinedClass
+ */
+class CircleHelper {
+	private bool $circlesEnabled;
+	private ?CirclesManager $circlesManager;
+
+	/**
+	 * @psalm-suppress UndefinedClass
+	 */
+	public function __construct(
+		private LoggerInterface $logger,
+		IAppManager     $appManager
+	) {
+		$this->circlesEnabled = $appManager->isEnabledForUser('circles');
+		$this->circlesManager = null;
+
+		if ($this->circlesEnabled) {
+			try {
+				$this->circlesManager = Server::get(CirclesManager::class);
+			} catch (Throwable $e) {
+				$this->logger->warning('Failed to get CirclesManager: ' . $e->getMessage());
+				$this->circlesEnabled = false;
+			}
+		}
+	}
+
+	public function isCirclesEnabled(): bool {
+		return $this->circlesEnabled;
+	}
+
+	public function getCircleDisplayName(string $circleId, string $userId): string {
+		if (!$this->circlesEnabled) {
+			return $circleId;
+		}
+
+		try {
+			$federatedUser = $this->circlesManager->getFederatedUser($userId, Member::TYPE_USER);
+			$this->circlesManager->startSession($federatedUser);
+
+			$circle = $this->circlesManager->getCircle($circleId);
+			return $circle ? ($circle->getDisplayName() ?: $circleId) : $circleId;
+		} catch (Throwable $e) {
+			$this->logger->warning('Failed to get circle display name: ' . $e->getMessage(), [
+				'circleId' => $circleId,
+				'userId' => $userId
+			]);
+			return $circleId;
+		}
+	}
+
+	public function getUserCircles(string $userId): array {
+		if (!$this->circlesEnabled) {
+			return [];
+		}
+
+		try {
+			$federatedUser = $this->circlesManager->getFederatedUser($userId, Member::TYPE_USER);
+			$this->circlesManager->startSession($federatedUser);
+			$probe = new CircleProbe();
+			$probe->mustBeMember();
+			return $this->circlesManager->getCircles($probe);
+		} catch (Throwable $e) {
+			$this->logger->warning('Failed to get user circles: ' . $e->getMessage());
+			return [];
+		}
+	}
+}
--- a/lib/Helper/UserHelper.php
+++ b/lib/Helper/UserHelper.php
@@ -18,6 +18,7 @@ class UserHelper {
 	private IUserManager $userManager;
 
 	private LoggerInterface $logger;
+
 	private IGroupManager $groupManager;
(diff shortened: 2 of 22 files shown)


lib/Helper/UserHelper.php
```diff
@@ -56,6 +72,17 @@ public function getGroupsForUser(string $userId): array {
 		return $this->groupManager->getUserGroups($user);
 	}
 
+	/**
+	 * @psalm-suppress UndefinedDocblockClass
+	 * @return Circle[]
+	 * @throws InitiatorNotFoundException
+	 * @throws RequestBuilderException
+	 */
+	public function getCircles(): array {
```
blizzz, line 81: could be moved to the CircleHelper?

lib/Service/ShareService.php
```diff
@@ -72,6 +77,7 @@ public function __construct(
 		$this->viewMapper = $viewMapper;
 		$this->userHelper = $userHelper;
 		$this->groupHelper = $groupHelper;
+		$this->circleHelper = $circleHelper;
```
blizzz, line 80: For `CircleHelper` has direct dependencies on `\OCA\Circles` this will probably crash when `circles` is disabled. Perhaps also the reason why integration tests are crashing. 

Either it has to be loaded only when it is ensured that circles is loaded (use `IAppManager`) or change `CircleHelper` accordingly.

blizzz (changes requested):
I looked at the backend bits only so far.

hweihwang:
Thanks Arthur @blizzz 🙏 - those are really helpful!

juliusknorr:
> I think we can hide those shares when circles disabled

Yes, would agree with that. I think we do it the same way in the files app 👍

hweihwang:
Yes @enjeck, we check from BE so when getting all tables of a user will exclude those circle's shares

lib/Helper/CircleHelper.php
```diff
@@ -0,0 +1,87 @@
...
+
+		if ($this->circlesEnabled) {
+			try {
+				$this->circlesManager = Server::get(CirclesManager::class);
+			} catch (Throwable $e) {
+				$this->logger->warning('Failed to get CirclesManager: ' . $e->getMessage());
+				$this->circlesEnabled = false;
+			}
+		}
+	}
```
blizzz, line 45:
```suggestion
	private bool $circlesEnabled;
	private ?CirclesManager $circlesManager;

	/**
	 * @psalm-suppress UndefinedClass
	 */
	public function __construct(
		private LoggerInterface $logger,
		IAppManager $appManager
	) {
		$this->circlesEnabled = $appManager->isEnabledForUser('circles');
		$this->circlesManager = null;

		if ($this->circlesEnabled) {
			try {
				$this->circlesManager = Server::get(CirclesManager::class);
			} catch (Throwable $e) {
				$this->logger->warning('Failed to get CirclesManager: ' . $e->getMessage());
				$this->circlesEnabled = false;
			}
		}
	}
```

nitpick: use property promotion

lib/Helper/UserHelper.php
```diff
@@ -14,17 +14,26 @@
 use OCP\IUserManager;
 use Psr\Log\LoggerInterface;
 
+/**
+ * @psalm-suppress UndefinedClass
+ * @psalm-suppress UndefinedDocblockClass
```
blizzz, line 19: What are they about? There should be nothing extra ordinary here?

lib/Helper/UserHelper.php
```diff
@@ -14,17 +14,26 @@
...
 class UserHelper {
 	private IUserManager $userManager;
 
 	private LoggerInterface $logger;
+
 	private IGroupManager $groupManager;
 
+	/**
+	 * @psalm-suppress UndefinedClass
+	 */
```
blizzz, line 30: same question

lib/Service/PermissionsService.php
```diff
@@ -18,14 +18,19 @@
...
 use OCA\Tables\Model\Permissions;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\MultipleObjectsReturnedException;
 use OCP\DB\Exception;
 use Psr\Log\LoggerInterface;
+use Throwable;
 
+/**
+ * @psalm-suppress UndefinedDocblockClass
+ */
```
blizzz, line 33: What is psalm complaining about here?

lib/Service/ShareService.php
```diff
@@ -35,9 +35,11 @@
 use OCP\DB\Exception;
 use OCP\IDBConnection;
 use Psr\Log\LoggerInterface;
+use Throwable;
 
 /**
  * @psalm-import-type TablesShare from ResponseDefinitions
+ * @psalm-suppress UndefinedDocblockClass
```
blizzz, line 42: also wondering here

lib/Service/ShareService.php
```diff
@@ -85,7 +93,9 @@ public function findAll(string $nodeType, int $nodeId, ?string $userId = null, b
 		$userId = $this->permissionsService->preCheckUserId($userId);
 
 		try {
-			$shares = $this->mapper->findAllSharesForNode($nodeType, $nodeId, $userId);
+			$excluded = !$this->circleHelper->isCirclesEnabled() ? ['circle'] : [];
```
blizzz, line 96: Another nitpick: ideal turn `'cirlce'` into a constant (and adapt follow up usages).

(Yes, existing `group` string suffers from the same)

blizzz (commented):
Backendwise looks OK (only nitpicks, not blocking), did not do functional tests so far.

src/shared/mixins/searchUserGroup.js
```diff
@@ -44,23 +44,24 @@ export default {
...
+			const typeLabels = {
+				[this.SHARE_TYPES.SHARE_TYPE_USER]: 'User',
+				[this.SHARE_TYPES.SHARE_TYPE_GROUP]: 'Group',
+				[this.SHARE_TYPES.SHARE_TYPE_CIRCLE]: 'Circle',
 			}
+
+			const selectedLabels = shareTypes.map(type => typeLabels[type])
+			return selectedLabels.length > 0
+				? selectedLabels.join(' or ')
+				: 'User, group or circle'
```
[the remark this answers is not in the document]
blizzz: Circle was used befrore 30(?), ideally it depends on it to avoid confusion.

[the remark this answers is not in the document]
juliusknorr: Team was introduced with 29 as a term, so I'd say it is fine to just use that one.

juliusknorr (approved):
Tested and works

PR #1466, opened 2024-11-18
build(dev-deps): update phpunit for PHP 8.4 compat

blizzz:
latest phpunit has some commits regarding PHP 8.4, let's see whether tests will pass.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -762,16 +762,16 @@
         },
         {
             "name": "myclabs/deep-copy",
-            "version": "1.11.1",
+            "version": "1.12.1",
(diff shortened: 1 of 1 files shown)


blizzz:
Sheesh, it kills openapi :scream_cat:

blizzz:
> Sheesh, it kills openapi 🙀

OK, openapi-extractor has been bumped as well, and it was also problematic on server, cf. https://github.com/nextcloud/server/pull/49333

trying to get rid of it

juliusknorr (approved):

PR #1476, opened 2024-11-25
Fix "start of week" filter not working on table views

burdoto:


burdoto:
Hi, I have noticed that upon updating to Nextcloud 30.0.4, Tables was updated to version 0.8.2; but shipped without this fix.

PR #1477, opened 2024-11-27
fix: Update menu switch CSS selectors

enjeck:
Once again, the CSS styling changed and we have to update this. Otherwise, we have the error message in the console: `"TypeError: Cannot read properties of null (reading 'classList')"`

Would be nice to someday have a better solution for this, since we've had to update these styles several times already

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -82,20 +82,20 @@ export default {
 				this.$store.commit('setActiveTableId', parseInt(currentRoute.params.tableId))
 				this.setPageTitle(this.$store.getters.activeTable.title)
 				if (!currentRoute.path.includes('/row/')) {
-					this.switchActiveMenuEntry(document.querySelector('header .header-left .app-menu a[title="Tables"]'))
+					this.switchActiveMenuEntry(document.querySelector('header .header-start .app-menu a[title="Tables"]'))
 				}
 			} else if (currentRoute.path.startsWith('/view/')) {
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -82,20 +82,20 @@ export default {
 				this.$store.commit('setActiveTableId', parseInt(currentRoute.params.tableId))
 				this.setPageTitle(this.$store.getters.activeTable.title)
 				if (!currentRoute.path.includes('/row/')) {
-					this.switchActiveMenuEntry(document.querySelector('header .header-left .app-menu a[title="Tables"]'))
+					this.switchActiveMenuEntry(document.querySelector('header .header-start .app-menu a[title="Tables"]'))
```
juliusknorr, line 85: I think this change also only happened on master/stable30 not older versions so we might need a fallback.

juliusknorr:
> Would be nice to someday have a better solution for this, since we've had to update these styles several times already

Yes, definitely a good idea. The app menu in server has some API to update or set a counter already, maybe we can think about extending that to set the active entry:

https://github.com/nextcloud/server/blob/c5baf3d7ef2715def79ff7734d505c56bb20012c/core/src/components/MainMenu.js#L29
https://github.com/nextcloud/server/blob/21c30e52591f6e73f28bf7dc5d9584f609f911f7/core/src/components/AppMenu.vue#L89-L95

Issue #1524, opened 2025-01-06
Shared View is empty when using sort rule


ISC-PRO:
### Steps to reproduce

1. Create a simple list "HowTo" with only column "ToDo", Type "Text"
2 .Create a view "Items" of table "HowTo" with default, selected columns
3. Create two items one minute apart: "ToDo1" AND "ToDo2"
4. Change the view "Items" and add sorting with rule "Erstellt am" by DESC -> Result First: ToDo2, Second: ToDo1
5. Share the view with person or group
-> RESULT: view "Items" **is empty** for all shared users

6. Delete (as owner) the sort rule for view "Items"
-> RESULT: view "Items" is visible to all shared users

### Expected behavior

Show all items with sorting (DESC) for all shared users.
For the own user the view correct.

### Actual behavior

The view is empty for all shared users.

### Tables app version

0.8.1

### Browser

Edge 131.0.2903.99

### Operating system

Nextcloud Hub 9 (30.0.2)

### Web server

None

### PHP engine version

None

### Database

None

blizzz:
And indeed the server replies with sort data as null 🤔 

```
[
  {
    "id": 175,
    "tableId": 881,
    "title": "Items",
    "description": "",
    "emoji": "😇",
    "ownership": "User1",
    "createdBy": "User1",
    "createdAt": "2025-01-06 13:27:28",
    "lastEditBy": "User1",
    "lastEditAt": "2025-01-06 13:28:57",
    "columns": [
      2920
    ],
    "sort": [
      null
    ],
    "isShared": true,
    "favorite": false,
    "onSharePermissions": {
      "read": true,
      "create": true,
      "update": true,
      "delete": false,
      "manage": false
    },
    "hasShares": false,
    "rowsCount": 3,
    "ownerDisplayName": "User1",
    "filter": []
  }
]
```

So looks like a backend thing.

ISC-PRO:
Thanks for the quick reply.

ISC-PRO:
The problem is solved. We use version 0.8.4.

PR #1525, opened 2025-01-06
fix(Views): allow to sort for meta columns

blizzz:
- meta columns included in sorting do not need to be hidden from the response
- changed the return value to an empty array, to avoid a Vue error in the frontend (it expects an array)

Now however, for owners the sorting is working ~~,but for share recipients it is not (but the content is at least visible).~~

~~Not sure this is the best approach, and the sorting needs frontend changes as well. Hoping for your feedback @enjeck~~

addresses #1524

### DIFF
--- a/lib/Service/ViewService.php
+++ b/lib/Service/ViewService.php
@@ -449,13 +449,18 @@ class ViewService extends SuperService {
 			$rawSortArray = $view->getSortArray();
 			if ($rawSortArray) {
 				$view->setSortArray(
-					array_map(static function ($sortRule) use ($view) {
+					array_map(static function (array $sortRule) use ($view): array {
 						$columnsArray = $view->getColumnsArray();
-						if (isset($sortRule['columnId']) && $columnsArray && in_array($sortRule['columnId'], $columnsArray, true)) {
+						if (isset($sortRule['columnId'])
+							&& (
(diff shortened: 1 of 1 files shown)


lib/Service/ViewService.php
```diff
@@ -449,13 +449,17 @@ private function enhanceView(View $view, string $userId): void {
 			$rawSortArray = $view->getSortArray();
 			if ($rawSortArray) {
 				$view->setSortArray(
-					array_map(static function ($sortRule) use ($view) {
+					array_map(static function ($sortRule) use ($view): array {
 						$columnsArray = $view->getColumnsArray();
-						if (isset($sortRule['columnId']) && $columnsArray && in_array($sortRule['columnId'], $columnsArray, true)) {
+						if (isset($sortRule['columnId'])
+							&& $sortRule['columnId'] >= 0 // meta columns can be exposed
```
blizzz, line 455: I have a better test in a different commit and would adjust once that's in.

blizzz, line 455: I would adapt to negated `isValidMetaTypeId()` check from https://github.com/nextcloud/tables/pull/1526 once that's in.

blizzz:
Rebased, adjusted, and fixed a logic i had in, and sorting for share recipients works now.

juliusknorr (approved):

PR #1540, opened 2025-01-13
feat: migrate from vuex to pinia

### LINKED ISSUE #1492: Migrate from vuex to pinia

enjeck:
fixes #1492

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -3,24 +3,25 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 import { generateFilePath } from '@nextcloud/router'
+import { createPinia, PiniaVuePlugin } from 'pinia'
 import Vue from 'vue'
 import App from './App.vue'
-import Vuex from 'vuex'
-import store from './store/store.js'
 import router from './router.js'
 import VuePapaParse from 'vue-papa-parse'
+import { translate as t, translatePlural as n } from '@nextcloud/l10n'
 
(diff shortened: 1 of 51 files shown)


src/main.js
```diff
@@ -15,12 +14,11 @@ __webpack_nonce__ = btoa(OC.requestToken) // eslint-disable-line
 __webpack_public_path__ = generateFilePath('tables', '', 'js/')
 
 Vue.mixin({ methods: { t, n } })
-Vue.use(Vuex)
+Vue.use(createPinia())
```
nfebe, line 17:
```suggestion
```

src/main.js
```diff
@@ -15,12 +14,11 @@ __webpack_nonce__ = btoa(OC.requestToken) // eslint-disable-line
...
 Vue.use(VuePapaParse)
 
 export default new Vue({
 	el: '#content',
 	router,
```
nfebe, line 22:
```suggestion

const pinia = createPinia()

export default new Vue({
	el: '#content',
	router,
	pinia,
```


Plus I think if this is vue 2 you need `Vue.use(PiniaVuePlugin)`

nfebe (commented):
:eyes:

juliusknorr (approved):

juliusknorr:
While testing i found some frequent console error when navigating between tables, but something separate as it also happens on main, but should be fixed. Is that known/tracked somwhere?

Screenshot 2025-01-29 at 07 06 40

PR #1545, opened 2025-01-16
fix(Import): DateTime import fixes

### LINKED ISSUE #1518: Import CSV file from another Nextcloud Tables instance not working
### Steps to reproduce

1. On server 1 with Nextcloud+Tables, we go to the Data of one table, and select Export to CSV file
2. On server 2 with same version of Nextcloud+Tables, create a new Table, Import a table from file, and inject the CSV file get from Server 1
3. The process start with the mention "This could take a long time", but after a while, no data are imported

### Expected behavior

The data should be imported

### Actual behavior

The data are not imported

### Tables app version

0.8.2

### Browser

Brave Version 1.73.104 Chromium: 131.0.6778.204 (Build officiel) (64 bits)

### Client operating system

Linux

### Operating system

debian

### Web server

Apache

### PHP engine version

PHP 8.2

### Database

MySQL

### Additional info

See the CSV file as attachment
[contacts.csv](https://github.com/user-attachments/files/18276717/contacts.csv)

blizzz:
- do not try to pass `null` to DateTimeImmutable as it would cause an Exception
- be stricter on small letter-only values that might be turned into a DateTimeImmutable object because they match with timezone abbreviations
- do not consider pure numbers a date

fixes #1518

### DIFF
--- a/lib/Service/ImportService.php
+++ b/lib/Service/ImportService.php
@@ -31,6 +31,9 @@ use PhpOffice\PhpSpreadsheet\Worksheet\Worksheet;
 use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\NotFoundExceptionInterface;
 use Psr\Log\LoggerInterface;
+use function is_string;
+use function mb_strlen;
+use function preg_match;
 
 class ImportService extends SuperService {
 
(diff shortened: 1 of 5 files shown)


tests/integration/features/APIv1.feature
```diff
@@ -218,6 +218,32 @@ Feature: APIv1
     | import-from-ms365.xlsx       |
     | import-from-libreoffice.csv  |
 
+  @api1 @import @current
```
blizzz, line 221: gotta remove `@current` again

lib/Service/ImportService.php
```diff
@@ -382,30 +380,14 @@ private function createRow(Row $row): void {
 				$hasData = $hasData || !empty($value);
 
 				if ($column->getType() === 'datetime') {
-					if ($column->getType() === 'datetime' && $column->getSubtype() === 'date') {
+					if ($column->getSubtype() === 'date') {
```
juliusknorr, line 383: Could we also add constants for the subtypes like we have for types in https://github.com/nextcloud/tables/blob/c3c97cfbd963edb05ce245241c11f25fc76cdd83/lib/Db/Column.php#L95 ?

I always struggle finding the ones that are used, but may be unrelated

juliusknorr (approved):

lib/Service/ImportService.php
blizzz: added and adjusted this whole service file.

Issue #974, opened 2024-04-12
Display full text in cell easily


enjeck:
If the text in the cell is longer than column can display, it seems you can't see the full content unless you open the Edit window or export the whole table. I think this is not straightforward and not logical when you just need to view the content.

Quick ideas:
- the ability to manually adjust the column width for your view, or even simpler - a tip with full content when you hover the mouse over it.
- Another idea could be wrapping the text and have variable height rows

![1024-689](https://github.com/nextcloud/tables/assets/32180937/b4ea60a5-6cb2-47ef-aee6-85bd6ba82487)

viktorix:
To add another idea, ability to expand the row on click would be nice. Keep rows the same height, and if user needs to see more data hidden in that row, they click on a plus button or chevron down button to expand the row down. Typical expand/collapse feature.

enjeck:
Related issue about text-wrapping: https://github.com/nextcloud/tables/issues/619

GoOz:
I second @viktorix's idea. Looks a bit easier to implement and as effective as the other ideas.

pat-githubber:
I wish, I had so many characters visible:
Nextcloud Hub 8 (29.0.5)
Tables Version 0.8.1
Firefox 131.0.2 (64-bit)

![Image](https://github.com/user-attachments/assets/6cea39b8-54a7-4f91-b396-0c65452d1117)

PR #1551, opened 2025-01-20
fix: wrap usergroup's cell values

enjeck:
We need to wrap the container of usergroup items. Otherwise, they overflow the container when there are too many items. 

### Before
Screenshot from 2025-01-20 18-11-57

### After
Screenshot from 2025-01-20 18-18-56


blizzz (approved):

Ornanovitch:
Nice :)

But why not using a global rule in order to harmonize this behavior across the app and fix also #974?

And what if you have 100 users in the cell? A "click to see more" would be ideal.

PR #1560, opened 2025-01-28
ci(config): stop updates for 0.7, enable PHP updates in renovate

blizzz:
- removes stable0.7 from npm-audit, the series is EOL
- adds PHP dependency updates to renovate
- adds workflow updates to renovate
- adds stable0.8 as target branch to renovate
- removes stable0.7 (EOL) as target branch to renovate

I am not familiar much with renovate config and took looks at the configs at some other apps :crossed_fingers:

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,23 +1,29 @@
 {
 	"$schema": "https://docs.renovatebot.com/renovate-schema.json",
 	"extends": [
-		"config:base"
+		"config:base",
+		":semanticCommits",
+		":gitSignOff"
 	],
 	"timezone": "Europe/Berlin",
 	"schedule": [
 		"every weekend"
 	],
-	"reviewers": ["datenangebot", "enjeck"],
-	"labels": ["dependencies"],
+	"labels": [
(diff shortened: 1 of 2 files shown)


blizzz:
Cypress is a false-negative,  the change does not affect any code logic.

renovate.json
```diff
@@ -1,33 +1,57 @@
...
 	"ignoreDeps": [
 		"node",
 		"npm"
 	],
 	"packageRules": [
 		{
-			"matchUpdateTypes": ["major", "minor"],
-			"matchBaseBranches": ["stable0.7"],
-			"enabled": false
+            "description": "Request JavaScript reviews",
```
juliusknorr, line 34: Tabs instead of spaces?

juliusknorr (approved):
Nitpick on indentation, otherwise 👍

renovate.json
blizzz: Ough, that was not caught by cs-fixer :crying_cat_face:  Good catch!

Committed the style fixes now!

PR #1565, opened 2025-01-28
feat(Api): add v2 OCS Api to get table/view rows

### LINKED ISSUE #1507: Row fetch endpoint that also accepts filters and sort rules
Contributes to #941

For pagination we need to have a endpoint, that fetches the rows of a table or view also in consideration of user supplied filter and sorting rules.

The current controller methods of `\OCA\Tables\Controller\Api1Controller` (`indexTableRows` and `indexViewRows`) accept parameters for pagination, but that is it.

In this context it makes sense to integrate the approaches of https://github.com/nextcloud/tables/pull/927 and create a new v2 endpoint. 

The current v1 endpoint shall be deprecated, as it is already not used by the web frontend. It should not be removed (the android app might make use of it – needs to be verified).

The internal endpoints (`\OCA\Tables\Controller\RowController::index()` and `::indexView()`)  shall be deprecated, and later on removed as soon as the web interfaces also switches to the v2 endpoint.

blizzz:
What is left is now basically tests. And removal of old internal method. And deprecation of v1 endpoint if any. Also TODOs in the code, there was some prototyping.

Intended test matrix:

[test matrix.ods](https://github.com/user-attachments/files/19059350/test.matrix.ods)


enjeck (commented):
I'll take over this

lib/Analytics/AnalyticsDatasource.php
```diff
@@ -338,7 +338,7 @@ private function formatBooleanValue(mixed $value): string {
 		return '';
 	}
 
-	private function formatTextValue(Column $column, mixed $value): string {
+	private function formatTextValue(Column $column, string $value): string {
```
blizzz, line 325: this isn't necessarily a string that is coming in… or is it?

lib/Controller/RowOCSController.php
```diff
@@ -136,77 +137,103 @@ public function createRow(string $nodeCollection, int $nodeId, mixed $data): Dat
...
+	public function getRows(string $nodeCollection, int $nodeId, ?int $limit = null, ?int $offset = null, ?string $filter = null, ?string $sort = null): DataResponse {
 		try {
 			if (($limit !== null && ($limit <= 0 || $limit > 500))
 				|| ($offset !== null && $offset < 0)
 			) {
-				// TODO: this check can be removed once NC 32 is the lowest supported server versions,
-				// as then the app framework handles nullable ranges
-				throw new \InvalidArgumentException('Offset or limit parameter is out of bounds');
+				throw new InvalidArgumentException('Offset or limit parameter is out of bounds');
 			}
```
blizzz: this check can be removed nowadays (see old comment) as we already specify the range in the annotation ( `@psalm-param ?int<1,500> $limit`)

lib/Controller/RowOCSController.php
```diff
@@ -136,77 +137,103 @@ public function createRow(string $nodeCollection, int $nodeId, mixed $data): Dat
...
+			);
 			$queryData->setLimit($limit)
 				->setOffset($offset)
-				// we set the provided filter here, any existing filter
-				// definitions (if specified on views) are applied on service level
-				->setFilter($filter)
-				->setSort($sort)
+				// the provided filter is set here; any filter defined on a view
+				// is merged in on the service level
+				->setFilter($this->parseFilter($filter))
```
blizzz, line 175: why not use `FilterSet::createFromInputArray()`?

lib/Controller/RowOCSController.php
```diff
@@ -136,77 +137,103 @@ public function createRow(string $nodeCollection, int $nodeId, mixed $data): Dat
...
+				->setSort($this->parseSort($sort))
```
blizzz, line 176: and `SortRuleSet::createFromInputArray()` here

blizzz (commented):
And there are more static analysis complaints

enjeck:
> 1. Implement an [`If-Modified-Since`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-Modified-Since) header based on the `lastModifiedAt` property on each row to exclude "old", already known rows to reduce traffic. This covers newly created and edited rows on the server since the last synchronization

@stefan-niedermann  Should be done at https://github.com/nextcloud/tables/pull/1844 ?

Koc:
hey @blizzz @enjeck !

This PR looks very promising. I would like to move all filters/soring/paginaton to BE side, see #2855 and this PR perfectly fits for it.

Do you plan to merge it more or less soon?

Also we can switch reading to v2 and implement one point for #2727

PR #1620, opened 2025-02-17
fix: Export empty view array when exporting shared tables

enjeck:
When a table is shared, the export code can't retrieve the views. So we return an empty array. Otherwise, we get the error:

```
  "message": "OCA\\Tables\\Model\\TableScheme::__construct(): Argument #4 ($view) must be of type array, null given, called in /var/www/cloud.nextcloud.com/nextcloud/apps/tables/lib/Service/TableService.php on line 531 in file '/var/www/cloud.nextcloud.com/nextcloud/apps/tables/lib/Model/TableScheme.php' line 27",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64; rv:134.0) Gecko/20100101 Firefox/134.0",
  "version": "31.0.0.15",
  "exception": {
    "Exception": "Exception",
    "Message": "OCA\\Tables\\Model\\TableScheme::__construct(): Argument #4 ($view) must be of type array, null given, called in /var/www/cloud.nextcloud.com/nextcloud/apps/tables/lib/Service/TableService.php on line 531 in file '/var/www/cloud.nextcloud.com/nextcloud/apps/tables/lib/Model/TableScheme.php' line 27",
    "Code": 0,
    "Trace": [
```

### DIFF
--- a/lib/Db/Table.php
+++ b/lib/Db/Table.php
@@ -43,7 +43,7 @@ use OCP\AppFramework\Db\Entity;
  * @method setRowsCount(int $rowsCount)
  * @method getColumnsCount(): int
  * @method setColumnsCount(int $columnsCount)
- * @method getViews(): array
+ * @method getViews(): ?array
  * @method setViews(array $views)
  * @method getColumns(): array
  * @method setColumns(array $columns)


blizzz:
> When we do it this way, that means that when you export a shared table, the "views" array is empty. This is fine, right? @blizzz

Empty array is fine, but if some or all of the views are accessible those should be exported all along.

blizzz:
> > > When we do it this way, that means that when you export a shared table, the "views" array is empty. This is fine, right? @blizzz
> > 
> > 
> > Empty array is fine, but if some or all of the views are accessible those should be exported all along.
> 
> I would expect `getViews()` to automatically do this, but it seems it doesn't

Views are null, because they are not set yet. https://github.com/nextcloud/tables/blob/9a4656171688fe954455915e23c079ac7d9b2be3/lib/Db/Table.php#L46 is misleading, should also indicate that null is possible. `$this->enhanceTable($table)` would need to be called previously so that views are set (they still might be null without proper permissions – https://github.com/nextcloud/tables/blob/5e1ea605c3d08391738b029dd7b73745630bda82/lib/Service/TableService.php#L235-L238).

PR #1621, opened 2025-02-19
enhancement: UI fixes for create/edit row dialogs

enjeck:
- Realign "expand" icon for rich text in the "create row" modal
- Wrap column description text to avoid overflow

### Before
Screenshot from 2025-02-17 13-29-03

### After
Screenshot from 2025-02-17 13-32-17

- Resizes "edit row" modal since it was too large
### Before
Screenshot from 2025-02-17 13-40-14

### After
Screenshot from 2025-02-17 13-39-50

### DIFF
--- a/src/modules/modals/EditRow.vue
+++ b/src/modules/modals/EditRow.vue
@@ -6,7 +6,7 @@
 	<NcDialog v-if="showModal"
 		data-cy="editRowModal"
 		:name="t('tables', 'Edit row')"
-		size="large"
+		size="normal"
 		@closing="actionCancel">
 		<div class="modal__content" @keydown="onKeydown">
 			<div v-for="column in nonMetaColumns" :key="column.id">
--- a/src/modules/modals/EditRow.vue
+++ b/src/modules/modals/EditRow.vue
@@ -6,7 +6,7 @@
 	<NcDialog v-if="showModal"
 		data-cy="editRowModal"
 		:name="t('tables', 'Edit row')"
-		size="normal"
+		size="large"
 		@closing="actionCancel">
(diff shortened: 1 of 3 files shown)


src/modules/modals/EditRow.vue
```diff
@@ -6,7 +6,7 @@
 	<NcDialog v-if="showModal"
 		data-cy="editRowModal"
 		:name="t('tables', 'Edit row')"
-		size="large"
+		size="normal"
```
blizzz, line 9: That was done intentionally in https://github.com/nextcloud/tables/pull/1611

blizzz (changes requested):
reverts a change that was purposefully done recently

blizzz:
Adding @marcoambrosini for review. Maybe the size of the modals need a discussion. Or maybe expand/narrow buttons, maybe remembering users preference (in browser).

juliusknorr:
Please remember to add labels

marcoambrosini:
We talked about this in our 1on1 with @enjeck, I think it's fine to keep size large for all these dialogs

src/shared/components/ncTable/partials/rowTypePartials/RowFormWrapper.vue
```diff
@@ -91,7 +91,11 @@ export default {
 }
 
 pre {
-	white-space: pre;
+	white-space: pre-wrap;
+    white-space: -moz-pre-wrap;
```
juliusknorr, line 95: Spacing should use tabs like everywhere else?

juliusknorr (approved):
Nitpick on indentation, otherwise 👍

PR #1623, opened 2025-02-19
Fix: Fix errors related to navigation

Koc:
Right now we have 2 annoying errors in console. This pr fixes both of them.

image

image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -22,6 +22,7 @@ import Sidebar from './modules/sidebar/sections/Sidebar.vue'
 import { useResizeObserver } from '@vueuse/core'
 import { loadState } from '@nextcloud/initial-state'
 import { useTablesStore } from './store/store.js'
+import { generateUrl } from '@nextcloud/router'
 
 export default {
 	name: 'App',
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -85,20 +88,20 @@ export default {
 				this.setActiveTableId(parseInt(currentRoute.params.tableId))
 				this.setPageTitle(this.activeTable.title)
 				if (!currentRoute.path.includes('/row/')) {
-					this.switchActiveMenuEntry(document.querySelector('header .header-left .app-menu a[title="Tables"]'))
+					this.switchActiveMenuEntry(document.querySelector(`header .header-start .app-menu a[href="${url}"]`))
```
Koc, line 91: If you have any other idea how to make it works for non-en locales - feel free to suggest

Koc: see https://github.com/nextcloud/server/pull/51207

It can looks like:

```
if (currentRoute.path.startsWith('/table/')) {
	this.setActiveTableId(parseInt(currentRoute.params.tableId))
	this.setPageTitle(this.activeTable.title)
	if (!currentRoute.path.includes('/row/')) {
		emit('nextcloud:app-menu.active', 'tables')
}
```

blizzz:
@enjeck any updates?

Koc:
@enjeck cypress failures look unrelated to current PR

Koc:
So, let’s merge and fix Cypress separately?

PR #1656, opened 2025-02-25
enh: Restrict text link protocols

enjeck:
We want to only allow certain protocols for the text link field. For now, just http and https.

Error when an invalid protocol is entered looks like this, and you're unable to save:
Screenshot from 2025-02-25 05-17-01

### DIFF
--- a/src/modules/modals/EditRow.vue
+++ b/src/modules/modals/EditRow.vue
@@ -17,6 +17,10 @@
 					type="error">
 					{{ t('tables', '"{columnTitle}" should not be empty', { columnTitle: column.title }) }}
 				</NcNoteCard>
+				<NcNoteCard v-if="localRow[column.id] && column.type === 'text-link' && !isValidUrlProtocol(localRow[column.id])"
+					type="error">
+					{{ t('tables', 'Invalid protocol. Allowed: {allowed}', {allowed: allowedProtocols.join(', ')}) }}
+				</NcNoteCard>
 			</div>
 			<div class="row">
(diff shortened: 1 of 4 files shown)


src/modules/modals/EditRow.vue
```diff
@@ -17,6 +17,10 @@
 					type="error">
 					{{ t('tables', '"{columnTitle}" should not be empty', { columnTitle: column.title }) }}
 				</NcNoteCard>
+				<NcNoteCard v-if="localRow[column.id] && column.type === 'text-link' && !isValidUrlProtocol(localRow[column.id])"
+					type="error">
+					{{ t('tables', 'Invalid protocol. Allowed: {allowed}', {allowed: allowedProtocols.join(', ')}) }}
```
juliusknorr, line 22:
```suggestion
					{{ t('tables', 'Invalid protocol. Allowed: {allowed}', {allowed: allowedProtocols.map(p => p.replace(/:$/,'')).join(', ') }}
```

Could make the list a bit easier to read without the extra `:`

src/modules/modals/EditRow.vue
```diff
@@ -17,6 +17,10 @@
 					type="error">
 					{{ t('tables', '"{columnTitle}" should not be empty', { columnTitle: column.title }) }}
 				</NcNoteCard>
+				<NcNoteCard v-if="localRow[column.id] && column.type === 'text-link' && !isValidUrlProtocol(localRow[column.id])"
```
juliusknorr, line 20: Something separate but we could think about how we can reduce the code duplication between CreateRow and EditRow

juliusknorr (approved):

PR #1661, opened 2025-02-25
fix(Import): Fix DateTime subtype checks

michaelkuhn:
Trying to import a table with dates resulted in the following error:
```
Cannot use object of type OCA\Tables\Db\Column as array in file
'.../tables/lib/Service/ImportService.php' line 171
```


michaelkuhn:
The following schema and CSV reproduce the problem for me (Nextcloud 30, Tables 0.9.0): [demo.json](https://github.com/user-attachments/files/19028975/demo.json) [25-02-28_15-28_demo.csv](https://github.com/user-attachments/files/19028973/25-02-28_15-28_demo.csv)

michaelkuhn:
How did you import the data? If I import the CSV file directly, it also works for me. However, if I first import the schema (or use an existing table) and then populate the table using the CSV file, it fails.

blizzz:
@michaelkuhn thanks for your contribution. Alas, I cannot reproduce the issue either.

1. In Tables hit on the :heavy_plus_sign: sign to create a table
2. Give it a name and select *Import table* and continue
3. Upload from device and use your example file, click *Preview*
4. Click Import
5. See that importing is done successfully; click *Done*
6. See the expected result in the web interface (also after reload)
7. Check logs, see no error

blizzz (approved):

Issue #1288, opened 2024-08-12
Allow changing default value for new rows only


nickvergessen:
From my usability point of view, it would be cool if I can create a new column with a different value then the future default.

Usecase explanation:

1. I track a weekly thing.
2. From time to time a new column has to be created with Dropdown value "New", so they list as "New" on all old rows
3. But now I would like to change the default to "Missed" afterwards, so that when I add a new "date row" they are correctly set as "Missed"

But when I change the default to "Missed", all existing rows with the default get their value changed to "Missed" as well, instead of staying on "New".

Issue #1154, opened 2024-06-18
View feature (display column but non editable column)


Aveyron-RetD:
### Is your feature request related to a problem? Please describe.

On a specific workflow... 
You want to still see the previous column, but the workflow as validated the previous steps so they should not be editable.
You might still want to display them in the view, but they should not editable anymore

### Describe the solution you'd like

In here add the option to be read only/or editable.
![image](https://github.com/nextcloud/tables/assets/53221471/5ac94a7a-8607-45bc-9925-d9d86b4ef157)

gerherha:
I totally agree.
Without this option you can not make audit-proof work flows.
In the example with the vacation request, the boss can change the request of the employee without anybody noticing. This is questionable.

So it would be great to see an option like @Aveyron-RetD suggested.

Aveyron-RetD:
Similar to #953

blizzz:
- [ ] Needs: https://github.com/nextcloud/tables/pull/1688
- [ ] BE: handle storing, retrieving and updating *read-only* flag with the dedicated column content. The permission layer (service, perhaps middleware) may need to respect it in at least some places – especially creating and updating rows. Manage permissions always overrule a setting.
- [ ] FE: agree on UI to show additional options to columns in a view that is easily extendable. We would need it not only for this feature, but for others as well (e.g. toggling mandatory state). (This step does not need to wait for 1688.)
- [ ] FE: implementation of the results ^ and of course related business logic.

Issue #573, opened 2023-09-15
Views: Disable Editing for specific columns


xbluemonkx:
### Is your feature request related to a problem? Please describe.

The new Views-feature is amazing and showing that this app is absolutely heading in the right direction.

But what we need is the option to specify for each column of a view is editable or not. E.g. you want users to edit the column "comment" but not the column "supplier name"

### Describe the solution you'd like

add a switch-button for each column to specifiy if it is editable or not.

ISC-PRO:
Hello. Can you tell me what the further planning looks like?

blizzz:
@ISC-PRO this feature request is filed, but not picked up or planned at the moment. Are you interested in getting involved with it?

ISC-PRO:
Hello Arthur, how does the collaboration work and what are the requirements? Regards, Patrick

Von: Arthur Schiwon ***@***.***>
Gesendet: Montag, 6. Januar 2025 14:23
An: nextcloud/tables ***@***.***>
Cc: Rommerskirchen, Patrick (lff) ***@***.***>; Mention ***@***.***>
Betreff: [EXTERN]Re: [nextcloud/tables] Views: Disable Editing for specific columns (Issue #573)


[DIESE EMAIL STAMMT VON EINEM EXTERNEN ABSENDER. BITTE BEACHTEN SIE DIES VOR DEM ÖFFNEN VON INTERNET-LINKS ODER DATEIANHÄNGEN.]


@ISC-PRO<https://github.com/ISC-PRO> this feature request is filed, but not picked up or planned at the moment. Are you interested in getting involved with it?

ISC-PRO:
> [@ISC-PRO](https://github.com/ISC-PRO) this feature request is filed, but not picked up or planned at the moment. Are you interested in getting involved with it?

[@blizzz](https://github.com/blizzz) How does the collaboration work and what are the requirements?

blizzz:
> > [@ISC-PRO](https://github.com/ISC-PRO) this feature request is filed, but not picked up or planned at the moment. Are you interested in getting involved with it?
> 
> [@blizzz](https://github.com/blizzz) How does the collaboration work and what are the requirements?

Coming up with an idea how to implement this features, also UI-wise, and later on implementing it. Knowledge in PHP and JS (Vue) would be quite helpful.

xbluemonkx:
hi,
thanks for the hin. looks like this was done in #1958

PR #1688, opened 2025-03-12
refactor: restructure column config to support additional settings

AIlkiv:
I did this refactoring to unlock adding new properties to columns in the View. For example, overriding column attributes from a table or adding a column width.


AIlkiv:
I've made the necessary changes. Please run the pipeline again

blizzz:
Hi @AIlkiv, thanks for this contribution as well. Goal-wise it goes into the same direction of my current considerations to satisfy some requirements. My current approach would be to resolve the json column types however and add another database table. One difference however is that I need to override or adjust some behaviour for columns based in views. 

But this is also in-progress conceptual work so far. 

Bigger and more fundamental changes are ideally discussed and agreed up front before working on a implementation. Otherwise it might be much work put on somewhere where rework becomes necessary.

Can you lay out your considerations for this design?

AIlkiv:
@blizzz  Thank you for your feedback. Your task is more global and interesting. And I think such tasks should be decomposed. Because it's hard to get a good architectural solution if you have a task to “urgently add +1 column property”. With this PR, we make:
1. One point of saving properties. I think this will simplify further migration to a separate table (if necessary)
2. The pressure “we need new features for columns” is removed. Accordingly, you can thoroughly think through the solution

If you look at it from a broader perspective. I think it would be cool if the end user in a View could be offered the same form of editing a column as in a Table (override all values).
Maybe even make it a joint abstraction for View and Table columns at the code level.

blizzz:
@AIlkiv So there is urgency on your side, do I understand correctly? I get your point anyways.

I was double checking a few things, but that's not a full review so far, and I have to make up my mind on the view-column-case yet, that's still in progress. 

I had a look again at `oc_tables_columns` and have seen so far only the use case that the columns where references by either `id` or their `table_id`. That's helpful (as long as it stays like this) and having a column with flexible content is a feasible approach. In fact I am not that happy with having many many columns here that we have to extend, but this is a different topic. 

For we cannot really use the database's JSON features – each engine does something different, we tried that in the past, and it is very painful and we get away from this. A text/clob type therefore would be sufficient (i suppose it has less overhead on the DB than native json type, but did not double check).

Another thought that comes to my mind is e.g. moving `mandatory` out of a column into the new field, as it would be just such a setting (i suppose?) but would probably just leave it as is. Not settled on this yet. It would have an impact on the migration duration and such as well.

But actually most importantly: while this PR adds the functionality (which is good) is there also one where you make use of this? So that we have a real existing case to work towards? (Or did I oversee something?)

AIlkiv:
@blizzz
1. If you already have a final migration vision (a separate table or other structure). I can make the necessary changes.
2. I want to add the ability for the list field to specify a subset of options in the view. I have a table with task tracking and want to limit the available statuses in the view (when creating and editing rows).
3. I think not many users do complex internal tasks. And this is the kind of task. I assume that after this task, the community will start contributing more interesting features for end users. For example: https://github.com/nextcloud/tables/issues/35 https://github.com/nextcloud/tables/issues/573 https://github.com/nextcloud/tables/issues/1154 https://github.com/nextcloud/tables/issues/1288

blizzz:
@AIlkiv sorry, i could not drive this forward much on my side either. So about,

1) Nothing definite, but i start to fancy having the additional column directly here, and without  a structure enforced on write, that makes it more flexible. I played around with the thought of another table, which I would defer now unless structural changes on the same table are too expensive.

Furthermore I am wondering whether we actually could (and should?) stick with the columns we have and repurpose the `columns` column. It is already a json type, and contains at the moment an array of column IDs. Instead we could turn this into objects and have the column id as only mandatory element. Downside is we either have to transform the data or keep logic for both value types inside this array. Also I did not table check whether `columns` is used in any query part so far. 

2) Are your status a free text field, or would this be a subset of the options of a (single) selection column? This would be interesting.  

We are also looking into making columns read-only in views (like #1154), or toggle the state of being mandatory.

3) :+1:

AIlkiv:
@blizzz 
Thanks for the explanation!

The status field is intended to be a selection column with a predefined list of options — not free text.

To keep things clean and avoid any risk of confusion in the logic, I created a separate column specifically for this purpose.

My goal is to allow a view to limit the available options for this column. For example, in a task-tracking table, a certain view could allow users to pick only between “In Progress” and “Blocked,” even though the full list of statuses includes more values like “New,” “Done,” etc.

blizzz:
> What makes the current solution better than having a separate DB table?

Each additional column property/setting would require an additional column and thus an additional database migration path. While this is something you want to avoid (takes longer the more data you have) it does not come with much benefit, unless you really really need to be able to include those values in WHERE clauses or alike.

blizzz:
> > Each additional column property/setting would require an additional column and thus an additional database migration path
> 
> And the only other alternative is using JSON, which you said comes with its own problems?

JSON content is fine, but we cannot make efficient use of the native JSON type, because they differ far too much across the different supported databases. It makes sense to use TEXT instead (and potentially save overhead in the databases) and still store JSON inside.

blizzz (changes requested):
First review (but also continued discussion)

My first idea was to create an additional table, but I came more to liking the approach to contain the data as json in a column, and thus closer to you approach. This works under the idea that the values are used in business logic only, but not within the queries.

This also requires the migration step and converting the values from the `columns` column. This will have update impact especially on bigger installations, what we ideally avoid. Row data has to be written here probably twice by the database: once for adding the column, and a second time to fill it with values.

We might want to re-purpose the `columns` therefore and either only update it once in the migration steps, or only on demand and live with two formats at least for a while.

The latter also helps with the topic of backwards compatibility, and especially for other clients (cc @stefan-niedermann for the android app). This is a topic nevertheless, because either we have now a changed response with a different field (`columnSettings` instead of `columns`) vs the same field with different formats. Another approach (but with the migrations pain) is to keep both `column` and `columnSettings` working. It might be an opportunity to leave  the population of settings_columns for an expensive migration job that would run in the maintenance timeframe for background jobs :thinking: 

My original idea was also to extend the API with the new settings, but otherwise not change it. This keeps existing apps and scripts working, while still providing new features. This is not bound to the means of storing the information in the DB, i.e. also the combined `columns` column can be used to deliver information with both keys and formats. And also when repurposing the `columns` column the job can be done once in the background (only updated views would get the new format earlier, this can be handled though).

stefan-niedermann:
Thank you for looping me in, @blizzz !

I did not yet start implementing support for views, so I'll wait how all those decisions will affect the APIs in the end.

AIlkiv:
@blizzz Thank you for your feedback.

1. I made changes according to the comments.
2. Now the API returns the “columns” key again for better backward compatibility.

lib/ResponseDefinitions.php
```diff
@@ -25,7 +25,7 @@
  * 	lastEditBy: string,
  * 	lastEditAt: string,
  *  description: string|null,
- *  columns: int[],
```
blizzz, line 28: should be kept then (or with the remaining task, updated instead)

blizzz (commented):
Thank you @AIlkiv! Great progress.

So what is left, and I pointed in my previous reply to it, is the database migration part, boiling down to:

_DB column: re-purpose `columns` column instead of adding `columnSettings`?_

It would be cleaner to have the additional column added to the database table. Unfortunately this is costly with MySQL afaik. On big installation it may increase the upgrade time procedure not only for adding the column, but also populating it. 

We can avoid it by using the `columns` column and have business logic handle cases, where the old format is written to the database. On an update, we would save it with the new format. And we can have a background job (for maintenance hours) and/or an occ command to let it run in the background, outside of the upgrade cycle (does not need to go into this PR, can be a follow up). 

The downside of course is that we have consider both cases of payload in that column, at least for a while.

With the upgrade scenario in mind i think this is the better way to go, although we are operating a bit blind (we do not know what the largest tables are in the wild), this would be the defensive approach to tackle this.

This is the last bigger thing I would ask, for this thought kept haunting me.

AIlkiv:
@blizzz 
Done.
1. I am not adding the column
2. The migration is performed in the background task.
3. Two formats of data stored in the database are supported at the code level.

blizzz:
I missed the notification :pensive: Thanks @AIlkiv I'll have a look

lib/BackgroundJob/ConvertViewColumnsFormat.php
```diff
@@ -0,0 +1,99 @@
...
+
+namespace OCA\Tables\BackgroundJob;
+
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\IJobList;
+use OCP\BackgroundJob\Job;
+use OCP\IDBConnection;
+use OCP\IUserManager;
+
+class ConvertViewColumnsFormat extends Job {
```
blizzz, line 13:
```suggestion
class ConvertViewColumnsFormat extends TimedJob {
	protected int $timeSensitivity = TIME_INSENSITIVE;
```

Mark it time insensitive so that it can be executed in the maintenance window (cf. https://docs.nextcloud.com/server/latest/developer_manual/basics/backgroundjobs.html#heavy-load-and-time-insensitive)

lib/BackgroundJob/ConvertViewColumnsFormat.php
```diff
@@ -0,0 +1,99 @@
...
+		$result = $qb->executeQuery();
+
+		// Process new views if any
+		while ($view = $result->fetch()) {
+			$this->processView($view, $updateQb);
+		}
+
+		$result->closeCursor();
+	}
+
```
blizzz, line 63: for this should be run once only, the job has to be removed from the list, as in
 
https://github.com/nextcloud/server/blob/a8f46af20f4fccac0257eba950e70d0da96c4a5a/lib/public/BackgroundJob/QueuedJob.php#L37-L43

lib/Migration/Version001000Date20240328000000.php
```diff
@@ -0,0 +1,48 @@
...
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 * @throws Exception
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		return null;
+	}
```
blizzz, line 37: can be left out completely

lib/Migration/Version001000Date20240328000000.php
```diff
@@ -0,0 +1,48 @@
...
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		// Register background job to convert column format
+		$this->jobList->add(ConvertViewColumnsFormat::class);
```
blizzz, line 46: Need to dobule check that this does not stumble on (web) upgrades due to cached autoloading info

lib/Service/ViewService.php
```diff
@@ -257,24 +257,26 @@ public function update(int $id, array $data, ?string $userId = null, bool $skipT
 				throw new PermissionError('PermissionError: can not update view with id ' . $id);
 			}
 
-			$updatableParameter = ['title', 'emoji', 'description', 'columns', 'sort', 'filter'];
+			$updatableParameter = ['title', 'emoji', 'description', 'sort', 'filter', 'columnSettings'];
```
blizzz, line 260: Mh, is removal of `columns` a hidden API break here?

blizzz (commented):
Yes, it is taking shape :) A few comments and suggestions. Please also address the findings from psalm.

AIlkiv:
@blizzz psalm fixed

AIlkiv:
@blizzz kind reminder

lib/Db/View.php
```diff
@@ -84,7 +86,35 @@ public function __construct() {
...
+	/**
+	 * @return array<array{columnId: int, order: int}>
+	 */
+	public function getColumnsSettingsArray(): array {
+		$columns = $this->getArray($this->getColumns());
+		if (empty($columns)) {
+			return [];
+		}
+		
+		if (is_array($columns[0])) {
```
blizzz: not a blocker, generally it is safer to use `array_key_first()` instead of assuming `[0]`, but should be fairly ok here.

lib/Migration/Version001000Date20240328000000.php
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		// Register background job to convert column format
+		$this->jobList->add("OCA\Tables\BackgroundJob\ConvertViewColumnsFormat");
```
blizzz, line 33:
```suggestion
		$this->jobList->add(\OCA\Tables\BackgroundJob\ConvertViewColumnsFormat::class);
```

Ah, I saw you have had it before as a string, this should not make any difference.

lib/Service/ViewService.php
```diff
@@ -520,15 +535,19 @@ function (array $filter) use ($columnId) {
 					},
 					$view->getFilterArray()
 				),
-
 				fn ($filterGroup) => !empty($filterGroup)
-
 			);
 
+			$columnSettings = $view->getColumnsSettingsArray();
+			$columnSettings = array_filter($columnSettings, function (array $setting) use ($columnId) {
```
blizzz:
```suggestion
			$columnSettings = array_filter($columnSettings, function (array $setting) use ($columnId): bool {
```

blizzz (approved):
Works, code looks good, some non-blocking cosmetics findings, but mergable

lib/Migration/Version001000Date20240328000000.php
AIlkiv: I changed it to a string to ensure that during updates through the web there wouldn’t be any issues with the class not being found.

AIlkiv:
@blizzz I’ve already made the suggested changes.

lib/Db/View.php
```diff
@@ -125,6 +155,10 @@ public function setColumnsArray(array $array):void {
 		$this->setColumns(\json_encode($array));
 	}
 
+	public function setColumnsSettingsArray(array $array): void {
```
blizzz, line 158: This seems unused, right? @AIlkiv 

No action needed, I am on another PR, just want to check back.

PR #1689, opened 2025-03-12
fix: fix creation failure when using creator filter

AIlkiv:
When create view with filter

image

when adding new records, I get an error

```
An internal error or exception occurred: OCA\\Tables\\Service\\RowService - create: OCA\\Tables\\Db\\Row2Mapper - insertCell: Can not insert cell, because the given column-id is not known
```

this PR fixes this bug

### DIFF
--- a/lib/Service/RowService.php
+++ b/lib/Service/RowService.php
@@ -261,6 +261,11 @@ class RowService extends SuperService {
 					continue;
 				}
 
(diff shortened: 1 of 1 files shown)


AIlkiv:
@juliusknorr Thank you. I already have an account.

AIlkiv:
A failed test is not related to this task

blizzz:
Thanks for your contribution @AIlkiv!

Issue #69, opened 2022-10-09
duplicate rows / copy and modify


tatrapikao:
First and foremost thank you for this app! It's gonna be amazing. It would be nice to be able to copy one or multiple rows. Perhaps an additional "copy and modify" Option in the Edit Menu of an existing Row would be a great feature.

jancborchardt:
Sounds good! Some options how to do that:
1. We put it in the "Edit row" modal, like how "Delete" is placed currently. Good thing about this is that all is in one place and editing is quick, bad thing is that it’s not easy to discover, and stuff is a bit cramped.
2. We make the "Edit" button a menu with 3 entries: Edit row, Duplicate row, Delete row. This is a fairly standard and well understandable approach, it just requires an extra click to edit a row compared to now. I would say though it’s fine.
3. Extra inbetweener option: We do it like we do it in our own user/account management of Nextcloud with 2 buttons next to each other. "Edit row" is outside of the "More actions" menu, and the other actions like "Duplicate row" and "Delete row" are inside the 3-dot menu.

@datenangebot @nimishavijay what do you think? I tend towards 2 or 3.

nimishavijay:
Really nice idea! I like the 3rd approach as well! We could do 3 on desktop and 2 on mobile :)

ChristophWurst:
Workaround: insert one row, export as CSV. Make duplicates in the downloaded file. Upload again.

tatrapikao:
> Workaround: insert one row, export as CSV. Make duplicates in the downloaded file. Upload again.

Does not even work for me. If i export the selcted row to csv and then try to re-import it in to the table to "duplicate" it, the import dialog just hangs forever..

ChristophWurst:
Open the browser console. It could be something like https://github.com/nextcloud/tables/issues/1067. Also :eyes: @luka-nextcloud there might still be no error handling for generic errors.

tatrapikao:
> Open the browser console. It could be something like #1067. Also 👀 @luka-nextcloud there might still be no error handling for generic errors.

absolutely. So we identified following issues:

- not possible to copy Data
- csv export works but uses wrong mime Type "text/plain"
- csv import has no error handling for MIME-Check #1067 
- csv import does not accept "text/csv" as MIME-Type  (..it really should when used in export)

xbluemonkx:
Hi,
[...]

Issue #660, opened 2023-11-09
UNIQUE constraint on columns


mortee:
### Is your feature request related to a problem? Please describe.

I can't specify that a column's values need to be unique. If Tables is to be considered anything like a database, then this would be essential.

### Describe the solution you'd like

Just like there's a "required" switch on the column creation dialog (NOT NULL), a "unique" switch should be there too. Row creation should fail if the value is not unique.

datenangebot:
Do you have a special use case in mind?

We don't want to cover all features a db would have, so lets talk about use cases to see if this helpful.

mortee:
My case at hand involves storing some information about correspondents based on their email address. I would like Tables to prevent me from inserting the same address twice.

I'm definitely not suggesting covering all DB features here, but I think a QUNIQUE requirement on a field is so utterly common across all kinds of table data, that it doesn't even require much justification other than that.

Also, I think it would definitely not be confusing for users, this is a very clear and obvious principle. 

Also (just guessing here) implementing it would be close to trivial: the UI would consist of a single additional switch on the column dialog, and a failure condition on creation/update. Also, since it is a primitive feature of any underlying DB, the back-end part would be trivial to implement too.

xbluemonkx:
> Do you have a special use case in mind?
> 
> We don't want to cover all features a db would have, so lets talk about use cases to see if this helpful.

I don't think that this crucial feature https://github.com/nextcloud/tables/issues/172 will work well without the ability to avoid duplicates.

PR #1713, opened 2025-03-31
Add row actions to delete or duplicate a row

juliusknorr:
- [x] Cypress tests
- [x] Allow to duplicate and edit (from https://github.com/nextcloud/tables/issues/69)
- [x] Test cloning in views, discuss if we may want to clone and keep hidden data if we are in a view with limited columns)
- [x] Handlle unique columns (probably just open an editable modal of the new row)

## Follow up

 - [ ] Opening an edit modal

### DIFF
--- a/cypress/e2e/tables-rows.cy.js
+++ b/cypress/e2e/tables-rows.cy.js
@@ -107,4 +107,107 @@ describe('Rows for a table', () => {
 		cy.get('[data-cy="ncTable"] table').contains('Edited inline').should('exist')
 		cy.get('[data-cy="ncTable"] table').contains('Test inline editing').should('not.exist')
 	})
+
+	it('Duplicate row using action menu', () => {
+		cy.loadTable('Welcome to Nextcloud Tables!')
+		cy.get('[data-cy="createRowBtn"]').click({ force: true })
+		cy.get('[data-cy="createRowModal"] .slot input').first().type('Original row')
+		cy.get('[data-cy="createRowModal"] .ProseMirror').first().click()
+		cy.get('[data-cy="createRowModal"] .ProseMirror').first().clear()
+		cy.get('[data-cy="createRowModal"] .ProseMirror').first().type('Original description')
+		cy.get('[data-cy="createRowModal"] [aria-label="Increase stars"]').click().click()
+		cy.get('[data-cy="createRowSaveButton"]').click()
+
+		cy.get('[data-cy="createRowModal"]').should('not.exist')
+		cy.get('[data-cy="ncTable"] table').contains('Original row').should('exist')
+		
+		// Find the row and click duplicate action
+		cy.get('[data-cy="ncTable"] [data-cy="customTableRow"]').contains('Original row').closest('[data-cy="customTableRow"]').within(() => {
+			// Click the actions menu button (three dots)
+			cy.get('.action-item__menutoggle').click()
+		})
+		
+		// Click duplicate action
+		cy.get('[data-cy="duplicateRowBtn"]').click()
+		
+		// Verify the row was duplicated
+		cy.get('.icon-loading').should('not.exist')
+		cy.get('[data-cy="ncTable"] table').contains('Original row').should('have.length.at.least', 2)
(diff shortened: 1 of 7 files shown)


xbluemonkx:
@juliusknorr  - great that you are working on this. - when duplicating rows please keep in mind that this ticket exists https://github.com/nextcloud/tables/issues/660

There are good reasons to provide a "unique" setting for fields. - even this is not yet an existing feature, please keep in mind that it might be implemented one day and copying rows must not circumvent this setting.

enjeck:
> when duplicating rows please keep in mind that this ticket exists #660

Thanks for the shout! Will keep it in mind

enjeck:
> * Handlle unique columns (probably just open an editable modal of the new row)

At the moment, it currently doesn't allow duplicating a row if there's a unique condition.

cypress/e2e/tables-rows.cy.js
```diff
@@ -75,7 +76,7 @@ describe('Rows for a table', () => {
 		cy.get('[data-cy="ncTable"] [data-cy="customTableRow"]').contains('My first task').closest('[data-cy="customTableRow"]').find('[data-cy="editRowBtn"]').click()
 		cy.get('[data-cy="editRowModal"] .notecard--error').should('not.exist')
 		cy.get('[data-cy="editRowModal"] .slot input').first().clear()
-		cy.get('[data-cy="editRowModal"] .notecard--error').should('exist')
+		//cy.get('[data-cy="editRowModal"] .notecard--error').should('exist')
```
juliusknorr, line 79: This is an unexpected change, can you explain?

src/shared/components/ncTable/partials/TableRow.vue
```diff
@@ -22,19 +22,50 @@
...
 </template>
 
 <script>
-import { NcCheckboxRadioSwitch, NcButton } from '@nextcloud/vue'
 import Fullscreen from 'vue-material-design-icons/Fullscreen.vue'
+import { mapActions } from 'pinia'
+import { useDataStore } from '../../../../store/data.js'
+import { NcCheckboxRadioSwitch, NcButton, NcActions, NcActionButton } from '@nextcloud/vue'
+import { showError, showSuccess } from '@nextcloud/dialogs'
+import Pencil from 'vue-material-design-icons/Pencil.vue'
```
juliusknorr, line 67: Pencil is unused now, I would actually vote to move from the Fullscreen to the Pencil icon for editing as that is more commonly used in Nextcloud for editing

src/shared/components/ncTable/partials/TableRow.vue
```diff
@@ -197,6 +234,33 @@ export default {
...
+			}, {})
+			const res = await this.insertNewRow({
+				viewId: this.isView ? this.elementId : null,
+				tableId: this.isView ? null : this.elementId,
+				data,
+			})
+			if (!res) {
+				showError(t('tables', 'Could not duplicate row.'))
+			} else {
+				showSuccess(t('tables', 'Row duplicated successfully.'))
```
juliusknorr, line 262: We can avoid a success message here if the new row is visible right away.

src/shared/components/ncTable/partials/TableRow.vue
```diff
@@ -197,6 +234,33 @@ export default {
...
 			}
 		},
+		...mapActions(useDataStore, ['removeRow', 'insertNewRow']),
+		handleDeleteRow() {
+			emit('tables:row:delete', { rows: [this.row.id], isView: this.isView, elementId: this.elementId })
+		},
+		async handleCloneRow() {
+			const data = this.row.data.reduce((acc, curr) => {
+				const column = this.visibleColumns.find(col => col.id === curr.columnId)
+				// Skip unique text columns to avoid constraint violations
```
juliusknorr, line 245: I'm unsure what the expected behaviour is. The test checks that this fails, but the logic here seems to be around skipping the values then on the newly inserted rows. 

However when testing this it did not skip the row. I'm fine either way, but should be clear what happens then

enjeck:
I prefer https://github.com/nextcloud/tables/pull/2456

PR #1715, opened 2025-04-01
Avoid renovate vueuse updates above version 11 until we migrate to vue3

enjeck:
@vueuse/core versions above 12 use vue3, which we don't support yet. We keep getting auto updates that need to be closed e.g https://github.com/nextcloud/tables/pull/1693

Also remove vuex since no longer necessary now that we've switched to Pinia

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -92,15 +92,15 @@
 		},
 		{
 			"matchPackageNames": [
-				"vuex"
+				"vue-router"
 			],
 			"allowedVersions": "<4"
 		},
 		{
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -92,15 +92,15 @@
 		},
 		{
 			"matchPackageNames": [
-				"vuex"
+				"vue-router"
```
max-nextcloud, line 95: Did you intend to remove the rule for `vuex` ?

juliusknorr (approved):

PR #1718, opened 2025-04-02
feat(Filter): add is-not-equal operator

blizzz:
:warning: This is a rudimentary and incomplete commit. I loaded a csv into tables for some analysis and quickly needed the not-equal operator for text lines. I do not expect to come back to this shortly, but it can be picked up by someone :)


reefsofts-rhosds:
@MiloBalara 
That's the source, not the final build, so this is normal, because you should not just "run the source" in production, even if you can. Especially with JS as backend. (currently remembering trying to run anything with node)

silverkszlo:
I picked up this branch and extended the implementation.

- [x] Completed the support for the “is not equal” operator
- [x] Added support for the “does not contain” operator
- [x] Adjusted column-specific filter handling where necessary (e.g. SelectionCheck, Datetime, Usergroup)
- [ ] Update relevant tests

This PR now introduces both negative filter operators (is not equal, does not contain) consistently across supported column types.

lib/Db/Row2Mapper.php
```diff
@@ -172,15 +172,20 @@ private function getWantedRowIds(string $userId, int $tableId, ?array $filter =
...
+			$rows = $this->getRows($wantedRowIdsArray, $showColumnIds);
 
-		// Sort rows in PHP to preserve the order from getWantedRowIds
-		return $this->sortRowsByIds($rows, $wantedRowIdsArray);
+			// Sort rows in PHP to preserve the order from getWantedRowIds
+			return $this->sortRowsByIds($rows, $wantedRowIdsArray);
+		} catch (DoesNotExistException $e) {
+			$this->logger->error($e->getMessage(), ['exception' => $e]);
+			throw new InternalError(get_class($this) . ' - ' . __FUNCTION__ . ': ' . $e->getMessage());
+		}
```
blizzz, line 188: What is the reason behind this change on the Mapper level (and the other switches to `InternalError` in this class)?

silverkszlo: To be fair, those changes may be slightly out of the scope of this PR and were more of a 'refactor-while-passing-by'. I added them primarily to make the functions more robust. By catching and logging DoesNotExistException and then throwing a controlled InternalError, the idea is to prevent unhandled exceptions from crashing the application and to make debugging easier.

blizzz: OK

blizzz:
> Ready to merge?? @blizzz

Let's go

PR #1719, opened 2025-04-03
feat: Inline editing (Number, Progress, Yes/No, Textline)

### LINKED ISSUE #364: Inline editing
Lets have view mode that allows us to navigate throw the cells (also with the keys) and edit them instantly.

- [ ] Add ability for selection column to have the user add new arbitrary values
- [ ] Autosave 

Data types we have now:
Image

## Progress:
- [ ] TableCellDateTime: https://github.com/nextcloud/tables/pull/1974
- [ ] TableCellEditor: https://github.com/nextcloud/tables/pull/1976
- [ ] TableCellHtml
- [ ] TableCellLink: https://github.com/nextcloud/tables/pull/1977
    - [ ] Check invalid protocols
- [ ] TableCellMultiSelection: https://github.com/nextcloud/tables/pull/1975
- [x] TableCellNumber: https://github.com/nextcloud/tables/pull/1719
    - [x] Ensure min, max and step work properly
- [x] TableCellProgress: https://github.com/nextcloud/tables/pull/1719
    - [x]  Ensure min, max works properly
- [ ] TableCellSelection: https://github.com/nextcloud/tables/pull/1975
- [ ] TableCellStars: https://github.com/nextcloud/tables/pull/1973
- [x] TableCellTextLine: https://github.com/nextcloud/tables/pull/1719
- [ ] TableCellUsergroup: https://github.com/nextcloud/tables/pull/1978
- [x] TableCellYesNo: https://github.com/nextcloud/tables/pull/1719

enjeck:
Fixes #364 

## TODO:
- [x] check duplicate updateRow request
- [x] Manual testing
- [x] Cypress testing
- [x] move common styles between `TableCell`s to global
- [x] Align loading icon after saving row value

### DIFF
--- a/src/shared/components/ncTable/partials/TableCellNumber.vue
+++ b/src/shared/components/ncTable/partials/TableCellNumber.vue
@@ -4,7 +4,7 @@
 -->
 <template>
 	<div class="cell-number">
-		<div v-if="!isEditing" @click="startEditing">
+		<div v-if="!isEditing" class="number-display" @click="startEditing">
 			{{ column.numberPrefix }}{{ getValue }}{{ column.numberSuffix }}
 		</div>
 		<div v-else class="inline-editing-container">
@@ -76,7 +76,17 @@ export default {
 				return
 			}
 
-			const newValue = this.editValue === '' ? null : Number(this.editValue)
+			let newValue = this.editValue === '' ? null : Number(this.editValue)
+
+			if (newValue !== null && !isNaN(newValue)) {
+				if (this.getMin !== null && newValue < this.getMin) {
+					newValue = this.getMin
+				}
+				if (this.getMax !== null && newValue > this.getMax) {
+					newValue = this.getMax
+				}
+			}
+
 			const success = await this.updateCellValue(newValue)
 
 			if (!success) {
@@ -98,6 +108,12 @@ export default {
 	div {
 		min-height: 20px;
 	}
+}
+
+.number-display {
+	width: 100%;
+	min-height: 20px;
+	cursor: pointer;
 }
 
 .inline-editing-container {
--- a/src/shared/components/ncTable/partials/TableCellProgress.vue
+++ b/src/shared/components/ncTable/partials/TableCellProgress.vue
@@ -105,6 +105,7 @@ export default {
 .progress-display {
 	width: 100%;
 	min-height: 20px;
+	cursor: pointer;
 }
 
 .empty-progress-placeholder {


marcoambrosini (commented):
Nice work @enjeck 
From a design perspective, we need to hack these input components a bit, or modify them upstream so that:
- they span the whole width and height of the cell
- remove border radius
- Remove the `--color-main-background` outline when focused, the (dark one is enough for these inputs)

This will result in a much cleaner interface and no jumping when these inputs are focused

marcoambrosini (approved):
sgtm!

juliusknorr:
> I think it's easier to review if we deliver this over several PRs, since the various column types would require varying effort and styling. 

Agreed, doing this in smaller batches of changes makes sense

marcoambrosini:
@enjeck @juliusknorr what prevents us from merging this as step 1 as we discussed before?

src/shared/components/ncTable/partials/TableCellNumber.vue
```diff
@@ -3,44 +3,151 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 <template>
-	<div>
-		{{ column.numberPrefix }}{{ getValue }}{{ column.numberSuffix }}
+	<div class="cell-number">
+		<div v-if="!isEditing" @dblclick="startEditing">
```
juliusknorr, line 7: @marcoambrosini @jancborchardt I remember double click was something we had some debates about in the past. Is this fine as a trigger for the inline editing even though it is not very common in web apps? WDYT?

src/store/data.js
```diff
@@ -158,13 +162,24 @@ export const useDataStore = defineStore('data', {
 				return false
 			}
 
+			res.data.forEach(row => {
+				this.rowMetadata[row.id] = {
```
juliusknorr, line 166: Could this lead to problems if we load a context that has two views that show the same row at the same time? Then we could have two different viewIds that are used

src/shared/components/ncTable/partials/TableCellProgress.vue
```diff
@@ -3,47 +3,141 @@
...
+
+		startEditing() {
+			this.editValue = this.value
+			this.isEditing = true
+			this.$nextTick(() => {
+				this.$refs.input.focus()
+			})
+		},
+
+		async saveChanges() {
```
juliusknorr, line 90: This seems to be quite some duplication with the other components, maybe there is a reasonable way to make parts reusable here?

juliusknorr:
>  We should not show inline spinners and wait for the request to be completed to update the checkbox state. Just flip it immediately and make the request in the background

I'm not sure, how do other solutions handle this like Airtable, baserow? Our checkbox component might be nice for this which also offers this loading indicator: https://nextcloud-vue-components.netlify.app/#/Components/NcCheckboxRadioSwitch

>  I Would remove the gray line hover feedback

I'm not sure, I tend to even suggest to use actual [NcTextFields](https://nextcloud-vue-components.netlify.app/#/Components/NcFields?id=nctextfield), especially for accessibility we should have some visual focus feedback.

juliusknorr:
Some more more UX feedback from my side:

- [ ] Number field: Editing does not show decimals and prefix/suffix, probably still worth to have shown as well
- [ ] Number field: Can we also right align the text in the input? Could be a [type number](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number)
- [ ] Progress field: It seems like the click target to double click is rather small and hard to click
- [ ] Progress field: Could be a [type number](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number)
- [ ] Progress field but might be a general thing. If we have an empty default and no value there is no progress bar to double click
- [ ] As long as we do not support all types, for the ones we do not support yet we should probably just open the edit modal on double click

juliusknorr:
Works quite nice already, but I've left a few more comments inline to check

src/shared/components/ncTable/partials/TableCellNumber.vue
marcoambrosini: No, forgot to comment on that, should be on single click

src/store/data.js
enjeck: Good catch. Yes, this causes problems

src/shared/components/ncTable/partials/TableCellNumber.vue
```diff
@@ -3,44 +3,115 @@
...
+			// Prevent multiple executions of saveChanges
+			if (this.localLoading) {
+				return
+			}
+
+			if (Number(this.editValue) === this.value) {
+				this.isEditing = false
+				return
+			}
+
```
ghost, line 78: The number validation doesn't work for me. I can enter and save numbers higher than max, lower than min, or to extra decimal places. One approach would be to have `saveChanges()` check `input.validity.valid` and either fail or "fix" (round up or down, or trim decimals from) the input if it's invalid.

emberfiend (commented):
I think numbers and progress bars would benefit from a hover clue (`cursor: pointer`) to show that they're editable (like the other inline-editables have).

Apart from this and my inline comment, looks good :)

emberfiend:
@enjeck The TableCellNumber min/max clamping looks good, but you can still enter numbers to any decimal length. They'll appear rounded to the desired number of decimal places after you leave the editing mode, but when you click to edit the cell again, whatever long decimal you entered originally reappears.

I think this is how the "full row editor" does it, using Number.toFixed:

https://github.com/nextcloud/tables/blob/main/src/shared/components/ncTable/partials/rowTypePartials/NumberForm.vue#L94

PR #1724, opened 2025-04-04
fix(UI): fetch avatar correctly by user id

### LINKED ISSUE #1723: Avatar of users in the share list is broken
Image

Also according to Andy the ... on my avatar is not working properly (most likely due to broken user id which also causes the avatar to not render)

blizzz:
If an avatar is not set individually, it will already create one based on the displayname. It is nothing we have to take care of here.

Fixes #1723

### DIFF
--- a/src/modules/sidebar/partials/ShareList.vue
+++ b/src/modules/sidebar/partials/ShareList.vue
@@ -12,7 +12,7 @@
 				class="row">
 				<div class="fix-col-2">
 					<div style="display:flex; align-items: center;">
-						<NcAvatar :user="getShareAvatar(share)" :is-no-user="share.receiverType !== 'user'" />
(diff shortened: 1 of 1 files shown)


src/modules/sidebar/partials/ShareList.vue
```diff
@@ -12,7 +12,7 @@
 				class="row">
 				<div class="fix-col-2">
 					<div style="display:flex; align-items: center;">
-						<NcAvatar :user="getShareAvatar(share)" :is-no-user="share.receiverType !== 'user'" />
+						<NcAvatar :user="share.receiver" :is-no-user="share.receiverType !== 'user'" />
```
nickvergessen, line 15:
```vue
<NcAvatar :user="share.receiver" :display-name="share.receiverDisplayName || share.receiver" :is-no-user="share.receiverType !== 'user'" />
```
So that the default avatar uses the color and letters of the displayname

PR #1736, opened 2025-04-08
[stable0.8]: build(NC): bump max compatible NC version

enjeck:


blizzz (changes requested):
Nope, that is covered by 0.9.

PR #1748, opened 2025-04-13
Fix: computed property duplicates action on import

Koc:
Right now we have duplicated `getColumnsFromBE` import: in computed and in actions. Let's keep actions only


juliusknorr (approved):
Good catch, thanks a lot 👍

PR #1749, opened 2025-04-14
fix(DB): reduce number of SQL queries to get shares

blizzz:


lib/Db/ShareMapper.php
```diff
@@ -139,6 +139,27 @@ public function findAllSharesForNodeFor(string $nodeType, int $nodeId, string $r
...
+						$qb->expr()->eq('receiver', $qb->createNamedParameter($receivingUser, IQueryBuilder::PARAM_STR)),
+						$qb->expr()->eq('receiver_type', $qb->createNamedParameter('user',IQueryBuilder::PARAM_STR)),
+					),
+					$qb->expr()->andX(
+						$qb->expr()->in('receiver', $qb->createNamedParameter($receivingGroups, IQueryBuilder::PARAM_STR_ARRAY)),
+						$qb->expr()->eq('receiver_type', $qb->createNamedParameter('group',IQueryBuilder::PARAM_STR)),
+					)
+				)
+			)
+			->andWhere($qb->expr()->eq('node_type', $qb->createNamedParameter($nodeType, IQueryBuilder::PARAM_STR)))
```
juliusknorr, line 158: While at it please add an index as well for this one if it is worth, seems we don't have any on oc_tables_share yet 🙈

blizzz: That's the plan. We get a anonymized production dump, when experimenting with indexes and dev data, we always ended up with a full table scan either way.

lib/Db/ShareMapper.php
```diff
@@ -139,6 +139,27 @@ public function findAllSharesForNodeFor(string $nodeType, int $nodeId, string $r
...
+		$qb->select('*')
+			->from($this->table)
+			->where(
+				$qb->expr()->orX(
+					$qb->expr()->andX(
+						$qb->expr()->eq('receiver', $qb->createNamedParameter($receivingUser, IQueryBuilder::PARAM_STR)),
+						$qb->expr()->eq('receiver_type', $qb->createNamedParameter('user',IQueryBuilder::PARAM_STR)),
+					),
+					$qb->expr()->andX(
+						$qb->expr()->in('receiver', $qb->createNamedParameter($receivingGroups, IQueryBuilder::PARAM_STR_ARRAY)),
```
AIlkiv: Please add a check here that $receivingGroups is not empty

lib/Db/ShareMapper.php
```diff
@@ -139,6 +139,27 @@ public function findAllSharesForNodeFor(string $nodeType, int $nodeId, string $r
 		return $this->findEntities($qb);
 	}
 
+	public function findAllSharesForNodeTo(string $nodeType, int $nodeId, string $receivingUser, array $receivingGroups): array {
```
AIlkiv: Why not add logic for circles to this method?

enjeck: Added!

lib/Helper/UserHelper.php
```diff
@@ -69,10 +69,9 @@ public function getGroupIdsForUser(string $userId): ?array {
 			return null;
 		}
 
-		$groupArray = [];
-		foreach ($userGroups as $group) {
-			$groupArray[] = $group->getGID();
-		}
+		$groupArray = array_map(function (IGroup $group) {
+			return $group->getGID();
+		}, $userGroups);
```
blizzz, line 74: Looks nicer, but is heavier on RAM. Not a blocker.

enjeck: Oh, I just copied from your earlier commit at https://github.com/nextcloud/tables/pull/1749/commits/61297b9d46cd460814fc37d962ca9eb39b66faa2

blizzz: Then this message is redirect to me. With small data sized it's not that significant. With groups, could be so so.

PR #1750, opened 2025-04-14
fix: Properly update usergroup cells

### LINKED ISSUE #1558: When multiple Users and Groups are used, saving changes in one loses data in the other
### Steps to reproduce

1. Create a table
2. Add a column (A) that accepts one user
3. Add a column (B) that accepts multiple users
4. Add a row with data in column A & B
5. Reload page in browser
6. Data in column A is gone
7. Add data again to column A and save row
8. Reload page in browser
9. Data in column B is gone
10. Edit data in column B and save row
11. Reload page in browser
12. Data in column A is gone
13. Repeat ad nauseam

### Expected behavior

Data stays in the column regardless of the other column.

### Actual behavior

Se above

### Tables app version

0.8.3

### Browser

All

### Client operating system

All

### Operating system

Rocky 8

### Web server

Nginx

### PHP engine version

None

### Database

MariaDB

### Additional info

_No response_

enjeck:
fixes #1558 
When a usergroup field has multiple values, each is saved as a separate row in the `oc_tables_row_cells_usergroup`. So, when we want to update a a field, we should delete all previous values and insert to avoid duplicates. Problem was, we were assuming that the row only has one usergroup column, so we were deleting all usergroup cells for that row. But when there are multiple usergroup columns, only the last is kept since others are deleting. Instead, we should be deleting based on the row and column, not just the row.

### DIFF
--- a/lib/Db/RowCellMapperSuper.php
+++ b/lib/Db/RowCellMapperSuper.php
@@ -75,6 +75,22 @@ class RowCellMapperSuper extends QBMapper {
 	/**
 	 * @throws Exception
 	 */
+	public function deleteAllForRowAndColumn(int $rowId, int $columnId): void {
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->tableName)
+			->where(
+				$qb->expr()->eq('row_id', $qb->createNamedParameter($rowId, IQueryBuilder::PARAM_INT))
+			)
+			->andWhere(
+				$qb->expr()->eq('column_id', $qb->createNamedParameter($columnId, IQueryBuilder::PARAM_INT))
+			);
+		$qb->executeStatement();
+	}
+
+
+	/**
+	 * @throws Exception
+	 */
 	public function deleteAllForColumn(int $columnId): void {
 		$qb = $this->db->getQueryBuilder();
 		$qb->delete($this->tableName)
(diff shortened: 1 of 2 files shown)


juliusknorr (approved):

blizzz (approved):

lib/Db/RowCellMapperSuper.php
```diff
@@ -72,6 +72,22 @@ public function deleteAllForRow(int $rowId): void {
...
+	 */
+	public function deleteAllForRowAndColumn(int $rowId, int $columnId): void {
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->tableName)
+			->where(
+				$qb->expr()->eq('row_id', $qb->createNamedParameter($rowId, IQueryBuilder::PARAM_INT))
+			)
+			->andWhere(
+				$qb->expr()->eq('column_id', $qb->createNamedParameter($columnId, IQueryBuilder::PARAM_INT))
+			);
```
juliusknorr, line 86: The column order may matter when using database indices, however in the case of this simple query it seems to work fine and still use the index that is `(column_id, row_id)`



```
EXPLAIN select * from oc_tables_row_cells_text where row_id = 1 and column_id = 2;

1	SIMPLE	oc_tables_row_cells_text	ref	IDX_3AC41548BE8E8ED583A269F2	IDX_3AC41548BE8E8ED583A269F2	8	const,const	1	
```

blizzz: Oh, good point. Maybe still better to turn it around? To be safe and be a good guide.

PR #1754, opened 2025-04-16
fix: Properly display teams in usergroup

### LINKED ISSUE #1751: Adding Teams to a row is broken
### Steps to reproduce

1. Create a table with a row which allows to select users, groups and teamsImage

2. Go back to the table and create a row where you select a team in the corresponding row

### Expected behavior

Result correctly shown in the table

### Actual behavior

While creating a row I can select a team (in this example "Sales Engieerung (2)" is a team):

Image

After clicking on save the table looks good (beside that the avatar for the team is not show correctly):

Image

After reload the team name changes into a cryptical, probably internal ID:

Image

### Tables app version

0.9.0

### Browser

_No response_

### Client operating system

_No response_

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

_No response_

enjeck:
fixes #1751

Now, it shows the team display name, not the ID. Plus, an icon is used to indicate it's a team:
image

### DIFF
--- a/lib/Db/RowCellUsergroupMapper.php
+++ b/lib/Db/RowCellUsergroupMapper.php
@@ -9,6 +9,7 @@ namespace OCA\Tables\Db;
 
 use OCA\Tables\Constants\UsergroupType;
 use OCA\Tables\Helper\CircleHelper;
+use OCA\Tables\Helper\GroupHelper;
 use OCP\IDBConnection;
 use OCP\IUserManager;
 use OCP\IUserSession;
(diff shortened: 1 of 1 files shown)


lib/Db/RowCellUsergroupMapper.php
```diff
@@ -34,7 +39,12 @@ public function applyDataToEntity(Column $column, RowCellSuper $cell, $data): vo
 	}
 
 	public function formatEntity(Column $column, RowCellSuper $cell) {
-		$displayName = $cell->getValueType() === 0 ? ($this->userManager->getDisplayName($cell->getValue()) ?? $cell->getValue()) : $cell->getValue();
+		$displayName = $cell->getValue();
+		if ($cell->getValueType() === UsergroupType::USER) {
+			$displayName = $this->userManager->getDisplayName($cell->getValue()) ?? $cell->getValue();
```
blizzz, line 44: fwiw groups also have displaynames

enjeck: Thanks!

PR #1764, opened 2025-04-20
refactor: optimized parseEntities for better efficiency

AIlkiv:
Optimised the tables/views view. The main optimisation is that for each cell we no longer get the CellMapper service. I also made a couple of minor improvements.

On my machine, for a table (4000 rows and 6 columns), I got the following results when opening the page.
Before: 45 seconds
After: 5 seconds


juliusknorr (approved):
Very nice optimization, thanks for your pull request

AIlkiv:
The mistakes are not related to my PR

PR #1765, opened 2025-04-22
fix: avoid marking empty link fields as invalid protocol

### LINKED ISSUE #1761: I can't add a row to an existing table when there is a field for URL/File and it is empty
### Steps to reproduce

1. Create a table
2. Put in some random non-mandatory columns
3. Put in a non-mandatory column with a type URL and File (and Notes)
4. Create a new row
5. In the modal window the Save is grayed out until you fill in the URL/File type of field with some value.

### Expected behavior

When the field is not mandatory don't block saving the row.

### Actual behavior

The Save button is grayed out and can't click. Can't save the row.

### Tables app version

0.9.1

### Browser

Brave

### Client operating system

Linux

### Operating system

Linux

### Web server

Apache

### PHP engine version

PHP 8.2

### Database

MySQL

### Additional info

NC 31.0.4

enjeck:
fixes #1761 

For an empty link field, we should return `true` when checking if the field has a valid URL protocol. Otherwise, even for non-mandatory fields, the "Save" button is not clickable if no input is provided. 
### Before (grayed out Save button, unable to save empty non-mandatory link field)
Screenshot from 2025-04-22 18-05-14

### After (able to save empty non-mandatory link field)
image

### DIFF
--- a/src/shared/components/ncTable/mixins/rowHelper.js
+++ b/src/shared/components/ncTable/mixins/rowHelper.js
@@ -39,7 +39,10 @@ export default {
 		},
 
 		isValidUrlProtocol(value) {
-			value = JSON.parse(value ?? '{}')
+			if (!value) {
+				return true
+			}
+			value = JSON.parse(value)
 			try {
 				const parsedUrl = new URL(value?.value)
 				return ALLOWED_PROTOCOLS.includes(parsedUrl.protocol)


src/shared/components/ncTable/mixins/rowHelper.js
```diff
@@ -39,7 +39,10 @@ export default {
 		},
 
 		isValidUrlProtocol(value) {
-			value = JSON.parse(value ?? '{}')
+			if (!value) {
```
blizzz, line 42: Would there be a necessity to also trim value up front?

blizzz (approved):
With a question.

biguenique:
Merged commit into main: does that means that the fix will be available at the next coming release (tag)?

PR #1774, opened 2025-04-25
ci(config): stop updates for 0.8

enjeck:
Tables 0.8 series is only relevant for NC 28 anymore. We do a last release of it at https://github.com/nextcloud/tables/pull/1773, and otherwise keep it low (i.e. disabling dependency updates and such), unless there's an issue targeting that version.


blizzz:
> @blizzz Should we have a stable0.9 now?

Only when we have to branch off from main for a reason.

PR #1789, opened 2025-04-28
build(deps): bump phpoffice/phpspreadsheet

blizzz:
Replaces #1786 

We are bound to 2.1.x as newer versions require PHP 8.1. 

But soon we can drop 29, branch off 0.9 and bump to 0.10, and with it NC and PHP requirements.


blizzz:
Import scenarios are failing :melting_face:

blizzz:
> Import scenarios are failing 🫠

Proposed fix upstream at https://github.com/PHPOffice/PhpSpreadsheet/pull/4476

PR #1798, opened 2025-05-04
feat: Enhance "Contains" filter to support "selection" type

AIlkiv:
I think this filter will improve the UX. Because now you don't have to remember the exact value.

image


blizzz (approved):
Tested, works

AIlkiv:
@enjeck  please look at this MR.

PR #1806, opened 2025-05-20
chore(tests): Cleanup bootstrap.php to be forward-compatible

come-nc:
This uses the new tests/autoload.php from nextcloud/server instead of messing directly with OC private autoloader.

See https://github.com/nextcloud/server/pull/52951 and https://github.com/nextcloud/server/pull/52945 for context.

### DIFF
--- a/tests/unit/bootstrap.php
+++ b/tests/unit/bootstrap.php
@@ -1,19 +1,27 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * SPDX-FileCopyrightText: 2016 Nextcloud GmbH and Nextcloud contributors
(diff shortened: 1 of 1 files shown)


come-nc:
It looks like the cypress tests are using circles master branch with server stable* branches, which are not compatible.
It should pull the matching circles branch.

blizzz (approved):

blizzz:
> It looks like the cypress tests are using circles master branch with server stable* branches, which are not compatible. It should pull the matching circles branch.

Addressed in https://github.com/nextcloud/tables/pull/1809

PR #1808, opened 2025-05-21
perf: Parallelize initial requests

enjeck:
Instead of calling some initial request sequentially, we parallelize the calls. As a result, we save >1 second during the initial load

### Before
Screenshot from 2025-05-21 05-46-15

### After
image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -54,10 +54,12 @@ export default {
 	},
 	async created() {
 		const store = useTablesStore()
-		await store.loadTablesFromBE()
-		await store.getAllContexts()
-		await store.loadViewsSharedWithMeFromBE()
-		await store.loadTemplatesFromBE()
+		await Promise.all([
+			store.loadTablesFromBE(),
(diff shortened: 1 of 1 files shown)


juliusknorr (approved):
This seems like a good low hanging fruit to make the loading faster.

I'd merge this, but would raise some ideas for discussion (maybe also just briefly in our call):

- Do we need to await the template list at all? I could imagine we just fire of the call without needing to wait for the promise to resolve
- Alternatively since the template list is a static array, the overhead to have an extra request seems rather high, it could be passed in as initial state on the page request
- We could think about having one backend route that lists all tables, views and contexts in one response

juliusknorr:
And don't forget to update the metadata like labels on prs ;)

juliusknorr:
> Alternatively since the template list is a static array, the overhead to have an extra request seems rather high, it could be passed in as initial state on the page request

As discussed this might be a good exercise to learn the initial state API we have, could you give that a try in a follow up?

PR #1809, opened 2025-05-21
fix: Update circles in cypress workflow

enjeck:


blizzz:
The circles thing is fixed, yet there are failing tests against both 31 and master.

juliusknorr (approved):

blizzz:
Awesome @enjeck :medal_sports:

PR #1812, opened 2025-05-21
fix: remove unused empty-content class

blizzz:
This PR is mainly to double-check that the integration-summary is passing, not blocking. Hence based on branch `ci/noid/integration-summary-unblock`.

Otherwise, the IDE said it is not being used and it really might be, but unsure :D can close, can take it nevertheless.


blizzz:
accidental merge :see_no_evil:

blizzz:
> accidental merge 🙈

Removed from the base, and replaced with https://github.com/nextcloud/tables/pull/1814

PR #1815, opened 2025-05-21
[stable0.8] refactor(Controller): remove private ColumnsController

blizzz:
… and use the Api1 instead. This one is also involved in integration tests while the ColumnsController not and could have only be used in the front- end anyways.


juliusknorr (approved):

src/store/data.js
```diff
@@ -136,7 +135,7 @@ export default {
 		},
 		async removeColumn({ state, commit }, { id, isView, elementId }) {
 			try {
-				await axios.delete(generateUrl('/apps/tables/column/' + id))
```
kieuduyan, line 139: await axios.delete(generateUrl('/apps/table/column/' + id));

lib/Controller/Api1Controller.php
```diff
@@ -922,7 +922,6 @@ public function updateColumn(
 		try {
 			$item = $this->columnService->update(
 				$columnId,
-				null,
```
kieuduyan, line 925: null(),

kieuduyan (commented):
ok

PR #1824, opened 2025-05-27
Enhancement: Update rows via import

### LINKED ISSUE #1403: Ewport/import data duplicate row instead of updating it (even with "ID" column left untouched)
### Steps to reproduce

1. create a table and a view with "ID" column selected
2. Export all row of this view
3. Edit it on your computer to modify data on one or many row (no change on columns)
4. import the modified file into the view

### Expected behavior

1. The row should just update and not duplicate, (because the "ID" column is left untouched and is unique per row)
2. The second column "ID" should not be created

### Actual behavior

1. Every row is now duplicate, half is the old version, half is the new version
2. Also a new column is automatically created (a ID column), even if the choice "create missing column" is not choosen

### Tables app version

0.7.5

### Browser

firefox

### Client operating system

windows

### Operating system

Debian

### Web server

Nginx

### PHP engine version

None

### Database

PostgreSQL

### Additional info

Nextcloud 29.0.5

The reason why i want to do this:
I want to modify a lots of row quickly, and doing it in Table, editing row one by one is just toooooooo long.
Right now, the only solution i found is to export data, delete it, modify it on my computer, and import it again.

Koc:
This PR adds possibility to update rows via import. Closes #1403


lib/Service/ImportService.php
```diff
@@ -426,26 +455,36 @@ private function createRow(Row $row): void {
...
-				$this->countInsertedRows++;
-			} else {
+			if (!$hasData) {
 				$this->logger->debug('Skipped empty row ' . $row->getRowIndex() . ' during import');
+				return;
+			}
+
+			if ($id) {
+				try {
+					$this->rowService->updateSet($id, $this->viewId, $data, $this->userId);
```
AIlkiv: I think there is a need to add a tableId check to updateSet. Otherwise, it will be possible to update values from table B when importing table A.

lib/Service/ImportService.php
```diff
@@ -426,26 +455,36 @@ private function createRow(Row $row): void {
...
-				$this->rowService->create($this->tableId, $this->viewId, $data);
-				$this->countInsertedRows++;
-			} else {
+			if (!$hasData) {
 				$this->logger->debug('Skipped empty row ' . $row->getRowIndex() . ' during import');
+				return;
+			}
+
+			if ($id) {
+				try {
```
AIlkiv: I think this try/cach is unnecessary, you should catch the exception in the existing try/catch

lib/Service/ImportService.php
Koc: Good catch! Fixed

lib/Service/ImportService.php
Koc: no, it's inconvenient because in case if update fails - we should fallback to create

Koc:
Any ideas why new not pass for NC 29?

```
    Given file "update-rows.csv" exists for user "participant1" with the following data    # FeatureContext::createCsvFile()
      | ID      | one         | two |
      | {rowId} | AHA updated | 99  |
      |         | new row     | 100 |
      Failed asserting that 401 matches expected 201.
```

at the same time it works fine for all other versions. I've inspired from already existent `Given user :user uploads file :file` which works on NC29 as well

AIlkiv (approved):

Koc:
@enjeck I've added Cypress test for this new functionality

Component tests failure is not related to my changes. It fails [everywhere](https://github.com/nextcloud/tables/actions/runs/16700955720/job/47272017966).

Koc:
@enjeck good catch!
I've fixed this issue. Now it works properly for both scenarios: create new column and insert rows duplicates; reuse id column and use id for update

PR #1827, opened 2025-05-30
Fix: Fix Cypress tests

Koc:
As you can see - many of the recent PRs have failed pipelines due to Cypress. This PR tries to fix that. Not the most elegant way but I hope it works :crossed_fingers:.

### DIFF
--- a/cypress/e2e/tables-table.cy.js
+++ b/cypress/e2e/tables-table.cy.js
@@ -67,9 +67,10 @@ describe('Manage a table', () => {
 		cy.get('[data-cy="customTableAction"] button').click()
 		cy.get('.action-button__text').contains('Edit table').click()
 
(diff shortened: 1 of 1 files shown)


cypress/e2e/tables-table.cy.js
```diff
@@ -67,9 +67,10 @@ describe('Manage a table', () => {
 		cy.get('[data-cy="customTableAction"] button').click()
 		cy.get('.action-button__text').contains('Edit table').click()
 
-		cy.get('[data-cy="editTableModal"]').should('be.visible')
+		cy.wait(5).get('[data-cy="editTableModal"]').should('be.visible')
```
enjeck, line 70: We usually avoid having `wait`(s), but if this is the only way to stop the tests from being flaky, I'm okay with it.

PR #1831, opened 2025-06-02
 Add "Internal Link" Sharing Tile with Copy-to-Clipboard

silverkszlo:
This PR introduces a new sharing option component called "Internal Link", which allows users to copy a direct link to the currently selected table or view. This feature is useful for quickly sharing internal resources with users who already have access to the folder.

Features:
- New ShareInternalLink.vue component
- Displays an icon, title, and description for internal links
- Includes a copy-to-clipboard button using NcActionButton and ContentCopy icon
- Localized (t('tables', ...)) and translations for English and German added

Screenshot from 2025-06-03 08-08-08

### DIFF
--- a/src/modules/sidebar/partials/ShareInternalLink.vue
+++ b/src/modules/sidebar/partials/ShareInternalLink.vue
@@ -1,29 +1,33 @@
+<!--
+  - SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
 <template>
-  <div class="share-internal-link">
-    <div class="share-internal-link__icon">
-      <OpenInNew :size="20" />
-    </div>
+	<div class="share-internal-link">
+		<div class="share-internal-link__icon">
+			<OpenInNew :size="20" />
+		</div>
 
-    <div class="share-internal-link__info">
-      <span class="share-internal-link__title">
-        {{ t('tables', 'Internal link') }}
-      </span>
-      <span class="share-internal-link__subtitle">
-        {{ t('tables', 'Only works for users with access to this folder') }}
-      </span>
-    </div>
+		<div class="share-internal-link__info">
+			<span class="share-internal-link__title">
+				{{ t('tables', 'Internal link') }}
+			</span>
+			<span class="share-internal-link__subtitle">
+				{{ t('tables', 'Only works for users with access to this folder') }}
+			</span>
+		</div>
 
-    <NcButton
-      :title="copyTooltip"
-      :aria-label="copyTooltip"
-      class="share-internal-link__button"
-      @click="copyUrl"
-      variant="tertiary">
-      <template #icon>
-        <ContentCopy :size="20" />
-      </template>
-    </NcButton>
-  </div>
+		<NcButton
+			:title="copyTooltip"
+			:aria-label="copyTooltip"
+			class="share-internal-link__button"
+			variant="tertiary"
+			@click="copyUrl">
+			<template #icon>
+				<ContentCopy :size="20" />
+			</template>
+		</NcButton>
+	</div>
 </template>
 
 <script>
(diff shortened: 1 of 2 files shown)


src/modules/sidebar/partials/ShareInternalLink.vue
```diff
@@ -0,0 +1,117 @@
...
+    <div class="share-internal-link__info">
+      <span class="share-internal-link__title">
+        {{ t('tables', 'Internal link') }}
+      </span>
+      <span class="share-internal-link__subtitle">
+        {{ t('tables', 'Only works for users with access to this folder') }}
+      </span>
+    </div>
+
+    <NcActionButton :title="copyTooltip" :aria-label="copyTooltip" @click="copyUrl" class="share-internal-link__button">
```
juliusknorr: NcActionButton is supposed to be only used within an NcActions component, probably worth to switch to https://nextcloud-vue-components.netlify.app/#/Components/NcButton which could save some custom styling

src/modules/sidebar/sections/SidebarSharing.vue
```diff
@@ -38,6 +41,13 @@ export default {
 
 	computed: {
 		...mapState(useTablesStore, ['activeElement', 'isView']),
+		currentUrl() {
+			const url = this.activeElement
```
juliusknorr: Something Nextcloud specific, we should use [generateUrl](https://nextcloud-libraries.github.io/nextcloud-router/functions/generateUrl.html) here to make sure this also works properly if Nextcloud was installed in a subdirectory.

src/modules/sidebar/partials/ShareInternalLink.vue
```diff
@@ -0,0 +1,117 @@
...
+  display: flex;
+  align-items: center;
+  min-height: 44px;
+  gap: 12px;
+
+  &__icon {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 32px;
```
juliusknorr: This is not very present in the tables app code base yet, but we can make more use of css variables here:

```suggestion
    width: var(--default-clickable-area);
```

https://docs.nextcloud.com/server/latest/developer_manual/html_css_design/css.html#element-structure-variables

src/modules/sidebar/sections/SidebarSharing.vue
```diff
@@ -38,6 +41,13 @@ export default {
...
+				? `${window.location.origin}/index.php/apps/tables/#/table/${this.activeElement.id}`
```
juliusknorr: We have two different URLs that could be shared from this sidebar, for tables and views. views would be a subset of a table that then have `#/view/:viewId` as their route.

You should be able to test this using the default table and the `Checkyourself!` example view

Screenshot 2025-06-03 at 09 26 52

juliusknorr (changes requested):
Thanks a lot for the pull request, this looks very good already. I've left a few minor comments on the code itself and one note about a functional issue. Could you have another quick look at those?

l10n/de_DE.js
```diff
@@ -383,6 +383,9 @@ OC.L10N.register(
     "Create data" : "Daten erstellen",
     "Update data" : "Daten aktualisieren",
     "Delete data" : "Daten löschen",
+    "Internal Link" : "Interner Link",
```
juliusknorr: Forgot to mention, the commit with translations can be dropped. We sync those separately to [transifex](https://explore.transifex.com/nextcloud/nextcloud/) and pull updated translations from there nightly, so they would be dropped on the next sync until they are translated over there.

juliusknorr:
Thanks a lot. Could you have a look at the failing REUSE and eslint checks? Those two seem related to the changes, the other two can be ignored for now.

src/modules/sidebar/sections/SidebarSharing.vue
```diff
@@ -38,6 +42,17 @@ export default {
 
 	computed: {
 		...mapState(useTablesStore, ['activeElement', 'isView']),
+		currentUrl() {
+			if (!this.activeElement) {
+				return ''
+			}
+			const baseUrl = generateUrl('/apps/tables/', {}, { baseURL: window.location.origin })
```
juliusknorr:
```suggestion
			const baseUrl = generateUrl('/apps/tables/')
```

silverkszlo: In more or less two hours I'll have time to work on that and the failing checks!

vendor-bin/openapi-extractor/composer.lock
enjeck, line 1: Don't think we should have any changes here?

src/modules/sidebar/sections/SidebarSharing.vue
silverkszlo: I used `baseURL: window.location.origin` when calling generateUrl to ensure the generated URL works correctly in my local development environment, where relative paths like /index.php/... were resolving incorrectly (e.g., resulting in a file:/// URL instead of http://localhost).

PR #1844, opened 2025-06-19
feat: add last modified information to response

blizzz:
NotModifiedMiddleware respects last modified information stored with a data response. The business logic still runs, but at least clients will not need to do any heavy lifting, as only 304 is returned if the condition pertains.

### DIFF
--- a/lib/Controller/Api1Controller.php
+++ b/lib/Controller/Api1Controller.php
@@ -1257,7 +1257,12 @@ class Api1Controller extends ApiController {
 	#[OpenAPI(scope: OpenAPI::SCOPE_DEFAULT)]
 	public function indexTableRows(int $tableId, ?int $limit, ?int $offset): DataResponse {
 		try {
-			return new DataResponse($this->rowService->formatRows($this->rowService->findAllByTable($tableId, $this->userId, $limit, $offset)));
+			$rows = $this->rowService->findAllByTable($tableId, $this->userId, $limit, $offset);
+			$response = new DataResponse($this->rowService->formatRows($rows));
+			$table = $this->tableService->find($tableId);
+			$lastModified = new \DateTime($table->getLastEditAt());
+			$response->setLastModified($lastModified);
+			return $response;
 		} catch (PermissionError $e) {
 			$this->logger->warning('A permission error occurred: ' . $e->getMessage(), ['exception' => $e]);
 			$message = ['message' => $e->getMessage()];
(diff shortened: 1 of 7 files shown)


lib/Controller/Api1Controller.php
```diff
@@ -1084,7 +1084,12 @@ public function indexTableRowsSimple(int $tableId, ?int $limit, ?int $offset): D
 	#[RequirePermission(permission: Application::PERMISSION_READ, type: Application::NODE_TYPE_TABLE, idParam: 'tableId')]
 	public function indexTableRows(int $tableId, ?int $limit, ?int $offset): DataResponse {
 		try {
-			return new DataResponse($this->rowService->formatRows($this->rowService->findAllByTable($tableId, $this->userId, $limit, $offset)));
+			$rows = $this->rowService->findAllByTable($tableId, $this->userId, $limit, $offset);
+			$response = new DataResponse($this->rowService->formatRows($rows));
+			$table = $this->tableService->find($tableId);
+			$lastModified = new \DateTime($table->getLastEditAt());
```
juliusknorr: Are we sure that any change of rows would update the getLastEditAt date? Otherwise this would not have much benefit i think.

lib/Controller/Api1Controller.php
```diff
@@ -1084,7 +1084,12 @@ public function indexTableRowsSimple(int $tableId, ?int $limit, ?int $offset): D
 	#[RequirePermission(permission: Application::PERMISSION_READ, type: Application::NODE_TYPE_TABLE, idParam: 'tableId')]
 	public function indexTableRows(int $tableId, ?int $limit, ?int $offset): DataResponse {
 		try {
-			return new DataResponse($this->rowService->formatRows($this->rowService->findAllByTable($tableId, $this->userId, $limit, $offset)));
+			$rows = $this->rowService->findAllByTable($tableId, $this->userId, $limit, $offset);
+			$response = new DataResponse($this->rowService->formatRows($rows));
+			$table = $this->tableService->find($tableId);
```
juliusknorr: We could move this line first and return a 304 early in case the if-modified-since matches, before doing the possibly heavier operations of fetching the rows

lib/Controller/Api1Controller.php
blizzz: I did not double check it yet, it is merely my expectation. I was wondering whether this is also updated, when a row is being deleted.

lib/Controller/Api1Controller.php
blizzz: Yes, definitely. First I thought i had to analyze the rows that were fetched, then i was reminded about the last modificated set in the table (if that really works as i think). I pushed this to not have it rotting away locally and turning to legend in first place :slightly_smiling_face:

enjeck (commented):
I'm gonna take over this

lib/Controller/Api1Controller.php
enjeck: Handled now. `TableMapper::touch()` updates `last_edit_at`, and `RowService` calls it on every row mutation, so inserts, updates, and deletes all bump the date. So yes, deleting a row updates it too. Column `create` and `update` also touch the table now, since the rows response includes alias data derived from the columns.

Issue #1769, opened 2025-04-24
Views: set mandatory-state of a column


blizzz:
When using **views** for different stages of processes, some columns might deviate from the original table definition and turn mandatory.  

- [x] Needs: https://github.com/nextcloud/tables/pull/1688
- [ ] BE: handle storing, retrieving and updating the *mandatory* flag with the dedicated column content. 
- [ ] BE: the create and update logic of rows in views need to take this setting into consideration
- [ ] FE: together with https://github.com/nextcloud/tables/issues/1154#issuecomment-2828168145 agree on an UI for such elements (this step does not need to wait for 1688.)
- [ ] FE: implementation of the results ^ and of course related business logic.

PR #1845, opened 2025-06-19
refactor: turn columnSettings data from loose array to value object

blizzz:
The aim is to make it easier to add properties like [read-only](https://github.com/nextcloud/tables/issues/1154) or [mandatory](https://github.com/nextcloud/tables/issues/1769) to user columns. Having the structure in a separate class makes it easier to organize it (no loose array structure flying around), to extend it and also to enforce the structure.

I thought creating a PR for the read-only feature first, but I think it is already a good size for a PR on its own.


AIlkiv:
@blizzz 
What do you think about this idea, instead of ViewColumnInformation, create ViewColumnsInformation (or just ViewColumns). Encapsulate all related logic (sorting, decoding, etc.) in this class.


For example instead of:
```
$decodedValue = \json_decode($value, true);
$value = [];
foreach ($decodedValue as $order => $id) {
	$value[] = [
		'columnId' => $id,
		'order' => $order
	];
```
Do this:
```
$value = ViewColumnsInformation::decode($value): ViewColumnsInformation
```

blizzz:
Hi @AIlkiv, thank you for the suggestion. I think then I would still leave the `ViewColumnInformation` as is, really as value for each column. It makes sense to have a collection – leaning on your suggestion – on top that provides this utility methods and can hold the values. I am not sure about the balance between real benefit and another layer though.

blizzz:
> Hi @AIlkiv, thank you for the suggestion. I think then I would still leave the `ViewColumnInformation` as is, really as value for each column. It makes sense to have a collection – leaning on your suggestion – on top that provides this utility methods and can hold the values. I am not sure about the balance between real benefit and another layer though.

OK, went through, but I did not see that many benefits really. Would keep it as option for the future though (or maybe it is late).

lib/Service/TableTemplateService.php
```diff
@@ -466,11 +467,8 @@ private function makeVacationRequests(Table $table):void {
 			[
 				'title' => $this->l->t('Create Vacation Request'),
 				'emoji' => '️➕',
-				'columnSettings' => json_encode(array_map(function ($columnId, $index) {
-					return [
-						'columnId' => $columnId,
-						'order' => $index
-					];
+				'columnSettings' => json_encode(array_map(function (int $columnId, int $index): ViewColumnInformation {
+					return new ViewColumnInformation($columnId, $index);
```
AIlkiv: for better readability
`order: $index`

lib/Service/TableTemplateService.php
```diff
@@ -480,11 +478,8 @@ private function makeVacationRequests(Table $table):void {
 			[
 				'title' => $this->l->t('Open Request'),
 				'emoji' => '️📝',
-				'columnSettings' => json_encode(array_map(function ($column, $index) {
-					return [
-						'columnId' => $column->getId(),
-						'order' => $index
-					];
+				'columnSettings' => json_encode(array_map(function (Column $column, int $index): ViewColumnInformation {
+					return new ViewColumnInformation($column->getId(), $index);
```
AIlkiv: for better readability
`order: $index`

lib/Service/TableTemplateService.php
```diff
@@ -508,11 +500,8 @@ private function makeVacationRequests(Table $table):void {
...
 				'title' => $this->l->t('Closed requests'),
 				'emoji' => '️✅',
-				'columnSettings' => json_encode(array_map(function ($column, $index) {
-					return [
-						'columnId' => $column->getId(),
-						'order' => $index
-					];
+				'columnSettings' => json_encode(array_map(function (Column $column, int $index): ViewColumnInformation {
+					return new ViewColumnInformation($column->getId(), $index);
 				}, array_values($columns), array_keys(array_values($columns)))),
```
AIlkiv:
```
json_encode(array_map(function (Column $column, int $index): ViewColumnInformation {
					return new ViewColumnInformation($column->getId(), $index);
				}, array_values($columns), array_keys(array_values($columns))))
```

It is possible to move these three lines to a new private method

lib/Db/View.php
```diff
@@ -93,14 +94,14 @@ public function __construct() {
 	public function getColumnsArray(): array {
 
 		$columnSettings = $this->getColumnsSettingsArray();
-		usort($columnSettings, function ($a, $b) {
-			return $a['order'] - $b['order'];
+		usort($columnSettings, static function (ViewColumnInformation $a, ViewColumnInformation $b) {
+			return $a[ViewColumnInformation::KEY_ORDER] - $b[ViewColumnInformation::KEY_ORDER];
```
AIlkiv: To make the code cleaner (type hints, etc.), it might be better to use a getter/setter. Remove the constants

$a[ViewColumnInformation::KEY_ORDER] - $b[ViewColumnInformation::KEY_ORDER]

VS

$a->getOrder() - $b->getOrder()

blizzz:
Thank you @AIlkiv

PR #1720, opened 2025-04-03
feat: Add usergroup column filters

enjeck:
Fixes https://github.com/nextcloud/tables/issues/1547

PR #1866, opened 2025-06-30
Enhancement: Improve filtering by groups and teams

Koc:
This is a continuation of #1720

1. Create column of type "Users and groups"
2. Create a view with filtered column, use "me" for filter value and "contains" operator type
    <details><summary>:mag: Preview</summary>
    <p>

    image

    </p>
    </details>

4. Add few rows with various users, groups and teams
    <details><summary>:mag: Preview</summary>
    <p>

    image

    </p>
    </details>

5. Share view to another user
6. Open view by another user

Expected result: user is able to see rows which met to criteria
Actual result: this filter works properly only for users and not works for groups and teams

<details><summary>:mag: Preview with fix</summary>
<p>

image

</p>
</details>

### DIFF
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -8,6 +8,7 @@
 namespace OCA\Tables\Db;
 
 use DateTime;
+use OCA\Tables\Constants\UsergroupType;
 use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Helper\ColumnsHelper;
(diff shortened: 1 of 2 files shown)


lib/Db/Row2Mapper.php
```diff
@@ -401,6 +401,30 @@ private function getFilterExpression(IQueryBuilder $qb, Column $column, string $
 				$filterExpression = $qb->expr()->like('value', $qb->createNamedParameter('%' . $this->db->escapeLikeParameter($value), $paramType));
 				break;
 			case 'contains':
+				$filterExpressions = [];
+				if (is_array($value) && $column->getType() === Column::TYPE_USERGROUP) {
+					$filterExpressions[] = $qb2->expr()->andX(
+						$qb->expr()->eq('value', $qb->createNamedParameter($value[UsergroupType::USER])),
```
enjeck, line 407: could we potentially get an undefined index error is the keys are not in `$value`??

Koc: It shouldn't be possible. `$value` can be a primitive instead of array, but in this case we are not entering into a condition. Otherwise we're always adding item with key `UsergroupType::USER`, see 

image

Issue #941, opened 2024-03-18
Performance: Evaluate options to make the table/view row listing paginated


juliushaertl:

blizzz:
## Tables

For tables there are already endpoints on the Api1 Controller that support pagination:

```
$ ack indexTableRows
appinfo/routes.php
46:             ['name' => 'api1#indexTableRowsSimple', 'url' => '/api/1/tables/{tableId}/rows/simple', 'verb' => 'GET'],
47:             ['name' => 'api1#indexTableRows',       'url' => '/api/1/tables/{tableId}/rows', 'verb' => 'GET'],

lib/Controller/Api1Controller.php
909:    public function indexTableRowsSimple(int $tableId, ?int $limit, ?int $offset): DataResponse {
939:    public function indexTableRows(int $tableId, ?int $limit, ?int $offset): DataResponse {
```

It seems to be unused by the frontend though, there `RowController::index()` is being called. Might be even a drop-in replacement. 

## Views

The same is acutally valid for Views. The Api1 Controller supports pagination:

```
$ ack indexViewRows 
appinfo/routes.php
48:             ['name' => 'api1#indexViewRows',        'url' => '/api/1/views/{viewId}/rows', 'verb' => 'GET'],

lib/Controller/Api1Controller.php
969:    public function indexViewRows(int $viewId, ?int $limit, ?int $offset): DataResponse {
```

but `RowController::indexView` is used. 

## Conclusio

If the frontend  cannot use the routes of Ap1Controller, we can easily add the optional arguments as the same service methods are being used..

blizzz:
### Related open tasks:

* https://github.com/nextcloud/tables/issues/1506
* https://github.com/nextcloud/tables/issues/1507
* (WIP)

AIlkiv:
@blizzz 
Please take a look at my pull request. It significantly speeds up the server's response.

https://github.com/nextcloud/tables/pull/1764

PR #1877, opened 2025-07-03
fix: sorting on backend with limit and offset

AIlkiv:
In context: #941

To make sorting with limit/offset work correctly, we need to sort in the same query that imposes the limit.


AIlkiv:
@enjeck 

image

Example:

| Id | Name    |
|---|---------|
| 1 | Alice   |
| 2 | Bob     |
| 3 | Charlie |
| 4 | Diana   |

sort: Name Desc
limit: 3

**Before:**

First step:
```
$wantedRowIdsArray = $this->getWantedRowIds($userId, $tableId, $filter, $limit, $offset);
```

return 1,2,3

Second step:

```
$this->getRows($wantedRowIdsArray, $showColumnIds, $sort ?? []);
```

Sort only rows: 1,2,3

And return: Charlie, Bob, Alice

**After:**

```
$wantedRowIdsArray = $this->getWantedRowIds($userId, $tableId, $filter, $sort, $limit, $offset);
```

return 4,3,2

```
$rows = $this->getRows($wantedRowIdsArray, $showColumnIds);
```
return: Bob, Charlie, Diana

```
$this->sortRowsByIds($rows, $wantedRowIdsArray);
```

return: Diana, Charlie, Bob

---

| Before | After    |
|---|---------|
| Charlie, Bob, Alice | Diana, Charlie, Bob |

---

I also added tests. I found several bugs:

1. When sorting is configured by column, and then the column is deleted. Previously, this crashed when receiving a list of available views.
2. I found that sorting by selection, textlink, and usergroup does not work, so I hid the ability to sort by this type of column.

AIlkiv:
On the local build, the tests “Cypress Component Tests / component-tests (20)” pass.

src/modules/main/partials/editViewPartials/sort/SortForm.vue
```diff
@@ -114,9 +115,17 @@ export default {
...
+				ColumnTypes.SelectionMulti,
+				ColumnTypes.TextLink,
+				ColumnTypes.Usergroup,
+			].includes(col.type)
+		},
 		eligibleColumns(selectedId) { // filter sortable and unused columns
-			if (this.hadHiddenSortingRules || !this.viewSort) return this.columns?.filter(col => col.canSort())
-			return this.columns.filter(col => col.canSort() && (!this.viewSort.map(entry => entry.columnId).includes(col.id) || col.id === selectedId))
+			if (this.hadHiddenSortingRules || !this.viewSort) return this.columns?.filter(col => col.canSort() && this.canBeSorted(col))
+			return this.columns.filter(col => col.canSort() && this.canBeSorted(col) && (!this.viewSort.map(entry => entry.columnId).includes(col.id) || col.id === selectedId))
```
enjeck, line 136: I don't understand why we need `canBeSorted()`?  Doesn't `canSort()` already check that a specific column type is sortable?

AIlkiv: @enjeck 
Not all types that can sort on the front-end can sort on the back-end.

A simple example of "selection": on the front-end, we have a separate sorting function where the value is present. On the back-end, for "selection", we only have the ID, and there is no way to neatly  (via JOIN) pull the value into the query that needs to be sorted.

Ideally, this should be added, but it requires a lot of refactoring.

enjeck: I see. Can we please add a comment to the function explaining this? I would add it myself but I can't push to your branch

AIlkiv: @enjeck I added a comment

enjeck: Thanks

Issue #1850, opened 2025-06-22
Error while creating column


enjeck:
### Steps to reproduce

1. Try to create column in existing table
2.
3.

### Expected behavior

Creates without any error

### Actual behavior

There's an error. New column does appear upon refresh though

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

Log:

```json 
{
  "reqId": "RE7nb4gZfuGuVv0c1ZyL",
  "level": 3,
  "time": "2025-06-22T16:34:06+00:00",
  "remoteAddr": "192.168.21.5",
  "user": "admin",
  "app": "tables",
  "method": "POST",
  "url": "/index.php/apps/tables/api/1/columns",
  "message": "An internal error or exception occurred: An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'owner_display_name' in 'SET'",
  "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36",
  "version": "32.0.0.0",
  "exception": {
    "Exception": "OCA\\Tables\\Errors\\InternalError",
    "Message": "An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'owner_display_name' in 'SET'",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/html/apps-extra/tables/lib/Service/ColumnService.php",
        "line": 233,
        "function": "addColumnToView",
        "class": "OCA\\Tables\\Service\\ViewService",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Tables\\Db\\View",
            "id": "*** sensitive parameters replaced ***"
          },
          {
            "__class__": "OCA\\Tables\\Db\\Column",
            "id": 5
          },
          "*** sensitive parameters replaced ***"
        ]
      },
      {
        "file": "/var/www/html/apps-extra/tables/lib/Controller/Api1Controller.php",
        "line": 830,
        "function": "create",
        "class": "OCA\\Tables\\Service\\ColumnService",
        "type": "->",
        "args": [
          "*** sensitive parameters replaced ***"
        ]
      },
      {
        "file": "/var/www/html/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 202,
        "function": "createColumn",
        "class": "OCA\\Tables\\Controller\\Api1Controller",
[...]

PR #1881, opened 2025-07-05
fix: Error while creating column

AIlkiv:
fix: #1850


lib/Db/EntitySuper.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace OCA\Tables\Db;
+
+use OCP\AppFramework\Db\Entity;
+
+class EntitySuper extends Entity {
+	protected const VIRTUAL_PROPERTIES = [];
+
+	protected function markFieldUpdated(string $attribute): void {
+		if (in_array($attribute, static::VIRTUAL_PROPERTIES)) {
```
blizzz:
```suggestion
		if (in_array($attribute, static::VIRTUAL_PROPERTIES, true)) {
```

AIlkiv: done

blizzz:
Thank you @AIlkiv!

PR #1887, opened 2025-07-08
chore: split value and ID parsing into separate methods

AIlkiv:
* Moved the logic of value parsing and display value parsing to separate methods (parseValue / parseDisplayValue and canBeParsed / canBeParsedDisplayValue) for the SelectionBusiness type.
* Updated calls in ImportService to use the new methods for display value.
* Added default implementations in SuperBusiness to maintain backward compatibility.
* Improved readability and separation of responsibilities in classes, which simplifies future support and extensions.

These changes allow for a clear separation of ID parsing and display value parsing, which is important for correct import and validation of data in tables.


AIlkiv:
> Do we need to update
> 
> https://github.com/nextcloud/tables/blob/e761f166d829bc695b8648f013027382c71edb80/lib/Service/RowService.php#L315-L317
> 
> too to use the new functions?

@enjeck  No, because in this place for type selection there is already a value id

PR #1898, opened 2025-07-11
Bump to 1.0.0, drop NC 29 support, bump PHP req to 8.1

blizzz:


blizzz:
partially networking, but openapi needs luv

blizzz:
Mh, openapi executes successfully locally :thinking:

PR #1903, opened 2025-07-12
chore: enable local PHPUnit runs with SQLite support

AIlkiv:
This PR allows you to run PHPUnit tests locally with SQLite in-memory support, which greatly simplifies development and debugging, since this module has a very large amount of business logic related to the database.

- Added a new test:unit:local script to composer.json to run tests locally
- Created a base class DatabaseTestCase for tests that require working with a database
- Implemented in-memory SQLite support for fast testing without the need to set up a full database
- Added tests covering https://github.com/nextcloud/tables/pull/1881

# Run tests locally from SQLite
composer test:unit:local

# Run tests as usual (as before)
composer test:unit


lib/Db/RowCellDatetime.php
```diff
@@ -10,9 +10,8 @@
 /** @template-extends RowCellSuper<RowCellDatetime> */
 class RowCellDatetime extends RowCellSuper {
 	protected ?string $value = null;
-	protected ?int $valueType = null;
 
 	public function jsonSerialize(): array {
-		return parent::jsonSerializePreparation($this->value, $this->valueType);
+		return parent::jsonSerializePreparation($this->value, 0);
```
enjeck: nit: i personally think it's less confusing to maintain the value as `null`, even if it gets eventually converted to `0`.

tests/unit/Db/EntityVirtualPropertiesTest.php
```diff
@@ -0,0 +1,355 @@
...
+	 * Check if class extends Entity or EntitySuper
+	 *
+	 * @param string $className Fully qualified class name
+	 * @return bool True if class extends Entity or EntitySuper
+	 */
+	private function isEntityClass(string $className): bool {
+		try {
+			$reflection = new ReflectionClass($className);
+
+			// Перевірити, що клас не абстрактний
```
enjeck: English, please

lib/Db/RowCellSuper.php
```diff
@@ -34,8 +34,6 @@ abstract class RowCellSuper extends Entity implements JsonSerializable {
 	protected ?int $rowId = null;
 	protected ?string $lastEditBy = null;
 	protected ?string $lastEditAt = null;
-	protected ?string $createdBy = null;
-	protected ?string $createdAt = null;
```
enjeck, line 38: we should remove their getters and setters above too?

tests/unit/Db/EntityVirtualPropertiesTest.php
```diff
@@ -0,0 +1,355 @@
...
+				$parentClass = $parentClass->getParentClass();
+			}
+		} catch (\Exception $e) {
+			// If we can't get VIRTUAL_PROPERTIES, return empty array
+		}
+
+		return [];
+	}
+
+	public function columnToProperty(string $columnName) {
```
enjeck: should this be private too?

PR #1916, opened 2025-07-18
enh: add read-only property to view columns (backend only)

blizzz:
contributes to #1154 

Currently work in progress. 

The approach is here to to still continue with the save operation and only leave out a provided column value if it is set to read-only instead to throw an error for the whole request. 

- [ ] needs tests
- [ ] need to check for completeness


blizzz:
> Mate, I've took your PR as an inspiration point and implemented whole functionality in #1958. Please check when you have time 🙏

Hey @Koc  thank you for picking it up  :smiley:  I'll try to get a closer look at it as soon as possible, it looks promising!

PR #1929, opened 2025-07-20
Enhancement: Add column width setting

Koc:


Koc:
@enjeck I've added Behat tests to make sure that we able to store value of the new property and fetch it back.

PR rebased, conflicts solved, tests added. So, please, review

lib/Migration/Version001000Date20250720000000.php
```diff
@@ -0,0 +1,40 @@
...
+	 * @return null|ISchemaWrapper
+	 * @throws Exception
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('tables_columns');
+		if (!$table->hasColumn('custom_settings')) {
+			$table->addColumn('custom_settings', Types::JSON, [
```
AIlkiv, line 33: There is no point in messing around with JSON here; plain TEXT will suffice.

src/modules/modals/CreateColumn.vue
```diff
@@ -321,7 +328,10 @@ export default {
 				mandatory: this.column.mandatory,
 				viewId: this.isView ? this.element.id : null,
 				tableId: !this.isView ? this.element.id : null,
+				customSettings: this.column.customSettings,
```
AIlkiv: It will probably look more readable.

Instead of:
`customSettings: this.column.customSettings,`
`data.customSettings = { width: data.customSettings.width }`

Do this here:
`customSettings: { width: this.column.customSettings.width }`

src/shared/components/ncTable/partials/columnTypePartials/forms/MainForm.vue
```diff
@@ -95,6 +109,12 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		customSettings: {
+			type: [Object, Array],
```
AIlkiv: If I understand correctly, the Array type is when it is empty. Can you correct it so that the type is only Object?

appinfo/info.xml
```diff
@@ -26,7 +26,7 @@ Share your tables and views with users and groups within your cloud.
 Have a good time and manage whatever you want.
 
 ]]></description>
-	<version>1.0.0-beta.1</version>
+	<version>1.0.0-beta.2</version>
```
AIlkiv: This is unnecessary.

AIlkiv (commented):
Thanks for the PR. I left a few comments.

lib/Migration/Version001000Date20250720000000.php
Koc: I'd prefer to be consistent with already existent columns https://github.com/nextcloud/tables/blob/main/lib/Migration/Version000600Date20230703000000.php#L68-L76

src/shared/components/ncTable/partials/columnTypePartials/forms/MainForm.vue
Koc: It will not work as expected then because server responds with `[]` instead of `{}` if width not defined yet.

I had an attempt to return `new \stdClass()` to always respond `{}` but in this case we can't properly serialize response and there is http 500 error occurs.

src/modules/modals/CreateColumn.vue
```diff
@@ -287,6 +289,11 @@ export default {
 			if (!this.column.title) {
 				showInfo(t('tables', 'Please insert a title for the new column.'))
 				this.titleMissingError = true
+			} else if (this.column.customSettings?.width
+				&& (this.column.customSettings?.width < 20 || this.column.customSettings?.width > 1000)) {
+				this.titleMissingError = false
+				showError(t('tables', 'Cannot save column. Column width must be between {min} and {max}.', { min: 20, max: 1000 }))
+				this.typeMissingError = true
```
silverkszlo: There are already quite a few boolean switches here, but using something like `this.widthInvalidError` would be more consistent and clear

src/shared/components/ncTable/partials/columnTypePartials/forms/MainForm.vue
```diff
@@ -28,6 +28,20 @@
 			<NcCheckboxRadioSwitch type="switch" :checked.sync="localMandatory" />
 		</div>
 
+		<!-- column width -->
+		<div class="fix-col-4 mandatory title space-T">
+			{{ t('tables', 'Column width') }}
+		</div>
+		<div class="fix-col-4">
+			<input
```
silverkszlo, line 36: It might be helpful to provide a small hint under the input field, for example: 'Column width must be between 20 and 1000'. This way, users know the limits from the start, rather than only discovering them through an error message. The pixel format is also only implicit and may not be obvious to everyone.

Screenshot from 2025-09-03 16-17-10

src/shared/components/ncTable/partials/columnTypePartials/forms/MainForm.vue
```diff
@@ -28,6 +28,20 @@
...
+				v-model.number="localColumnWidth"
+				type="number"
+				pattern="\d+"
+				min="20"
```
silverkszlo: I wonder if min-width should be higher than 20 because with 20px (1st col) one can not really see anything in the cell as there seems to be a bit of padding:

Screenshot from 2025-09-03 16-31-46

src/shared/components/ncTable/partials/columnTypePartials/forms/MainForm.vue
Koc: ok, I managed to fix it

src/shared/components/ncTable/partials/TableHeader.vue
```diff
@@ -10,7 +10,11 @@
 				<div v-if="hasRightHiddenNeighbor(-1)" class="hidden-indicator-first" @click="unhide(-1)" />
 			</div>
 		</th>
-		<th v-for="col in visibleColumns" :key="col.id">
+		<th v-for="col in visibleColumns" :key="col.id"
+			:style="{
+				width: col.customSettings?.width ? `${col.customSettings.width}px` : 'auto',
```
enjeck: can move `col.customSettings?.width ? ${col.customSettings.width}px : 'auto',` into a function and reuse

src/shared/components/ncTable/partials/TableRow.vue
```diff
@@ -7,9 +7,16 @@
 		<td v-if="config.canSelectRows" :class="{sticky: config.canSelectRows}">
 			<NcCheckboxRadioSwitch :checked="selected" @update:checked="v => $emit('update-row-selection', { rowId: row.id, value: v })" />
 		</td>
-		<td v-for="col in visibleColumns"
-			:key="col.id"
-			:class="{ 'search-result': getCell(col.id)?.searchStringFound, 'filter-result': getCell(col.id)?.filterFound }"
+		<td v-for="col in visibleColumns" :key="col.id"
+			:style="{
+				width: col.customSettings?.width ? `${col.customSettings.width}px` : 'auto',
```
enjeck: same here

src/shared/components/ncTable/partials/columnTypePartials/forms/MainForm.vue
```diff
@@ -28,6 +28,20 @@
...
+		<!-- column width -->
+		<div class="fix-col-4 mandatory title space-T" :class="{error: widthInvalidError}">
+			{{ t('tables', 'Column width') }}
+		</div>
+		<div class="fix-col-4" :class="{error: widthInvalidError}">
+			<input
+				v-model.number="localColumnWidth"
+				type="number"
+				pattern="\d+"
+				min="20"
```
enjeck: I thought we changed the min to be higher than 20?

src/shared/components/ncTable/partials/columnTypePartials/forms/MainForm.vue
```diff
@@ -116,6 +147,12 @@ export default {
 				this.$emit('update:selectedViews', selectedViews)
 			},
 		},
+		localColumnWidth: {
+			get() { return this.customSettings.width ?? null },
```
enjeck: To be safe:
```suggestion
			get() { return this.customSettings?.width ?? null },
```

src/shared/components/ncTable/partials/TableHeader.vue
enjeck: Also, should we consider how these fixed widths would be like on mobile devices? Might run into a responsiveness issue

enjeck: Maybe something using `vw` would be better:

```
 maxWidth: col.customSettings?.width ? 
  `min(${col.customSettings.width}px, 90vw)` : 'auto'
```

Koc: Right now it looks acceptable. Here an example with a columns that have width respectfully 50 and 200 pixels.

image


to be honest I don't know what to improve here. I'm not a profy in FE development :man_shrugging: . Feel free to suggest code snippets or just commit in current branch.

Koc: extracted into a reusable function for all styles

src/shared/components/ncTable/partials/columnTypePartials/forms/MainForm.vue
Koc: replaced with a constants

src/shared/components/ncTable/partials/TableHeader.vue
enjeck: Looks good to me!

PR #1958, opened 2025-07-27
Enhancement: add a read-only option to view columns

### LINKED ISSUE #1154: View feature (display column but non editable column)

Koc:
Closes #1154 and replaces #1916. This PR contains complete implementation for the both BE and FE sides.

## :construction: Todo:
- [ ] Tests


Aveyron-RetD:
1000 thanks

Would it not be possible that in edition mode those columns do appear but they are greyed out and cannot be entered in. By removing them from edition here you might not know what you are editing.

Aveyron-RetD:
\o/ thanks for all your contributions. Hope they all get merged soon.

lib/Controller/Api1Controller.php
```diff
@@ -728,19 +728,14 @@ public function updateShareDisplayMode(int $shareId, int $displayMode, string $t
...
 		try {
-			try {
-				// permission check in service class
-				// they might be missing legally when only the view was shared
-				$columns = $this->columnService->findAllByTable($tableId);
-			} catch (PermissionError $e) {
-				if (!$viewId) {
-					throw $e;
-				}
+			if ($viewId) {
```
blizzz, line 732: Changes in this file (or in the cypress ones) are OK refactorings, but should not be mixed into a PR that is about something else. Generally, such changes blow the PR and makes it hard to review, and might be confusing in retrospective by being out of the context. In future, please keep them for a separate PR 🙂

lib/Service/ValueObject/ViewColumnInformation.php
```diff
@@ -17,28 +17,40 @@
...
 		return $this->offsetGet(self::KEY_ORDER);
 	}
 
+	public function isReadonly(): bool {
+		return $this->offsetGet(self::KEY_READONLY) ?? false;
+	}
+
 	public static function fromArray(array $data): static {
-		$vci = new static($data[self::KEY_ID], $data[self::KEY_ORDER]);
+		$vci = new static($data[self::KEY_ID], $data[self::KEY_ORDER], $data[self::KEY_READONLY] ?? false);
```
blizzz: If we add all parameters into the constructor, please split this to multiple lines  to gain easier diffs in futures.

lib/Db/View.php
```diff
@@ -122,6 +122,15 @@ public function getColumnsSettingsArray(): array {
 		return $result;
 	}
 
+	public function findColumnSettingsForColumn(int $columnId): ?ViewColumnInformation {
```
blizzz, line 125: I am wondering whether we should index the settings array by id (at least after reading it from Db) 🤔 Probably to micro to be necessary tho.

blizzz (commented):
Looks good from backend side, no blockers, 👍 

@JuliaKirschenheuter @enjeck can we have a look on the frontend bits, too?

lib/Controller/Api1Controller.php
Koc: This exact is not refactoring. It's needed for case when table owner tries to edit his own view. There was a bug that we've loaded just raw view columns without enriching them with `readonly` property. So, table owner can't get correct value for view and overrides them during saving edit form

cypress change is indeed little bit offtopic

lib/Db/View.php
Koc: yeah, I thought about that as well 🤝

lib/Db/Column.php
```diff
@@ -249,6 +252,7 @@ public function jsonSerialize(): array {
 			'lastEditBy' => $this->lastEditBy,
 			'lastEditByDisplayName' => $this->lastEditByDisplayName,
 			'lastEditAt' => $this->lastEditAt,
+			'readonly' => $this->readonly,
```
AIlkiv: Why duplicate readonly in column instead of using ViewColumnInformation directly?

lib/Service/ValueObject/ViewColumnInformation.php
```diff
@@ -57,7 +69,8 @@ public function offsetSet(mixed $offset, mixed $value): void {
 		if (!$this->offsetExists($offset)) {
 			return;
 		}
-		$this->data[(string)$offset] = $value;
+		$this->ensureType($offset, $value);
+		$this->data[$offset] = $value;
```
AIlkiv: It might be better to make ensureType return “value”.

`$this->data[$offset] = $this->ensureType($offset, $value);`

src/modules/modals/EditRow.vue
```diff
@@ -181,7 +181,6 @@ export default {
 		},
 		reset() {
 			this.localRow = {}
-			this.dataLoaded = false
```
AIlkiv: Does this apply to this task?

AIlkiv (commented):
Thank you for the PR. I have a few minor comments.

lib/Controller/Api1Controller.php
Koc: cypress changes have been reverted

lib/Service/ValueObject/ViewColumnInformation.php
Koc: done

src/modules/modals/EditRow.vue
Koc: reverted

lib/Service/ValueObject/ViewColumnInformation.php
Koc: done

lib/Db/View.php
Koc: as this PR already huge, I propose to do any refactoring about columns indexing in a separate PR :pray:

lib/Db/Column.php
Koc: do you mean inject whole `ViewColumnInformation` object into `Column` instead of readonly?

cypress/e2e/view.cy.js
```diff
@@ -113,6 +113,26 @@ describe('Interact with views', () => {
...
+		cy.intercept({ method: 'PUT', url: '**/apps/tables/view/*' }).as('updateView')
+		cy.get('[data-cy="modifyViewBtn"]').contains('Create View').click()
+		cy.wait('@createView')
+		cy.wait('@updateView')
+
+		cy.get('[data-cy="navigationViewItem"]').contains(title).should('exist')
+
+		// Make sure that column is readonly during edit
+		cy.get('[data-cy="customTableRow"]').contains('first row').parent().parent().find('[data-cy="editRowBtn"]').click()
+		cy.get('[data-cy="editRowModal"]').contains('.row.space-T', 'title').find('input').should('have.attr', 'readonly')
```
enjeck: Might be better to check that the input is disabled
```suggestion
		cy.get('[data-cy="editRowModal"]').contains('.row.space-T', 'title').find('input')).should('be.disabled')
```

Koc: But in fact input element is readonly, not disabled. Only few inputs like toggle or select are disabled because they aren’t supporting readonly

enjeck: Okay!

lib/Db/Column.php
Koc: done, now we are injecting whole instance of the `ViewColumnInformation` into `Column` object

PR #1971, opened 2025-08-03
test(Cypress) Fix Cypress component tests

Koc:
[Pipeline fails](https://github.com/nextcloud/tables/actions/runs/16707741127/job/47288220333?pr=1971) even in `man` branch even if we change readme file.


enjeck (commented):
Looks like it's enough to just remove the `nbsp` like I did at https://github.com/nextcloud/tables/pull/1972

PR #1973, opened 2025-08-05
feat: Inline editing - Stars

enjeck:
Part of #364 
Supports inline editing for TableCellStars

### DIFF
--- a/src/shared/components/ncTable/partials/TableCellStars.vue
+++ b/src/shared/components/ncTable/partials/TableCellStars.vue
@@ -63,12 +63,6 @@ export default {
 			const v = this.value
 			return starFull.repeat(v) + starEmpty.repeat(5 - v)
 		},
-		getEditValue() {
-			const starEmpty = '☆'
-			const starFull = '★'
-			const v = this.editValue
-			return starFull.repeat(v) + starEmpty.repeat(5 - v)
-		},
 	},
 
 	watch: {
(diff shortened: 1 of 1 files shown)


marcoambrosini (changes requested):
Hi @enjeck, I think it would be better to allow clicking on the stars directly instead of having a plus and minus buttons to the sides. Could you change that?

src/shared/components/ncTable/partials/TableCellStars.vue
```diff
@@ -26,21 +55,138 @@ export default {
...
 	},
+
 	computed: {
 		getValue() {
 			const starEmpty = '☆'
 			const starFull = '★'
 			const v = this.value
 			return starFull.repeat(v) + starEmpty.repeat(5 - v)
 		},
+		getEditValue() {
```
silverkszlo: this is not being used anymore, right?

src/shared/components/ncTable/partials/TableCellStars.vue
```diff
@@ -26,21 +55,138 @@ export default {
...
+.align-center {
+	align-items: center;
+	display: flex;
+
+	&.is-loading {
+		opacity: 0.7;
+	}
+}
+
+.stars {
```
silverkszlo: this is not being used

src/shared/components/ncTable/partials/TableCellStars.vue
```diff
@@ -26,21 +55,138 @@ export default {
...
+			transform: scale(1.1);
+		}
+	}
+
+	&.filled {
+		color: var(--color-warning);
+	}
 }
 
+.editor-buttons {
```
silverkszlo: this is not being used

PR #1974, opened 2025-08-05
feat: Inline editing - Datetime

enjeck:
Part of #364 

Adds support for inline editing for Date, Time and Datetime column types

### DIFF
--- a/src/shared/components/ncTable/partials/TableCellDateTime.vue
+++ b/src/shared/components/ncTable/partials/TableCellDateTime.vue
@@ -4,13 +4,17 @@
 -->
 <template>
 	<div class="cell-datetime">
-		<div v-if="!isEditing" class="non-edit-mode" @click="handleStartEditing">
+		<div v-if="!isEditing" class="non-edit-mode" role="button" aria-label="Edit date/time"
+			tabindex="0"
+			@click="handleStartEditing">
 			{{ getValue }}
 		</div>
 		<div v-else
 			ref="editingContainer"
 			class="inline-editing-container"
 			tabindex="0"
+			role="group"
+			aria-label="Edit date/time"
 			@keydown.enter="saveChanges"
 			@keydown.escape="cancelEdit">
 			<div class="datetime-picker-container" :class="{ 'is-loading': localLoading }">


marcoambrosini (changes requested):
We should remove the "select a date" heading here. The fully accessible view with labels is always present in the row-editing-dialog. So it's ok if we do not have labels in inline editing.

src/shared/components/ncTable/partials/TableCellDateTime.vue
```diff
@@ -3,15 +3,46 @@
...
-	<div>
-		{{ getValue }}
+	<div class="cell-datetime">
+		<div v-if="!isEditing" class="non-edit-mode" @click="startEditing">
+			{{ getValue }}
+		</div>
+		<div v-else
+			ref="editingContainer"
+			class="inline-editing-container"
+			tabindex="0"
```
silverkszlo, line 15: concerning accessibility: when tabbing my way through the page on the keyboard I can't reach those fields and do the inline editing. Probably because `tabindex=0` only works, when `isEditing` is already `true` which can only turn `true` with a click?

src/shared/components/ncTable/partials/TableCellDateTime.vue
```diff
@@ -34,17 +72,197 @@ export default {
 			}
 			return this.column.formatValue(this.value)
 		},
+
 		getDefaultValue() {
```
silverkszlo: This function is not being used anymore or not yet?

src/shared/components/ncTable/partials/TableCellDateTime.vue
enjeck: Yes

src/shared/components/ncTable/partials/TableCellDateTime.vue
```diff
@@ -26,25 +56,199 @@ export default {
...
+		}
+	}
+
+	.datetime-picker-container input {
+		width: 100%;
+	}
+
+}
+
+.editor-buttons {
```
silverkszlo: This class is not being used anymore or yet?

src/shared/components/ncTable/partials/TableCellDateTime.vue
silverkszlo: also an aria-label and role would be nice (above in the non-edit-mode div)

src/shared/components/ncTable/partials/TableCellDateTime.vue
```diff
@@ -3,15 +3,46 @@
...
+			tabindex="0"
+			@keydown.enter="saveChanges"
+			@keydown.escape="cancelEdit">
+			<div class="datetime-picker-container" :class="{ 'is-loading': localLoading }">
+				<NcDateTimePickerNative
+					v-model="editDateTimeValue"
+					:type="getPickerType"
+					:label="getPlaceholder"
+					:disabled="localLoading || !canEditCell()" />
+				<div v-if="canBeCleared" class="icon-close make-empty" @click="emptyValue" />
```
silverkszlo: maybe a `role="button"` and `aria-label="Clear date"`

src/shared/components/ncTable/partials/TableCellDateTime.vue
```diff
@@ -34,17 +72,197 @@ export default {
...
+				})
+			} else {
+				// Remove click outside listener
+				document.removeEventListener('click', this.handleClickOutside)
+			}
+		},
+	},
+
+	methods: {
+		t,
```
silverkszlo, line 111: not being used anymore

silverkszlo (commented):
Looks very cool! I have tested it in Firefox and Chromium and all in all it works nicely. I have only suggested minor changes for more accessibility and removing potentially obsolete code.

src/shared/components/ncTable/partials/TableCellDateTime.vue
enjeck: Not sure what values to use here though :thinking:

silverkszlo: The `non-edit-mode` div could additionally receive:
```
role="button"
aria-label="Edit date/time"
tabindex="0"
```

And the `inline-editing-container` div:
```
role="group"
aria-label="Edit date/time"
```

PR #1975, opened 2025-08-05
feat: Inline editing - Selection

enjeck:
Part of #364 

Supports inline editing for the Selection and MultiSelection column types

TODO
- [ ] find alternative to setTimeout. Maybe as described at https://stackoverflow.com/questions/18504139/div-onblur-function

### DIFF
--- a/src/shared/components/ncTable/partials/TableCellMultiSelection.vue
+++ b/src/shared/components/ncTable/partials/TableCellMultiSelection.vue
@@ -4,7 +4,7 @@
 -->
 <template>
 	<div class="cell-multi-selection">
-		<div v-if="!isEditing" class="non-edit-mode" @click="startEditing">
+		<div v-if="!isEditing" class="non-edit-mode" @click="handleStartEditing">
 			<ul>
 				<li v-for="v in getObjects()" :key="v.id">
 					{{ v.label }}<span v-if="v.deleted" :title="t('tables', 'This option is outdated.')">&nbsp;⚠️</span>
@@ -66,6 +66,7 @@ export default {
 	data() {
 		return {
 			localEditValues: [],
+			isInitialEditClick: false,
 		}
 	},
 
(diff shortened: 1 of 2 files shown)


nimishavijay (changes requested):
Looks good except for the broken border in the dropdown, that would be something we have to fix.
A nice-to-have: in view mode show the multiple options styled as user bubbles, similar to how they are shown in the dropdown. Currently they are shown as bullet list items which looks very different. Can also be in a follow up

src/shared/components/ncTable/partials/TableCellMultiSelection.vue
```diff
@@ -35,20 +62,157 @@ export default {
...
+	computed: {
+		getOptions() {
+			return this.column.selectionOptions || []
+		},
+		getAllNonDeletedOrSelectedOptions() {
+			const options = this.getOptions.filter(item => {
+				return !item.deleted || this.optionIdIsSelected(item.id)
+			}) || []
+
+			options.forEach(opt => {
```
silverkszlo, line 82: `forEach` would change the `column.selectionOptions` objects directly (since filter doesn’t create copies). Was that the intention? If the goal is only to adjust the labels for display, maybe `map` could be a safer choice to avoid mutating the original data?

src/shared/components/ncTable/partials/TableCellSelection.vue
```diff
@@ -31,19 +55,121 @@ export default {
...
+			return this.column.isDeletedLabel(this.value)
+		},
+
+		getOptionObject(id) {
+			return this.getOptions.find(e => e.id === id) || null
+		},
+
+		initEditValue() {
+			if (this.value !== null) {
+				this.editValue = this.getOptionObject(parseInt(this.value))
```
silverkszlo, line 113: is `parseInt` here necessary if `value` is already a Number (in props)?

src/shared/components/ncTable/partials/TableCellMultiSelection.vue
enjeck: Yeah, that's a good point. I copied this function from the existing code at https://github.com/nextcloud/tables/blob/038a943b443e8ba6e99074b8ed721d4f03ee2b09/src/shared/components/ncTable/partials/rowTypePartials/SelectionMultiForm.vue#L48. Unrelated to inline editing so should be updated separately everywhere.

silverkszlo: #2015

PR #1977, opened 2025-08-05
feat: Inline editing - link

enjeck:
Part of #364 
Supports inline editing for the Link column type

### DIFF
--- a/src/shared/components/ncTable/partials/TableCellLink.vue
+++ b/src/shared/components/ncTable/partials/TableCellLink.vue
@@ -28,6 +28,7 @@
 				<NcTextField v-if="isPlainUrl"
 					v-model="plainLink"
 					:placeholder="t('tables', 'URL')"
+					:aria-label="t('tables', 'URL')"
 					:disabled="localLoading || !canEditCell()" />
 				<NcSelect v-else
 					v-model="editValue"


src/shared/components/ncTable/partials/TableCellLink.vue
```diff
@@ -62,15 +115,231 @@ export default {
...
+				item.subline = res.data?.ocs?.data?.name
+				item.value = item.resourceUrl
+			}
+			this.results = this.results.concat(res.data?.ocs?.data?.entries)
+			this.setProviderLoading(providerId, false)
+		},
+
+		isValidUrl(string) {
+			try {
+				return new URL(string)
```
silverkszlo, line 306: maybe it would be bug-safer if this would return `true` instead of the url object?

src/shared/components/ncTable/partials/TableCellLink.vue
```diff
@@ -62,15 +115,231 @@ export default {
...
 	},
 
+	watch: {
+		term() {
+			this.debounceSubmit()
+		},
+		isEditing(isEditing) {
+			if (isEditing) {
+				this.initEditMode()
+				this.$nextTick(() => {
```
silverkszlo, line 189: probably it would make sense to extract this to a separate helper function as it is being used in multiple 'inline editing'-PRs?

nimishavijay (changes requested):
Nice! This seems to be for a link to a file (unless I misunderstood), would also love to see how it looks for regular links. Either way, the border issue in the dropdown still exists here. 
Additionally, I noticed in view mode the link ellipsizes very early even though there is lots of space available. We can show more of it :)

src/shared/components/ncTable/partials/TableCellLink.vue
```diff
@@ -3,31 +3,75 @@
...
+				<NcTextField v-if="isPlainUrl"
+					v-model="plainLink"
+					:placeholder="t('tables', 'URL')"
+					:disabled="localLoading || !canEditCell()" />
+				<NcSelect v-else
+					v-model="editValue"
+					:options="results"
+					:clearable="true"
+					label="title"
+					:aria-label-combobox="t('tables', 'Link providers')"
```
silverkszlo, line 38: As there can be multiple columns with link provider dropdowns, it would be better to make the label more specific — for example:
```
:aria-label-combobox="t(
    'tables',
    'Choose link provider for column {column}',
    { column: column.title || t('tables', 'Unnamed column') }
)"
```

src/shared/components/ncTable/partials/TableCellLink.vue
```diff
@@ -3,31 +3,75 @@
...
+				:underline-title="true" />
+		</div>
+		<div v-else
+			ref="editingContainer"
+			class="edit-mode"
+			tabindex="0"
+			@keydown.enter="saveChanges"
+			@keydown.escape="cancelEdit">
+			<div class="link-input">
+				<NcTextField v-if="isPlainUrl"
```
silverkszlo, line 28: could use an aria-label

src/shared/components/ncTable/partials/TableCellLink.vue
```diff
@@ -3,31 +3,75 @@
...
+					<template #option="props">
+						<LinkWidget :thumbnail-url="props.thumbnailUrl" :icon-url="props.icon" :title="props.title" :subline="props.subline" :icon-size="40" />
+					</template>
+					<template #selected-option="props">
+						<LinkWidget :thumbnail-url="props.thumbnailUrl" :icon-url="props.icon" :title="props.title" :icon-size="24" />
+					</template>
+				</NcSelect>
+			</div>
+			<div v-if="localLoading" class="loading-indicator">
+				<div class="icon-loading-small icon-loading-inline" />
```
silverkszlo, line 52: `role="status"` and `aria-live="polite"` would be great

src/shared/components/ncTable/partials/TableCellLink.vue
```diff
@@ -62,15 +115,231 @@ export default {
...
+				}
+				this.setProviderLoading(providerId, false)
+				return
+			}
+
+			let res = null
+			try {
+				res = await axios.get(generateOcsUrl('/search/providers/' + providerId + '/search?term=' + term))
+			} catch (e) {
+				displayError(e, t('tables', 'Could not load link provider results.'))
```
silverkszlo, line 290: More of a todo for future enhancement: maybe consider combining errors from multiple providers—right now each failure (like network issues or provider unavailability) pops up its own message, which could get a bit much for users. A single, combined error might feel smoother

src/shared/components/ncTable/partials/TableCellLink.vue
enjeck: Yeah, I agree. Will consider it after all the various PRs have been independently merged

enjeck: Created a follow up at https://github.com/nextcloud/tables/issues/2009

src/shared/components/ncTable/partials/TableCellLink.vue
enjeck: If adding here, would need to add to all other loading icons and can be done separarely as a follow-up

src/shared/components/ncTable/partials/TableCellLink.vue
enjeck: Yeah, makes sense. And right now that I have added the link validation, trying to save several times by hitting enter multiple times with an invalid link leads to multiple popups with the same message. Could be improved for sure

src/shared/components/ncTable/partials/TableCellLink.vue
enjeck: I can't see how it'll be bug safer? Sure, we can call `new URL` in a separate line and return `true` afterwards but it looks like it's same effect and safeness to me. Unless I'm missing something

src/shared/components/ncTable/partials/TableCellLink.vue
enjeck: I'm not sure exactly what's the value of having a super-specific label like this? Because if we can justify the value of doing it here for the user, we should also consider doing it all over the app since there are many unspecific labels

silverkszlo: I think the main value here is accessibility: with a generic label like “Choose link provider”, a screen reader user wouldn’t know which column they’re editing if there are multiple link columns. Including the column title makes each field uniquely identifiable and usable. I don’t have the full picture of how accessibility is usually handled in the app — it could be worth considering more broadly later — but in this case it feels especially relevant, since otherwise the table isn’t really accessible for screen reader users.

src/shared/components/ncTable/partials/TableCellLink.vue
silverkszlo: Yeah, you're right — technically it doesn’t make a difference since a URL object is always truthy. The name `isValidUrl` just suggested returning a boolean instead of an object, but that’s really just semantics and probably too nitpicky. Never mind then!

silverkszlo: Ah, another thing though: currently it allows me to safe a value like "https:wikipedia" or "http:hui"

enjeck: > Ah, another thing though: currently it allows me to safe a value like "https:wikipedia" or "http:hui"

Oooh, good catch. Right now, our logic for checking invalid URLs only checks the protocol, not the tld. This is unrelated to inline editing and should be fixed separately everywhere. Opened an issue at https://github.com/nextcloud/tables/issues/2010

src/shared/components/ncTable/partials/TableCellLink.vue
silverkszlo: Which kind of error triggers the popup though? When I try to save "www.wikipedia.org" no error popup turns up and the value is not saved

src/shared/components/ncTable/partials/TableCellLink.vue
enjeck: > I think the main value here is accessibility: with a generic label like “Choose link provider”, a screen reader user wouldn’t know which column they’re editing if there are multiple link columns.

I see. So this applies to all columns then, not just specific to the link column. Since there could be multiple columns of any type. If we want to be this detailed, then we should do it everywhere too. Would be worth discussing with @JuliaKirschenheuter since she's an accessibility expert. Potential follow up task at https://github.com/nextcloud/tables/issues/2011

src/shared/components/ncTable/partials/TableCellLink.vue
enjeck: If you type a URL without a protocol, then it doesn't even show in the dropdown so you're not allowed to select it. It's only after you've selected and try to save does the error pop up. Can try with an unsupported protocol like `tel://` or `mailto://`

src/shared/components/ncTable/partials/TableCellLink.vue
silverkszlo: added an issue here: #2012

src/shared/components/ncTable/partials/TableCellLink.vue
silverkszlo: made an issue here: #2013

src/shared/components/ncTable/partials/TableCellLink.vue
silverkszlo: Oh and sorry I've mixed something up here: it's not `:aria-label-combobox` as `combobox` is a role.` :aria-label` is correct here.

enjeck: `:aria-label-combobox` is fine too, since other apps are using it for `NcSelect`

PR #1978, opened 2025-08-05
feat: Inline editing - usergroup

enjeck:
Part of #364 

Supports inline editing for the usergroup column

TODO
- [x] find alternative to setTimeout

### DIFF
--- a/src/shared/components/ncTable/partials/TableCellUsergroup.vue
+++ b/src/shared/components/ncTable/partials/TableCellUsergroup.vue
@@ -4,7 +4,7 @@
 -->
 <template>
 	<div class="cell-usergroup">
-		<div v-if="!isEditing" class="non-edit-mode" @click="startEditing">
+		<div v-if="!isEditing" class="non-edit-mode" @click="handleStartEditing">
 			<div v-if="value" class="table-cell-usergroup">
 				<div v-for="item in value" :key="item.id" class="inline usergroup-entry">
 					<NcUserBubble :user="item.id" :avatar-image="getAvatarImage(item)" :is-no-user="!isUser(item)" :display-name="item.displayName ?? item.id" :show-user-status="isUser(item) && column.showUserStatus" :size="column.showUserStatus ? 34 : 20" :primary="isCurrentUser(item)" />
@@ -88,6 +88,7 @@ export default {
 			selectUsers: this.column?.usergroupSelectUsers ?? true,
 			selectGroups: this.column?.usergroupSelectGroups ?? false,
 			selectCircles: false,
+			isInitialEditClick: false,
 		}
 	},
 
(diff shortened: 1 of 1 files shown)


marcoambrosini (changes requested):
Here we have a problem with the styling of the multiselect. The multiselect should have the black outline and white background. This is what we currently have there:

Screenshot 2025-08-06 at 16 23 24

src/shared/components/ncTable/partials/TableCellUsergroup.vue
```diff
@@ -3,26 +3,70 @@
...
+			@keydown.escape="cancelEdit">
+			<NcSelect v-model="editValue"
+				style="width: 100%;"
+				:loading="loading || localLoading"
+				:options="options"
+				:placeholder="getPlaceholder()"
+				:searchable="true"
+				:get-option-key="(option) => option.key"
+				label="displayName"
+				:aria-label-combobox="getPlaceholder()"
```
silverkszlo, line 28: sorry, this is probably my fault: this should be just `:aria-label` instead of `:aria-label-combobox`

src/shared/components/ncTable/partials/TableCellUsergroup.vue
```diff
@@ -55,18 +141,83 @@ export default {
...
+				return
+			}
+
+			let newValue = !Array.isArray(this.editValue) ? [this.editValue] : this.editValue
+			// If the column does not allow multiple items, limit to one item
+			// in case we switched from multiple to single selection
+			if (!this.column.usergroupMultipleItems) {
+				newValue = newValue.slice(0, 1)
+			}
+			const success = await this.updateCellValue(newValue)
```
silverkszlo, line 169: A try/catch would probably make sense here. Something like:
```
try {
  const success = await this.updateCellValue(newValue)
  if (!success) {
      this.cancelEdit()
  }
} catch (error) {
    console.error('Failed to save changes:', error)
    this.cancelEdit()
} finally {
    this.localLoading = false
    this.isEditing = false
 }
```

enjeck: I don't think there's ever an error to catch, since we already catch errors at https://github.com/nextcloud/tables/blob/15b4efaf26f1b32fbeee61c8dc700e91d08a3db1/src/store/data.js#L184 and return a boolean

src/shared/components/ncTable/partials/TableCellUsergroup.vue
enjeck: Not your fault, that's all me, haha. I notice that `aria-label-combobox` is used for `NcSelect` everywhere and in other apps, so let's keep it.

PR #1979, opened 2025-08-06
feat: remove rows in view after edit when they no longer match filter

### LINKED ISSUE #1691: update filtered view on entry change

I have a view which filters depending on if a checkbox is enabled or not. If I uncheck an entry, I have to refresh the page manually to get it to hide from the view (and also to update the row count in the sidebar). It would be nice if this is done automatically.

Image

Update the view on an entry change (that the filtering depends on) so that the filter stays in-sync with the data.

Manually refreshing the page

### Additional context

_No response_

benjaminfrueh:
- backend: add new endpoint
- controller & service:  for server-side check
- frontend: call check after insert and update, remove invalid rows
- wrap rows in <transition-group> to animate removal
- only animate when a row is removed due to filter mismatch
- CSS: fade+collapse leave transition

Closes #1691

### DIFF
--- a/src/store/data.js
+++ b/src/store/data.js
@@ -203,7 +203,7 @@ export const useDataStore = defineStore('data', {
 				const row = res.data
 				const index = this.rows[stateId].findIndex(r => r.id === row.id)
 				set(this.rows[stateId], index, row)
-				await this.removeRowIfNotInView({ rowId: row?.id, viewId: viewId, stateId: stateId })
+				await this.removeRowIfNotInView({ rowId: row?.id, viewId, stateId })
 			}
 
 			return true
@@ -226,7 +226,7 @@ export const useDataStore = defineStore('data', {
 				const row = res?.data?.ocs?.data
 				const newIndex = this.rows[stateId].length
 				set(this.rows[stateId], newIndex, row)
-				await this.removeRowIfNotInView({ rowId: row?.id, viewId: viewId, stateId: stateId })
+				await this.removeRowIfNotInView({ rowId: row?.id, viewId, stateId })
 			}
 
 			return true
--- a/lib/Controller/RowController.php
+++ b/lib/Controller/RowController.php
@@ -103,7 +103,7 @@ class RowController extends Controller {
 
 	#[NoAdminRequired]
 	public function presentInView(int $id, int $viewId): DataResponse {
-		return $this->handleError(function() use ($id, $viewId) {
+		return $this->handleError(function () use ($id, $viewId) {
 			$present = $this->service->isRowInViewPresent($id, $viewId, $this->userId);
 			return ['present' => $present];
 		});


appinfo/routes.php
```diff
@@ -88,6 +88,7 @@
 		['name' => 'row#updateSet', 'url' => '/row/{id}', 'verb' => 'PUT'],
 		['name' => 'row#destroyByView', 'url' => '/view/{viewId}/row/{id}', 'verb' => 'DELETE'],
 		['name' => 'row#destroy', 'url' => '/row/{id}', 'verb' => 'DELETE'],
+		['name' => 'row#presentInView', 'url'  => '/row/{id}/view/{viewId}/present', 'verb' => 'GET'],
```
enjeck: Maybe `/row/{id}/view/{viewId}/present` could be swapped to `/view/{viewId}/row/{id}/present` to match the existing `destroyByView` URL order for consistency

lib/Service/RowService.php
```diff
@@ -614,6 +614,22 @@ public function getViewRowsCount(View $view, string $userId): int {
 		}
 	}
 
+	public function isRowInViewPresent(int $rowId, int $viewId, string $userId): bool {
```
enjeck: I believe this throws `PermissionError` when not found due to permissions, but returns `false` if not present? This is worth documenting in PHPDoc so behavior is clear

src/store/data.js
```diff
@@ -252,5 +271,17 @@ export const useDataStore = defineStore('data', {
...
+
+		async checkRowInView({ rowId, viewId }) {
+			try {
+				const res = await axios.get(generateUrl('/apps/tables/row/{rowId}/view/{viewId}/present', { rowId, viewId }))
+				return res.data.present
+			} catch (e) {
+				showError(t('tables', 'Could not verify row. View is reloaded'))
+				await this.loadRowsFromBE({ viewId })
+			}
+			return true
```
enjeck: So we will return `true` even on error with this?

src/store/data.js
```diff
@@ -202,7 +203,16 @@ export const useDataStore = defineStore('data', {
 				const row = res.data
 				const index = this.rows[stateId].findIndex(r => r.id === row.id)
 				set(this.rows[stateId], index, row)
+
+				if (isView && row?.id != null) {
+					const rowInView = await this.checkRowInView({ rowId: row.id, viewId: viewId })
+					if (!rowInView) {
+						emit('tables:row:animate')
+						this.rows[stateId] = this.rows[stateId].filter(r => r.id !== row.id)
```
enjeck: This block is abso duplicated below, so we could refactor into one function and reuse?

appinfo/routes.php
benjaminfrueh: Changed the order for consistency in the latest commit.

lib/Service/RowService.php
benjaminfrueh: Added phpdoc to make return/throw behaviour clear in the latest commit. Also removed the unnecessary rethrow of InternalError.

src/store/data.js
benjaminfrueh: removed the unnecessary return on error because in the error case it can't verify whether it is in view or not, and added explicit checks for === false before removing rows

src/store/data.js
benjaminfrueh: Moved the duplicate logic with their checks into a reusable function removeRowIfNotInView().

benjaminfrueh:
@enjeck Thank you very much for the review and feedback. I committed the changes accordingly.

PR #1985, opened 2025-08-07
1296 infinite loading when visiting non existent application

silverkszlo:
This PR improves error handling when navigating to non-existent tables, views, or contexts in the Tables app.

Previously, accessing a non-existent resource (e.g. /view/999 or /context/999) would result in an unhandled exception (TypeError: undefined) and show only a loading spinner. This was confusing to users and cluttered the console with errors.

### Changes

- Added proper error catching in Table.vue, View.vue, and Context.vue
- Display a user-friendly error message and icon when a resource is not found (404)
- Ensured layout is visually centered and styled consistently with other apps (e.g. Deck)

404_table

### DIFF
--- a/src/modules/main/partials/ErrorMessage.vue
+++ b/src/modules/main/partials/ErrorMessage.vue
@@ -1,5 +1,5 @@
 <!--
-  - SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 
--- a/src/modules/main/partials/ErrorMessage.vue
+++ b/src/modules/main/partials/ErrorMessage.vue
@@ -1,0 +1,48 @@
+<!--
+  - SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<div class="error-container">
+		<IconTables :size="64" style="margin-bottom: 1rem;" />
+		<p>{{ message }}</p>
+	</div>
+</template>
+
+<script>
+import IconTables from '../../../shared/assets/icons/IconTables.vue'
+
+export default {
+	name: 'ErrorMessage',
+	components: { IconTables },
+	props: {
+		message: {
+			type: String,
+			required: true,
+		},
+	},
+}
+</script>
+
+<style lang="scss">
+.error-container {
+	display: flex;
+	flex-direction: column;
+	align-items: center;
+	justify-content: center;
+	text-align: center;
+	padding: 2rem;
+	height: 100dvh;
+	min-height: 100%;
+	color: var(--color-text);
+	opacity: 0.6;
+
+	p {
+		font-size: clamp(1.2rem, 4vw, 2rem);
+		font-weight: 600;
+		max-width: 90%;
+		word-wrap: break-word;
+	}
+}
+</style>
--- a/src/pages/Context.vue
+++ b/src/pages/Context.vue
@@ -121,14 +121,12 @@ export default {
 	},
 
 	watch: {
-		// Watch for changes to active context to make page reactive
 		async activeContext() {
-			if (this.activeContextId && !this.activeContext) {
-				// context does not exists, go to startpage
-				this.$router.push('/').catch(err => err)
-			} else {
-				await this.reload()
+			if (this.errorMessage) {
+				// Already showing an error, don't redirect
+				return
 			}
+			await this.reload()
 		},
 		'context.iconName': {
 			async handler(value) {
--- a/src/pages/Context.vue
+++ b/src/pages/Context.vue
@@ -5,7 +5,13 @@
 <template>
 	<div class="row">
 		<div v-if="loading" class="icon-loading" />
-		<div v-if="!loading && context">
+
+		<div v-else-if="errorMessage" class="error-container">
+			<IconTables :size="64" style="margin-bottom: 1rem;" />
+			<p>{{ errorMessage }}</p>
+		</div>
+
+		<div v-else>
 			<div class="content context">
 				<div class="row first-row">
 					<h1 class="context__title" data-cy="context-title">
@@ -50,11 +56,13 @@ import exportTableMixin from '../shared/components/ncTable/mixins/exportTableMix
 import svgHelper from '../shared/components/ncIconPicker/mixins/svgHelper.js'
 import { useTablesStore } from '../store/store.js'
 import { useDataStore } from '../store/data.js'
+import IconTables from '../shared/assets/icons/IconTables.vue'
 
 export default {
 	components: {
 		MainModals,
 		NcIconSvgWrapper,
+		IconTables,
 		TableWrapper,
 		CustomView,
 	},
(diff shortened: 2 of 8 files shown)


src/pages/Context.vue
```diff
@@ -235,4 +261,24 @@ export default {
...
+	color: var(--color-text);
+	opacity: 0.6;
+
+	p {
+		font-size: clamp(1.2rem, 4vw, 2rem);
+		font-weight: 600;
+		max-width: 90%;
+		word-wrap: break-word;
+	}
+}
```
enjeck: I think this is duplicated twice below too? I wonder if we can move out of this file into a more "global" scope 🤔

enjeck: Or maybe together with the `<IconTables>` + `{{ errorMessage }}` markup, extract this into a small <ErrorMessage> component to keep things consistent

src/pages/Context.vue
```diff
@@ -5,7 +5,13 @@
...
 	<div class="row">
 		<div v-if="loading" class="icon-loading" />
-		<div v-if="!loading && context">
+
+		<div v-else-if="errorMessage" class="error-container">
+			<IconTables :size="64" style="margin-bottom: 1rem;" />
+			<p>{{ errorMessage }}</p>
+		</div>
+
+		<div v-else>
```
enjeck: nit: Maybe we can make the if-else checks consistent with the format in Table and View too. Here, we use v-else-if="errorMessage" after v-if="loading", whereas in View.vue and Table.vue you use v-if="errorMessage" at the top level.

src/pages/Context.vue
```diff
@@ -146,49 +155,66 @@ export default {
...
-							table.isView = false
-							this.contextResources.push(table)
-						}
+			try {
+				await this.loadContext({ id: this.activeContextId })
+				const index = this.contexts.findIndex(c => parseInt(c.id) === parseInt(this.activeContextId))
+				this.context = this.contexts[index]
+
+				if (!this.context) {
+					this.errorMessage = t('tables', 'This application could not be found')
```
enjeck, line 162: This error message is duplicated on line 212 too. Maybe have a function that we can reuse. Same for Table and View

src/pages/Context.vue
```diff
@@ -146,49 +155,66 @@ export default {
...
 						}
 					}
 				}
+			} catch (e) {
+				if (e.message === 'NOT_FOUND') {
+					this.errorMessage = t('tables', 'This application could not be found')
+				} else {
+					this.errorMessage = t('tables', 'An error occurred while loading the application')
+					console.error(e)
+				}
```
enjeck: I'm trying to understand how this works. If a context loads partially (e.g., context exists but some child tables/views 404), would there be an error message?

silverkszlo: No, you're right. I think there would be no error message as the context loads. I adjusted reload() in Context.vue in order to check for existing children. Would be great if you could take a look at that. Thanks!

src/pages/Context.vue
silverkszlo: done

src/pages/Context.vue
silverkszlo: done

src/pages/Context.vue
silverkszlo: you're right, done :)

silverkszlo:
> Thanks! I tested and it mostly works. Only problem is the browser console errors after visiting non-existent resources. For applicatiions:
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'name')
>     at VueComponent.routing (main.js:22:2)
>     at VueComponent.$route (main.js:22:2)
>     at invokeWithErrorHandling (vue.runtime.esm.js:3033:30)
>     at Watcher2.run (vue.runtime.esm.js:3550:21)
>     at flushSchedulerQueue (vue.runtime.esm.js:4141:17)
>     at Array.<anonymous> (vue.runtime.esm.js:3159:20)
>     at flushCallbacks (vue.runtime.esm.js:3081:17)
> logError @ vue.runtime.esm.js:3065
> ```
> 
> For views:
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'title')
>     at VueComponent.routing (main.js:22:2)
>     at VueComponent.created (main.js:22:2)
> ```
> 
> This happens because we're trying to set the page title as seen in https://github.com/nextcloud/tables/blob/main/src/App.vue#L99 and https://github.com/nextcloud/tables/blob/main/src/App.vue#L108 but these do not exist

Thanks for pointing out the code lines! I have added default titles for those missing ressources. Do you think that makes sense?

src/modules/main/partials/ErrorMessage.vue
```diff
@@ -0,0 +1,48 @@
+<!--
+  - SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
```
enjeck:
```suggestion
  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

silverkszlo:
> Was testing again and I noticed that when we're on the "error" page, the `+` buttons against "Tables" and "Applications" that let us create a new Table or Application do not work. Issue might stem from https://github.com/nextcloud/tables/blob/main/src/modules/navigation/sections/Navigation.vue#L202, where we `emit` the events 🤔

Okay I probably have fixed that by always rendering MainModals in Table and View for the Navigation to work correctly. What do you think?

src/pages/Table.vue
```diff
@@ -4,42 +4,94 @@
...
+		async checkTable() {
+			const id = this.activeTableId || this.$route.params.tableId
+			if (!id) return
+
+			try {
+				if (this.activeTableId !== id) {
+					this.setActiveTableId(parseInt(id))
+				}
+
+				await this.loadContextTable({ id })
```
enjeck: I wonder if it's better to ensure that table exists with `this.loadContextTable({ id })` before setting as active table `this.setActiveTableId(parseInt(id))`

src/pages/Context.vue
```diff
@@ -146,49 +151,71 @@ export default {
...
+										tableId: view.tableId,
+									})
+									view.key = 'view-' + (view.id).toString()
+									view.isView = true
+									this.contextResources.push(view)
+								}
+							}
+						} catch (err) {
+							console.error(`Failed to load resource ${node.node_id}:`, err)
+							this.errorMessage = t('tables', 'Some resources in this application could not be loaded')
```
enjeck, line 205: Since there could be multiple errors, let's look at accumulating them, when we're doing https://github.com/nextcloud/tables/issues/2013

src/shared/utils/displayError.js
```diff
@@ -4,6 +4,14 @@
  */
 import { showError } from '@nextcloud/dialogs'
 
+export function getNotFoundError(type) {
+	return t('tables', `This ${type} could not be found`)
```
enjeck: I believe this won't be properly translatable. Please check https://docs.nextcloud.com/server/latest/developer_manual/basics/translations.html#javascript-typescript-vue for the proper format

src/pages/Table.vue
silverkszlo: you're right, I've changed it!

src/pages/Context.vue
silverkszlo: yes, I've added a comment in #2013

src/shared/utils/displayError.js
silverkszlo: oh yes, thank you!

Issue #1897, opened 2025-07-11
Activity support


blizzz:
### Is your feature request related to a problem? Please describe.

As a both an extensions but also minimal alternative to #984  we would like to add Activity Support to Tables. The goal is to be able to see who added, deleted or modified a row or values inside there, and when. 

### Describe the solution you'd like

1. Adding, modifying and deleting a row shall lead to an activity entry with the *activity* app (if enabled).
2. A *Tables* category should be added to the *Activity* app

<details>
<summary>Quickest mockup</summary>
<img width="402" height="792" alt="Image" src="https://github.com/user-attachments/assets/8d88af39-3afa-43a0-8159-87b993ffc600" />
</details>

3. We have a sidebar in tables with *Sharing* and *Integration* tabs. Similar as available with files, we would like to have an *Activity* tab there as well.
4. Ideally the Activity tab is specific to the recently activated row. This can be by clicking/tapping. It should not conflict with inline-editing. A specific *Activity* button next to the *Edit* button might be another approach.
5. General considerations: for activites are being cleaned up after a threshold time, maybe it can be indicated by a label to which timeframe the listed activity belong. Could be a thing for files? Best to be discussed with Design. 
6. The automatic clean-up does not make it a real alternative for auditing. Double check whether activities provides an API to specify a longer lifetime, but I do not expect so and do not really think that this matches with the vision for activities.
7. We have `Row{Added,Deleted,Updated)Events` that can be utilized for this purpose. 

### Describe alternatives you've considered

#984 

### Additional context

Writing to audit.log would be a a flanking measure and will get an issue of its own.

blizzz:
When I was discussing it with Marco, we agreed that a Primary Field as in descriptive identifiable column is useful and the best solution, but out of scope for this PR, and not achievable in that time frame. 

What could be used instead is some best effort attempt:
1. look for a simple text column (the first if there are more) and use it if available and not-empty, otherwise
2. look for a rich text column (the first if there are more) and use the first 20(?) characters of the first line, if available and not-empty, otherwise
3. use the ID

The ID in itself is not very helpful i agree, and should not be the long term solution, for now it may be acceptable as temporary approach.

PR #1801, opened 2025-05-12
Enhancement: Make tables import asynchronous

Koc:
Right now module not ready to import large tables. This PR improves import flow by moving it into asynchronous job.

## :spiral_notepad: TODO:
- [x] make import work for uploaded (not selected) files as well
- [x] revert changes to old endpoints, introduce new one instead
- [x] properly parse activity notification
- [x] fix pipelines and old tests

## :mag: Preview
Notification once import scheduled
image

Activity item with import stats
image

:movie_camera: Demo
https://github.com/user-attachments/assets/6be864ef-fb26-40a3-9e4c-c98e1d55419d

PR #2002, opened 2025-08-15
feat: activity support

luka-nextcloud:
Resolves: #1897


Koc:
can you please add import support :pray:? It can help me a lot for #1801

lib/Activity/ActivityManager.php
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	public function getActivityFormat($language, $subjectIdentifier, $subjectParams = [], $ownActivity = false) {
+		$subject = '';
+		$l = $this->l10nFactory->get(Application::APP_ID, $language);
+
+		switch ($subjectIdentifier) {
+			case self::SUBJECT_TABLE_CREATE:
+				$subject = $ownActivity ? $l->t('You have created a new table {table}'): $l->t('{user} has created a new table {table}');
+				break;
```
enjeck, line 210: Is this the right way to deal with translations with params? I figured it would be something using square brackets e.g `[table]`, as described in https://docs.nextcloud.com/server/latest/developer_manual/basics/translations.html#php-backend.

luka-nextcloud:
> @luka-nextcloud Are you done here? Because from reading #1897, there are some requirements listed there and I wonder if we will be covering them all here

@enjeck In my opinion, we are done for now. The remaining requirement is clean-up schedule (number 5 and 6) which I think should be discussed and done from the activity app.

lib/Activity/ActivityManager.php
```diff
@@ -0,0 +1,244 @@
...
+
+	public const SUBJECT_ROW_CREATE = 'row_create';
+	public const SUBJECT_ROW_UPDATE = 'row_update';
+	public const SUBJECT_ROW_DELETE = 'row_delete';
+
+	public function __construct(
+		private IManager $manager,
+		private IFactory $l10nFactory,
+		private TableMapper $tableMapper,
+		private Row2Mapper $rowMapper,
```
blizzz: unused

lib/Activity/ActivityManager.php
```diff
@@ -0,0 +1,244 @@
...
+use OCP\Activity\IEvent;
+use OCP\Activity\IManager;
+use OCP\L10N\IFactory;
+use OCP\Server;
+use Psr\Log\LoggerInterface;
+
+class ActivityManager {
+	public const SUBJECT_PARAMS_MAX_LENGTH = 4000;
+	public const SHORTENED_MAX_LENGTH = 2000;
+	public const ROW_NAME_MAX_LENGTH = 20;
```
blizzz: seem to be unusued?

lib/Activity/Filter.php
```diff
@@ -0,0 +1,76 @@
...
+
+namespace OCA\Tables\Activity;
+
+use OCP\Activity\IFilter;
+use OCP\IL10N;
+use OCP\IURLGenerator;
+
+class Filter implements IFilter {
+	private $l10n;
+	private $urlGenerator;
```
blizzz: provide type or use property promotion

lib/Activity/SettingChanges.php
```diff
@@ -0,0 +1,94 @@
...
+
+	/** @var IL10N */
+	protected $l;
+
+	/**
+	 * @param IL10N $l
+	 */
+	public function __construct(IL10N $l) {
+		$this->l = $l;
+	}
```
blizzz:
```suggestion
	public function __construct(protected IL10N $l) {
	}
```

lib/Activity/TablesProvider.php
```diff
@@ -0,0 +1,164 @@
...
+
+	public function __construct(
+		private $userId,
+		private IURLGenerator $urlGenerator,
+		private ActivityManager $activityManager,
+		private IUserManager $userManager,
+	) {
+	}
+
+	public function parse($language, IEvent $event, ?IEvent $previousEvent = null): IEvent {
```
blizzz, line 26:
```suggestion
	public function parse(string $language, IEvent $event, ?IEvent $previousEvent = null): IEvent {
```

lib/Helper/CircleHelper.php
```diff
@@ -104,4 +104,24 @@ public function getCircleIdsForUser(string $userId): ?array {
...
+			return [];
+		}
+
+		try {
+			$circle = $this->circlesManager->getCircle($circleId);
+			if ($circle === null) {
+				return [];
+			}
+			$members = $circle->getMembers();
+			return array_map(fn ($member) => $member->getUserId(), $members);
```
blizzz:
```suggestion
			return array_map(static fn ($member) => $member->getUserId(), $members);
```

lib/Service/ShareService.php
```diff
@@ -498,4 +498,35 @@ public function changeSenderForNode(string $nodeType, int $nodeId, string $newOw
...
+	 */
+	public function findSharedWithUserIds(int $elementId, string $elementType): array {
+		try {
+			$shares = $this->mapper->findAllSharesForNode($elementType, $elementId, '');
+			$sharedWithUserIds = [];
+
+			/** @var Share $share */
+			foreach ($shares as $share) {
+				if ($share->getReceiverType() === ShareReceiverType::USER) {
+					$sharedWithUserIds[] = $share->getReceiver();
```
blizzz: consider using the user id as key. Instead of `array_unique()`  later on, only `array_keys()` can be returned. Then  `array_flip()` may help with `array_merge`.

There might be a lot of users in the result set, this would take some short cuts in cpu and memory in such cases.

vite.config.ts
```diff
@@ -30,21 +30,6 @@ const config = createAppConfig({
...
-                    'micromark-util-resolve-all',
-                    'micromark-util-subtokenize',
-                    'micromark-util-decode-numeric-character-reference',
-                    'micromark-util-decode-string',
-					'micromark-util-html-tag-name',
-					'micromark-factory-label',
-					'micromark-factory-title',
-					'micromark-util-normalize-identifier',
-					'micromark-factory-whitespace',
-                ],
```
blizzz, line 47: Not sure what this removal is about, but if it is some cleanup of unnecessary things (smells like it!), all cool. I want to double check that this is not something unintended.

lib/Activity/ActivityManager.php
```diff
@@ -0,0 +1,244 @@
...
+				break;
+			case self::SUBJECT_ROW_UPDATE:
+				$columns = '';
+				foreach ($subjectParams['changeCols'] as $index => $changeCol) {
+					$columns .= '{col-' . $changeCol['id'] . '}';
+					if ($index < count($subjectParams['changeCols']) - 1) {
+						$columns .= ', ';
+					}
+				}
+				$subject = $ownActivity ? $l->t('You have updated cell(s) ' . $columns . ' on row {row} in table {table}') : $l->t('{user} has updated cell(s) ' . $columns . ' on row {row} in table {table}');
```
blizzz:
```suggestion
				$subject = $ownActivity ? $l->t('You have updated one or more cells ' . $columns . ' on row {row} in table {table}') : $l->t('{user} has updated one or more cells ' . $columns . ' on row {row} in table {table}');
```

The bracket is not that natural to read. But if we can figure out whether is one or more we can use the plural functions.

vite.config.ts
enjeck: I added this at https://github.com/nextcloud/tables/commit/a80e232764a2a7463a25dcdc412d2ebae948984c to fix vite build errors. I guess latest package updates make this no longer necessary. Nice to remove

blizzz:
> we can merge when the integration tests are fixed

they weren't broken actually, just some issue with the runners

PR #2018, opened 2025-08-20
fix: Scope CSS styles

enjeck:
To avoid these styles from leaking and affecting other apps


silverkszlo (commented):
`SmartPicker.vue` still contains the `picker-content` and `picker-title` classes, which don't exist anymore?

AndyScherzinger (approved):
🐘

PR #2019, opened 2025-08-20
chore: Added phpunit for the filtering function in Row2Mapper::findAll

AIlkiv:
These filters are mainly used for View.

image

In this PR, I covered the following column types with tests:
1. text
2. number
3. selection (single, check, multi)

In the future, I will try to add tests for other column types and check AND/OR operators.

### DIFF
--- a/tests/unit/Database/DatabaseTestCase.php
+++ b/tests/unit/Database/DatabaseTestCase.php
@@ -191,7 +191,12 @@ abstract class DatabaseTestCase extends TestCase {
 			'mandatory' => false,
 			'order_weight' => 0,
 			'number_prefix' => '',
-			'number_suffix' => ''
+			'number_suffix' => '',
+			'text_default' => '',
+			'number_default' => null,
+			'datetime_default' => '',
+			'selection_default' => '',
+			'usergroup_default' => '',
 		];
 
 		$testIdent = $data['test_ident'] ?? null;
@@ -316,26 +321,31 @@ abstract class DatabaseTestCase extends TestCase {
 	 */
 	protected function insertCellData(int $rowId, int $columnId, $value): void {
 		$qb = $this->connection->getQueryBuilder();
-		$result = $qb->select('type')
+		$result = $qb->select('type', 'subtype')
 			->from('tables_columns')
 			->where($qb->expr()->eq('id', $qb->createNamedParameter($columnId)))
 			->executeQuery();
 
-		$columnType = $result->fetchOne();
+		$column = $result->fetch();
 		$result->closeCursor();
 
-		if (!$columnType) {
+		if (!$column) {
 			throw new \InvalidArgumentException("Column with ID $columnId not found");
 		}
 
-		$this->insertCellIntoTypeTable($rowId, $columnId, $value, $columnType);
+		$this->insertCellIntoTypeTable($rowId, $columnId, $value, $column['type'], $column['subtype']);
 	}
 
 	/**
 	 * Inserts cell data into the appropriate type-specific table
 	 */
-	protected function insertCellIntoTypeTable(int $rowId, int $columnId, $value, string $columnType): void {
+	protected function insertCellIntoTypeTable(int $rowId, int $columnId, $value, string $columnType, string $columnSubtype): void {
 		$tableName = 'tables_row_cells_' . $columnType;
+
+		// Handle selection type - convert values to IDs based on selection_options
+		if ($columnType === 'selection' && $columnSubtype !== 'check') {
+			$value = $this->convertSelectionValuesToIds($columnId, $value);
+		}
 
 		$qb = $this->connection->getQueryBuilder();
 		$qb->insert($tableName)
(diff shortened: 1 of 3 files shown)


tests/unit/Db/Row2MapperTest.php
```diff
@@ -217,6 +244,133 @@ public static function sortingDataProvider(): array {
...
+				[[['columnId' => 'skills', 'operator' => 'is-equal', 'value' => ['@selection-id-9', '@selection-id-10']]],
+				['Bob'],  // Person with skills "Management" (id: 9) and "Communication" (id: 10)
+				'Filter by skills exactly equal to "Management" and "Communication" - should include 1 rows, exclude 4'
+			],*/
+			/* currently work only frontend
+			'Skills contains "Pyt"' => [
+				[[['columnId' => 'skills', 'operator' => 'contains', 'value' => 'Pyt']]],
+				['Alice', 'Charlie'],  // People with skills "Python" (id: 3)
+				'Filter by skills containing "Python" - should include 2 rows, exclude 3'
+			],*/
```
enjeck: If these don't work, no need keeping them? Might as well remove?

tests/unit/Database/DatabaseTestCase.php
```diff
@@ -403,6 +413,53 @@ protected function extractTestIdentMapping(array $results): array {
 		return $mapping;
 	}
 
+	/**
+	 * Converts selection values to IDs based on selection_options
+	 */
+	protected function convertSelectionValuesToIds(int $columnId, $value) {
```
enjeck, line 417: I wonder if this still works with https://github.com/nextcloud/tables/pull/1887. Gonna merge that shortly

tests/unit/Db/Row2MapperTest.php
```diff
@@ -217,6 +244,133 @@ public static function sortingDataProvider(): array {
...
+
+			// Number column (Experience Years)
+			'Experience years is equal to 5' => [
+				[[['columnId' => 'experience_years', 'operator' => 'is-equal', 'value' => 5]]],
+				['Alice'],  // Person with experience exactly 5 years
+				'Filter by experience years equal to 5 - should include 1 row, exclude 4'
+			],
+			'Experience years is greater than 5' => [
+				[[['columnId' => 'experience_years', 'operator' => 'is-greater-than', 'value' => 5]]],
+				['Eve'],  // People with experience 8, 7 years
```
enjeck: Should be this? 
```suggestion
				['Eve'],  // Person with experience 7 years
```

tests/unit/Db/Row2MapperTest.php
```diff
@@ -217,6 +244,133 @@ public static function sortingDataProvider(): array {
...
+				'Filter by experience years greater than 5 - should include 1 rows, exclude 4'
+			],
+			'Experience years is greater than or equal to 5' => [
+				[[['columnId' => 'experience_years', 'operator' => 'is-greater-than-or-equal', 'value' => 5]]],
+				['Alice', 'Eve'],  // People with experience 5, 8, 7 years
+				'Filter by experience years greater than or equal to 5 - should include 3 rows, exclude 2'
```
enjeck: Are we including 3 or 2? We have just Alice and Eve?

tests/unit/Db/Row2MapperTest.php
```diff
@@ -92,10 +102,14 @@ private function initializeTestData(): void {
...
 						'name' => 'Bob',
+						'surname' => 'Thompson',
 						'age' => 32,
 						'birthday' => '1991-12-03 14:20:00',
 						'department' => 'HR',
-						'score' => 92.0
+						'score' => 92.0,
+						'status' => 'Inactive',
+						'skills' => ['Management', 'Communication'],
+						'is_available' => '"false"',
```
enjeck: Is it intentional not to have `experience_years` for Bob?

tests/unit/Db/Row2MapperTest.php
```diff
@@ -217,6 +244,133 @@ public static function sortingDataProvider(): array {
...
+			],
+			'Experience years is lower than 5' => [
+				[[['columnId' => 'experience_years', 'operator' => 'is-lower-than', 'value' => 5]]],
+				['Bob', 'Charlie', 'Diana'],  // People with experience 2, 3 years
+				'Filter by experience years lower than 5 - should include 2 rows, exclude 3'
+			],
+			'Experience years is lower than or equal to 5' => [
+				[[['columnId' => 'experience_years', 'operator' => 'is-lower-than-or-equal', 'value' => 5]]],
+				['Alice', 'Bob', 'Charlie', 'Diana'],  // People with experience 2, 3, 5 years
+				'Filter by experience years lower than or equal to 5 - should include 3 rows, exclude 2'
```
enjeck: Also confused here

tests/unit/Db/Row2MapperTest.php
```diff
@@ -217,6 +244,133 @@ public static function sortingDataProvider(): array {
...
+			],
+			'Experience years is empty' => [
+				[[['columnId' => 'experience_years', 'operator' => 'is-empty', 'value' => '']]],
+				['Bob'],  // No person with empty experience years
+				'Filter by experience years is empty - should include 1 rows, exclude 4'
```
enjeck: Oh, so it's intentional for Bob not to have experience? So the comment saying `No person` is not correct?

tests/unit/Db/Row2MapperTest.php
AIlkiv: I think it makes sense to keep them.
Because we need to synchronize filters with the front end in order to paginate on the server in the future.

https://github.com/nextcloud/tables/issues/941

tests/unit/Database/DatabaseTestCase.php
AIlkiv: I did a rebase, it works

AIlkiv:
@enjeck I did a git rebase and made changes according to the comments.

tests/unit/Db/Row2MapperTest.php
enjeck: I prefer removing them in the meantime. Who knows when in the future we'll do anything, and it could get messy if we start commenting things like this

enjeck:
@AIlkiv So I rebased this branch onto master and kept only what is still unique here which are your selection support in DatabaseTestCase, and the selection filter cases, which now live in the existing Row2MapperFilterTest data provider instead of a separate block. I also had to load subtype in setupRealColumnMapper, otherwise multi-selection filtering is not exercised. Commented-out cases are removed for now.

enjeck (approved):

PR #2034, opened 2025-08-28
Update further icons to outline 

AndyScherzinger:
and use TashBin instead of Delete for a more Material Symbol look


enjeck (approved):
Nice. I updated one more

PR #2041, opened 2025-09-02
fix: Tiptap editor and heading styles

enjeck:
Fixes the headings and the rendering of the Tiptap editor. Having the styles in a "scoped" doesn't work. It's okay to leave it unscoped since we give a unique (hopefully) classname, and so shouldn't leak and affect other apps. 

### **Before**

Screenshot 2025-09-02 at 05 12 43

### **After**
Screenshot 2025-09-02 at 05 42 25

### DIFF
--- a/src/shared/components/ncTable/partials/TiptapMenuBar.vue
+++ b/src/shared/components/ncTable/partials/TiptapMenuBar.vue
@@ -216,9 +216,6 @@ export default {
 
 /* Basic editor styles */
 .tables-tiptap-wrapper .ProseMirror, .tabulator-cell .ProseMirror {
-	> * + * {
-		/*margin-top: 0.75em;*/
-	}
 
 	ul,
 	ol {


src/shared/components/ncTable/partials/TiptapMenuBar.vue
```diff
@@ -215,7 +215,7 @@ export default {
 }
 
 /* Basic editor styles */
-.tiptap-wrapper .ProseMirror, .tabulator-cell .ProseMirror {
+.tables-tiptap-wrapper .ProseMirror, .tabulator-cell .ProseMirror {
 	> * + * {
 		/*margin-top: 0.75em;*/
```
silverkszlo: do we still need this comment?

enjeck: Removed!

Issue #1980, opened 2025-08-06
Inline editing: prevent editing read-only values


enjeck:
As a follow up task, let's also make sure the read-only values are not editable with inline editing that's currently being added: https://github.com/nextcloud/tables/issues/364. A few column types already merged as of yesterday: https://github.com/nextcloud/tables/pull/1719

_Originally posted by @enjeck in https://github.com/nextcloud/tables/pull/1958#pullrequestreview-3090379155_

PR #2049, opened 2025-09-07
feat(Inline editing): prevent editing read-only values

enjeck:
fixes https://github.com/nextcloud/tables/issues/1980, and a follow up for https://github.com/nextcloud/tables/pull/1958. 
For inline editing in views, we prevent editing for read-only columns. 
We also gray it out:
Screenshot from 2025-09-07 09-02-01

### DIFF
--- a/src/shared/components/ncTable/partials/TableCellDateTime.vue
+++ b/src/shared/components/ncTable/partials/TableCellDateTime.vue
@@ -3,7 +3,7 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 <template>
-	<div class="cell-datetime">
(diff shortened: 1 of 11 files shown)


src/shared/components/ncTable/partials/TableCellDateTime.vue
```diff
@@ -3,7 +3,7 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 <template>
-	<div class="cell-datetime">
+	<div class="cell-datetime" :style="{ opacity: !canEditCell() ? 0.6 : 1 }">
```
silverkszlo, line 6: a title would be great here and in the other divs with similar functionality: `:title="!canEditCell() ? t('tables', 'Read only') : ''"`

PR #2050, opened 2025-09-07
fix: stylelint

enjeck:
Just ran `npm run stylelint:fix` to fix stylint error

### DIFF
--- a/css/grid.css
+++ b/css/grid.css
@@ -9,7 +9,7 @@
 
 .row.first-row {
 	padding-top: calc(var(--default-grid-baseline) * 4);
-	padding-left: 60px;
+	padding-inline-start: 60px;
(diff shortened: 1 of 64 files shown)


css/grid.css
```diff
@@ -9,7 +9,7 @@
 
 .row.first-row {
 	padding-top: calc(var(--default-grid-baseline) * 4);
-	padding-left: 60px;
+	padding-inline-start: 60px;
```
Koc, line 12: do we really need all of this changes? Will it better to tune eslint/prettier/whatever config to ignore that changes? Or is it mandatory for RTL support?

blizzz: This is related to RTL Support, yes: https://github.com/nextcloud-libraries/stylelint-config/pull/121

PR #2053, opened 2025-09-10
1769 views set mandatory state of a column

### LINKED ISSUE #1769: Views: set mandatory-state of a column

silverkszlo:
Resolves #1769 

Add the possibility to make columns in views mandatory.

TODO:

- [ ] add tests
- [ ] fix validation for mandatory datetime fields

Screenshot from 2025-09-15 12-15-31


lib/Service/RowService.php
```diff
@@ -327,9 +339,102 @@ private function cleanupAndValidateData(RowDataInput $data, array $columns, ?int
...
+			return $column->getSelectionDefault() !== null && $column->getSelectionDefault() !== '';
+		}
+		if ($column->getType() === 'selection-multi') {
+			if (is_array($value) && count($value) > 0) {
+				return true;
+			}
+			$defaultValue = $column->getSelectionDefault();
+			return $defaultValue !== null && $defaultValue !== '' && $defaultValue !== '[]';
+		}
+		return $value !== null && $value !== '';
```
enjeck: Doing all this to ensure that when row requests are made via the API, the mandatory condition is respected. Maybe there's an easier way to do it?

silverkszlo:
@enjeck and @blizzz I would appreciate another review of the latest commits beginning with that one: https://github.com/nextcloud/tables/pull/2053/commits/e309b7e6086dcdc5b7e31d8078ad1c3ad9cc65e9

cypress/e2e/view.cy.js
```diff
@@ -38,15 +40,10 @@ describe('Interact with views', () => {
...
-		cy.get('[data-cy="viewSettingsDialogSection"] input').type(title)
 	})
 
-	// cleanup
-	afterEach(function() {
+	// cleanup after all tests
+	after(function() {
 		// delete table (with view)
 		cy.get('[data-cy="navigationTableItem"]').contains('View test table').click({ force: true })
 		cy.get('[data-cy="customTableAction"] button').click()
```
enjeck, line 52: We do want to delete after each test. Otherwise, we have multiple views named the same and it's not obvious which is clicked /used for the various tests. We can avoid the deleting if we give all views different titles

cypress/e2e/view.cy.js
```diff
@@ -60,6 +57,13 @@ describe('Interact with views', () => {
 	})
 
 	it('Create view and insert rows in the view', () => {
+		cy.loadTable('View test table')
+
+		// create view
+		cy.get('[data-cy="customTableAction"] button').click()
+		cy.get('[data-cy="dataTableCreateViewBtn"]').contains('Create view').click({ force: true })
+		cy.get('[data-cy="viewSettingsDialogSection"] input').type(title)
+
```
enjeck, line 71: if we do this in all tests, why not keep in `beforeEach`?

cypress/e2e/view-mandatory-state.cy.js
enjeck, line 1: For just a mandatory state, I think this is way too much testing. Keeping in mind that too many tests bring a maintenance burden plus longer CI checks. It should be enough to have just one test case for confirming the core feature -- that we can set a mandatory column and it's respected, keeping it as minimal as possible

cypress/e2e/view.cy.js
silverkszlo: Oh yes, you're right! Since we are testing views and not tables here, I'll keep the logic of creating only one mock table for the file and adjust the individual titles per view.

cypress/e2e/view.cy.js
silverkszlo: I tried to explain this in my [comment](https://github.com/nextcloud/tables/pull/2053#issuecomment-3308386375) above. The tests were failing because they could not find `customTableAction`, which apparently requires `cy.loadTable` to be called first. For some reason, this did not work within the existing test structure and its `beforeEach` loops. Creating one table and multiple views now also seems reasonable to me.

cypress/e2e/view-mandatory-state.cy.js
silverkszlo: Okay, I'll reduce the amount of tests now

Issue #2089, opened 2025-09-30
Contributor week 2025 Brainstorming outcomes


GVodyanov:
## What do people use it for?

- HR
- Feature tracking and status
- Viewing and filtering data incoming from other applications after importing it as CSV files

## Bugs Found
- Bad contrast in row titles and the "metadata" tag because of styling variable changes in the last release
> <img width="418" height="550" alt="Image" src="https://github.com/user-attachments/assets/576c12c5-fcb3-48a1-b34e-7eb15756e793" />
- When editing rich text inline, some things are clipped (like when creating a heading the `#` icon). It also has a big black border it isn't supposed to have.
> <img width="701" height="299" alt="Image" src="https://github.com/user-attachments/assets/6a3938b3-4fff-4af0-bdb6-67e85131b676" />
- Sidebar duplicating
> <img width="355" height="301" alt="Image" src="https://github.com/user-attachments/assets/00662bcb-5d21-4f66-92a9-d1d61f705e26" />
- Dropdown in rich text editor clipped when creating a table
> <img width="829" height="457" alt="Image" src="https://github.com/user-attachments/assets/837b7d66-62b3-43d4-bb89-ec14d93ba4af" />

## Features requested (sorted by interest)
- _7 votes_ Add basic math operations and counting (like being able to sum or calculate averages on rows)
- _4 votes_ Decrease the amount of modals and dialogs in the app, requiring less clicks to get stuff done (for example, removing the modal needed to create a new row)
- _4 votes_ Add autosyncing with other apps in a user-friendly and discoverable way
- _3 votes_ Add an integration with collectives
- _2 votes_ Add Private Views, so views that you can create with the tables that are available to you and that aren't visible to other users
- _2 votes_ Add rule based conditional formatting for colors and other properties, similar to Google Sheets.
- _1 vote_ Add a full screen view and reduce padding around elements, making the app easier to use on small screens
- _1 vote_ Making deck a view of a table and vice versa (and in general have more integration of the apps into each other)
- Add a basic styling bar when editing rich text inline (maybe on hover?)
> <img width="1105" height="184" alt="Image" src="https://github.com/user-attachments/assets/f97da56b-de05-43df-a7c7-976819f7435b" />
- Add data types for phone numbers and emails (with links to relevant apps like talk, contacts, and calendar)
- Make table settings easier to use (remove actions menu and inline)
[...]

PR #2094, opened 2025-10-06
fix: Avoid duplicate navigation items

enjeck:
To reproduce problem:
- Select table or view
- refresh web page
- notice that there are two navigation entries for the active view/table
Problem also reported at https://github.com/nextcloud/tables/issues/2089

This PR fixes it by just removing duplicate entries in the store

### DIFF
--- a/src/store/store.js
+++ b/src/store/store.js
@@ -105,24 +105,10 @@ export const useTablesStore = defineStore('store', {
 			}
 		},
 		setTables(tables) {
-			// Deduplicate tables by ID to prevent navigation duplication
-			const uniqueTables = tables.reduce((acc, table) => {
-				if (!acc.find(t => t.id === table.id)) {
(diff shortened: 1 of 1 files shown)


benjaminfrueh:
Hi @enjeck and @blizzz
I just ran into this issue too when implementing a new feature, so I decided to look into it quickly.

The issue is caused in the store.js, in the functions loadContextTable() and loadContextView().
These functions try to select a table or view first, and then return if the view or table was found.
```js
const view = this.views.find(view => view.id === id)
if (view) {
    return true
}
```

The issue is, on page reload, the variable id is a string, not a integer anymore, so it is not able to find the view and duplicates it here with tables.push() and views.push().

I have not looked into why this happens yet or what change caused it, but this duplication could be prevented by making sure the id for searching a view or table is a integer, with e. g. parseInt().

Issue #2054, opened 2025-09-12
Inline editing polishing


marcoambrosini:
https://github.com/user-attachments/assets/9b132f4b-0912-4056-b37d-76cb98b94f0b

## General
- [ ] There should be no jump in the whole table when a user clicks in and out an inline editable field: https://github.com/nextcloud/tables/pull/2095
- [ ] Only the field that a user is currently editing should have a border, right now some field keep the border even when blurred
- [ ] Fields should activate edit mode with just one click: https://github.com/nextcloud/tables/pull/2095

## Text fields and input fields
- [ ] The border should coincide with the boundaries of the cell

## Rich text field
- [ ] We should not show drag handles or + icons in text fields that are not currently being edited: https://github.com/nextcloud/tables/pull/2095. Screnshot below:<img width="226" height="248" alt="Image" src="https://github.com/user-attachments/assets/0a202d62-468f-4d96-9723-44c3efc785e3" />

## Stars
- [ ] The stars shouldn't change in style and size, they should stay of --color-main-text and with the same distance: https://github.com/nextcloud/tables/pull/2095

## Switch
- [ ] Don't show scrollbars: https://github.com/nextcloud/tables/pull/2095
<img width="130" height="587" alt="Image" src="https://github.com/user-attachments/assets/42e05294-c130-43e4-8841-e744e2c76fa7" />

cc @enjeck @blizzz

PR #2095, opened 2025-10-06
enh: Polish inline editing fields

enjeck:
Fixes some of https://github.com/nextcloud/tables/issues/2054

- Richtext: There should be no jump in the whole table when a user clicks in and out an inline editable field
- Textline: Fields should activate edit mode with just one click
- Richtext: We should not show drag handles or + icons in text fields that are not currently being edited
- Stars: The stars shouldn't change in style and size, they should stay of --color-main-text and with the same distance
- Switch: Don't show scrollbars
- When editing rich text inline, some things are clipped (like when creating a heading the # icon). It also has a big black border it isn't supposed to have.

The last point was reported at https://github.com/nextcloud/tables/issues/2089

### DIFF
--- a/src/shared/components/ncTable/partials/TableCellStars.vue
+++ b/src/shared/components/ncTable/partials/TableCellStars.vue
@@ -11,14 +11,14 @@
 					:key="star"
 					class="star"
 					:class="{
-						'filled': star <= (hoverValue !== null ? hoverValue : editValue),
-						'clickable': !localLoading && canEditCell(),
+						'filled': star <= displayValue,
+						'clickable': isClickable,
 						'hovering': hoverValue !== null
 					}"
 					:aria-label="t('tables', 'Set {star} stars', { star })"
 					@mouseenter="hoverValue = star"
 					@click="setStar(star)">
-					{{ star <= (hoverValue !== null ? hoverValue : editValue) ? '★' : '☆' }}
+					{{ star <= displayValue ? '★' : '☆' }}
 				</span>
 			</div>
 			<div v-if="localLoading" class="icon-loading-small icon-loading-inline" />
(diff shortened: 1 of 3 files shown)


src/shared/components/ncTable/partials/TableCellStars.vue
```diff
@@ -4,76 +4,54 @@
...
-				<div v-if="localLoading" class="icon-loading-small icon-loading-inline" />
+		<div class="inline-editing-container">
+			<div class="interactive-stars"
+				@mouseleave="hoverValue = null">
+				<span v-for="star in 5"
+					:key="star"
+					class="star"
+					:class="{
+						'filled': star <= (hoverValue !== null ? hoverValue : editValue),
+						'clickable': !localLoading && canEditCell(),
```
silverkszlo: Minor refactoring suggestion: you could instead implement them as computed properties to cache the results and not trigger the calculation on each render. Might make a difference in bigger tables. Something like:

```
computed: {
		displayValue() {
			return this.hoverValue !== null ? this.hoverValue : this.editValue
		},

		isClickable() {
			return !this.localLoading && this.canEditCell()
		},
	},

**Comment by silverkszlo** [comment]
it seems that `truncate(string, num)` in `TableCellTextLine.vue` is not being used anymore?

File: src/shared/components/ncTable/partials/TableCellEditor.vue [hunk 2]
```diff
@@ -109,11 +109,24 @@ export default {
...
 
-.cell-editor > div {
+.cell-display-mode {
 	cursor: pointer;
 	min-height: 24px;
+	border: 2px solid transparent;
+	position: relative;
+
+	:deep(.content-wrapper) {
+		padding-left: 22px;
```
silverkszlo, line 122: Screenshot from 2025-10-15 12-57-48

that's a lot of padding-left in the richtext-column

enjeck: Necessary evil for the issue: "When editing rich text inline, some things are clipped (like when creating a heading the # icon" reported at https://github.com/nextcloud/tables/issues/2089

silverkszlo (approved):
Very nice, thank you! I have tested UI and it all seems to work.

PR #2127, opened 2025-10-30
fix: Update cypress, add cypress-split

enjeck:
Trying if using cypress-split lets cypress work better


enjeck:
> @copilot fix cypress failures

Lets see if this works, lol

PR #2136, opened 2025-11-03
build(dependencies): scope dependencies in lib/Vendor

### LINKED ISSUE #2133: Scope (at least phpoffice/phpspreadsheet) dependency/ies
This breaks the forms app as well as every installation of Nextcloud using 32bit architecture.

For forms: Dependencies can only be available once in PHP, so apps (tables and forms) need to use some kind of scoping ( maybe https://github.com/humbug/php-scoper ?). In this case tables uses v5 and forms v4 and it breaks.

For 32bit compatibility: v5 uses `maennchen/zipstream-php` v3 which is not compatible with 32bit.
As Nextcloud officially still supports 32bit the question is:
- can we ignore this for apps we maintain?
- if not you need to do the same as we did for forms: https://github.com/nextcloud/forms/pull/2600

_Originally posted by @susnux in https://github.com/nextcloud/tables/issues/2096#issuecomment-3464170384_

blizzz:
fixes #2133 

Introduces php-scoper to scope our production dependencies (incl. its dependencies :upside_down_face:)

Pretty much close to what I described at https://arthur-schiwon.de/isolating-nextcloud-app-dependencies-php-scoper once upon a time, with a few adjustments.
 
1. Added a script to update a list of production dependencies as well as committed that list. The idea here is to have a bit of overview when a prod dependency changes, including their sub-dependencies. A Github workflow as added to check this. We could do this on-the-fly, I'd prefer to be a bit defensive in the beginning.
2. I had to change the bin plugin to be a production dependency for the full flow to work (similar in other repos). This one is ignored from scoping and also ignore from release building.

:information_source:  When running this branch, or after merging to main, a `composer install [--no-dev]` becomes necessary.

### DIFF
--- a/vendor-bin/php-scoper/composer.json
+++ b/vendor-bin/php-scoper/composer.json
@@ -1,10 +1,23 @@
 {
-  "require": {
-    "humbug/php-scoper": "^0.18.7"
-  },
-  "config": {
-    "platform": {
-      "php": "8.1"
-    }
-  }
+	"require": {
+		"humbug/php-scoper": "^0.18.7",
+		"cweagans/composer-patches": "^2.0"
+	},
+	"config": {
+		"platform": {
+			"php": "8.1"
+		},
+		"allow-plugins": {
+			"cweagans/composer-patches": true
+		}
+	},
+	"extra": {
+		"patches": {
+			"thecodingmachine/safe": {
+				"PHP 8.4 compatibility, patch 1": "https://patch-diff.githubusercontent.com/raw/thecodingmachine/safe/pull/464.patch",
+				"PHP 8.4 compatibility, patch 2": "https://patch-diff.githubusercontent.com/raw/thecodingmachine/safe/pull/466.patch",
+				"PHP 8.4 compatibility, patch 3": "https://github.com/blizzz/safe/commit/6eeee2a20ffb2be3456ae8615b227c78e1ca8a20.patch"
+			}
+		}
+	}
 }
--- a/vendor-bin/php-scoper/composer.lock
+++ b/vendor-bin/php-scoper/composer.lock
@@ -4,8 +4,131 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "0ed0862265c8cad0416babb2030fb2f6",
+    "content-hash": "a6fa31156d91d29c66f9ca9a7bdc5996",
     "packages": [
+        {
+            "name": "cweagans/composer-configurable-plugin",
+            "version": "2.0.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/cweagans/composer-configurable-plugin.git",
+                "reference": "15433906511a108a1806710e988629fd24b89974"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/cweagans/composer-configurable-plugin/zipball/15433906511a108a1806710e988629fd24b89974",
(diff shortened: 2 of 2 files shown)


Koc (approved):

composer.json
```diff
@@ -1,55 +1,71 @@
...
+      "composer dump-autoload -o"
+    ],
+    "post-update-cmd": [
+      "composer bin all update --ansi",
+      "vendor/bin/php-scoper add-prefix --force",
+      "@php .lib-vendor-organizer.php lib/Vendor/ OCA\\\\Tables\\\\Vendor",
+      "cat .scoper-production-dependencies | uniq | xargs -I {} rm -Rf vendor/{} # Remove origins",
+      "cat .scoper-production-dependencies | cut -d / -f1 | uniq | xargs -I {} rmdir vendor/{} || true # Remove empty directories",
+      "composer dump-autoload -o"
+    ]
```
enjeck, line 64: At what point does one run these new commands? After installing any new library?

blizzz: with  `composer install` and  `composer update`, and effectively also with `composer require` and `uninstall`.  Cf. https://getcomposer.org/doc/articles/scripts.md

.lib-vendor-organizer.php
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+$sourceDirectory = $argv[1];
+$sourceDirectory = rtrim($sourceDirectory, '/') . '/';
+
+if (!str_starts_with('/', $sourceDirectory)) {
```
enjeck: Does this work properly? The format is `str_starts_with(string $haystack, string $needle)`, and here, `$sourceDirectory` is the haystack?

psalm.xml
```diff
@@ -21,10 +21,12 @@
         <directory name="lib" />
         <ignoreFiles>
             <directory name="vendor" />
+            <directory name="lib/Vendor" />
         </ignoreFiles>
 	</projectFiles>
 	<extraFiles>
 		<directory name="vendor" />
+		<directory name="lib/Vendor" />
 	</extraFiles>
```
enjeck, line 30: If I understand correctly,
```
<extraFiles>
		<directory name="vendor" />
		<directory name="lib/Vendor" />
Comment view	</extraFiles>
```
the above  specifies directories that Psalm should inspect while

```
<ignoreFiles>
            <directory name="vendor" />
            <directory name="lib/Vendor" />
        </ignoreFiles>
```
is the list of directories Psalm should load but not inspect.
Why do we need both?

composer.json
enjeck: Oooh, cool. What about for `scoper:update-deps`? When do I run it?

.lib-vendor-organizer.php
blizzz: good catch. Works so far [since Jan 2024](https://github.com/nextcloud/sharepoint/commit/74f3fa7c8ee6195f9dd5032b5a9158ea6ca873c1), but nevertheless it is a defect, yeah.

psalm.xml
blizzz: It should take them into account (so when we use them wrongly, it should shout out), but not evaluate them.

composer.json
blizzz: After installing, updating or removing a dependency. But CI will also tell you.

.lib-vendor-organizer.php
blizzz: fixed

PR #2181, opened 2025-12-03
fix: Specify min width for selection cell

enjeck:
Simple fix to avoid excessive jumping when inline editing. 
## Before
[Bildschirmaufzeichnung vom 2025-11-13 15-57-10.webm](https://github.com/user-attachments/assets/b546b677-1361-4e1a-a380-f4f65b31f61f)

## After

https://github.com/user-attachments/assets/ee7f882c-b5a0-4dee-8809-67b860937928

### DIFF
--- a/src/shared/components/ncTable/partials/TableCellSelection.vue
+++ b/src/shared/components/ncTable/partials/TableCellSelection.vue
@@ -152,6 +152,7 @@ export default {
 <style lang="scss" scoped>
 .cell-selection {
 	width: 100%;
+	min-width: 150px;
 
 	.non-edit-mode {
 		cursor: pointer;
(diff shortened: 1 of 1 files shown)


juliusknorr:
A few remarks:
- We have two "jumps" in the layout there, where i think one could be solved a bit more elegant
- The first happens when the plain rendering switches to the multiselect
  - Could we also give the non-edit rendering some padding so that ideally the text is staying in position
  - Enough padding to reserve space for the multiselect could already help for the first problem jumping
- The second happens when the multiselect is selected, i have not found the css rule that influences that on a quick check, but as soon as the focus is there the select has a `.vs--open` class which then leads to the label not taking up available space anymore

enjeck:
No longer an issue, from what I see

PR #2182, opened 2025-12-03
fix: improve number input validation and update handling

### LINKED ISSUE #2154: Input behaviour broken for numbers with defined min/max values
### Steps to reproduce

1. Create number column (no decimals) with defined min value of more than one digit - e.g. min = 10
2. Enter value of 70 in said column

### Expected behavior

Entered value is compared to min value when leaving entry field

OR

Entered value is compared to min value on every keystroke, when smaller than required min value, visual cue is applied, e.g. frame colour, text colour, comment text under entry field. 

### Actual behavior

Entered value is evaluated on each key stroke. Entered value replaced by min value after entering one digit. 

E.g. trying to enter “70”, i type ‘7’ (smaller than 10) and the ‘7’ i just typed is instantly replaced by the minimal value of “10”

### Tables app version

1.0.0

### Browser

Firefox Klar on ios

### Client operating system

_No response_

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

_No response_

enjeck:
fixes #2154

### DIFF
--- a/playwright/e2e/column-number.spec.ts
+++ b/playwright/e2e/column-number.spec.ts
@@ -53,24 +53,33 @@ test.describe('Test column number', () => {
 		await loadTable(page, tableTitle + ' second')
 		await createNumberColumn(page, columnTitle, '3.5', 1, 2, 20, 'PRE', 'SUF', true)
 
-		// insert row with default values
 		await page.locator('button').filter({ hasText: 'Create row' }).click()
 		await expect(page.locator('.modal__content input').first()).toHaveValue(/3.5/)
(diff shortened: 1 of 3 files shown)


enjeck:
Okay, should be fixed now:
- the check uses value < min and value > max, both inclusive, matching native HTML5 min/max. 
- there's no silent overwrite anymore.

blizzz (approved):
Lovely!

blizzz:
Locally, i have number tests failing, might need adjustement?

image

PR #2210, opened 2025-12-07
Enhancement: Apply user's timezone to meta fields

Koc:
Right now `createdAt/lastEditAt` displays UTC timezone. This makes Tables module non-user friendly for a non-technical people who live not in England :smile: . With this PR BE will continue store dates in UTC format without any transformations. At the same time FE handles:
- formatting from UTC to local browser's timezone during output
- formatting back from local browser's timezone to UTC before storing data


Koc:
indeed, I've changed approach. Now BE stores dates in UTC format without any transformations. At the same time FE handles:
- formatting from UTC to local browser's timezone during output
- formatting back from local browser's timezone to UTC before storing data

Work still in progress, I'm testing various scenarios

Issue #67, opened 2022-09-14
Share table with link


dosch:
Thnx for making this! It is 95% fixing a need we have!

We want to make a table where visitors can change values in the cells. 
Basically the need is for people to put their name in a field of their choice. And alter this in the future if they need to...

But for this I need to be able to share a table with outsiders, not only with people who already have an account on my instance. 
Is this possible?

Thnx already!

SublimeProphets:
Upvote for this issue :-)

I imagine a quiet simple implementation for the beginning: share a public link (like we are used to from files) and the possibility to set the right to read or read+write.

Thanks for your effort and great community work!

datenangebot:
Yes, I like to have this feature myself :+1: 
I put it on the roadmap.
But here is a lot to code and think, so it will not be the next upcoming feature...

lachmanfrantisek:
Another usease from me:

A (scout) course needs to store the results of tasks of attendees. Organisers need to see all the results, every participant can see just their row.

Now, I can create a nice table (compared to office document) and create a view for each participant. (It's manual but better than nothing. I can pick the columns I need and create a filter based on a name to show a single row.)
Now, the issue is to share this with the participants who do not need to have an account on my nextcloud instance.


![image](https://github.com/nextcloud/tables/assets/20214043/52e79597-8f3d-4f5f-be5e-fbd4a8c70a99)
![image](https://github.com/nextcloud/tables/assets/20214043/f9a41fab-4ecb-4427-90ed-ea841d68bcf8)

fnetX:
Just adding: In my environment (small to medium NGOs as well as university), most contributors don't have an account (e.g. because they are not formally employed by the NGO, not members of the university, or simply because they already have an account in their own Nextcloud and don't want to join so many). To me, this is how the Nextcloud ecosystem works (although I understand that this is likely different in companies).

Every feature which is impossible to share via link or federation simply does not work for us at all, so we cannot make use of most apps like this one.

To illustrate: One NGO I'm loosely affiliated with currently considered migrating from Microsoft Sharepoint to the Nextcloud (and we investigated using this app), but only two team members have an account (as part of the organization), while seven of them don't (as externals, volunteers, freelancers etc). Registering all of them to the Nextcloud is, unfortunately, not possible. Further, I suppose they would complain about needing "yet another account".

I hope this helps you to understand and classify the priority of this issue for uses of Nextcloud outside enterprise environments.

gchalakovmmi:
this is a great idea

PR #2211, opened 2025-12-08
enh(Sharing): backend infrastructre for read-only link shares

blizzz:
contributes to #67 

  - modifies oc_tables_share structure with two columns, token and password
  - adds ShareOCSController with a route to create link shares
  - adds a PageController front route to display the link share
  - adds a ApiPublicColumnsController to retrieve columns for public
      links. It was not added to the existing ApiColumnsController, as it
      requires the userId of the logged-in user and I did not want to weaken
      this detail.
  - adds an abstract controller for columns with shared functionality and
      make ApiColumnsController extend it.
  - adds a PublicRowOCSController for retrieving rows through link shares
  - adds a ShareToken value object
  - adds a ShareControlMiddleware for share token and existance
      validation. It comes with the AssertShareToken attribute.
  - extends Share entity with ShareToken and Password properties
  - extends ShareMapper to find a share by the share token
  - extends ShareService with a method to easily create link shares
  - extends ResponseDefinitions with TablesPublicRow and
      TablesPublicColumn specs. Essentially tableIDs are not exposed and
      also user ids in lastEditBy and createdBy are not disclosed.
  - extends RowService and ColumnService with methods to return such ^
      formatted result arrays.
  - extends OpenAPI spec

### Create a share link without password

```
curl -u user:secret -X POST https://cloud.example.com/ocs/v2.php/apps/tables/api/2/tables/${TABLE_ID}/share -H "OCS-APIRequest: true"

```

### fetch column information

```
curl -i https://cloud.example.com/ocs/v2.php/apps/tables/api/2/public/${SHARE_TOKEN}/columns -H "OCS-APIRequest: true"

```

### fetch all rows

```
curl -i https://cloud.example.com/ocs/v2.php/apps/tables/api/2/public/${SHARE_TOKEN}/rows -H "OCS-APIRequest: true"
```

### DIFF
--- a/cypress/e2e/tables-sharing-link.cy.js
+++ b/cypress/e2e/tables-sharing-link.cy.js
@@ -55,7 +55,7 @@ describe('Public link sharing', () => {
 			cy.get('[data-cy="customTableAction"] button').click()
 			cy.get('[data-cy="dataTableShareBtn"]').click()
 
-			cy.get('[data-cy="sharingEntryLinkTitle"]').should('be.visible')
+			cy.get('[data-cy="sharingEntryLinkTitle"]').scrollIntoView().should('be.visible')
 			cy.get('[data-cy="sharingEntryLinkDeleteButton"]').click()
 
 			cy.get('[data-cy="sharingEntryLinkTitle"]').should('not.exist')
@@ -106,7 +106,7 @@ describe('Public link sharing', () => {
 			cy.get('[data-cy="customTableAction"] button').click()
 			cy.get('[data-cy="dataTableShareBtn"]').click()
 
-			cy.get('[data-cy="sharingEntryLinkTitle"]').should('be.visible')
+			cy.get('[data-cy="sharingEntryLinkTitle"]').scrollIntoView().should('be.visible')
 			cy.get('[data-cy="sharingEntryLinkDeleteButton"]').click()
 
 			cy.get('[data-cy="sharingEntryLinkTitle"]').should('not.exist')


lib/Controller/PageController.php
```diff
@@ -72,4 +72,25 @@ public function context(int $contextId): TemplateResponse {
 
 		return $this->index();
 	}
+
+	#[PublicPage]
+	#[NoCSRFRequired]
+	#[OpenAPI(scope: OpenAPI::SCOPE_IGNORE)]
+	#[FrontpageRoute(verb: 'GET', url: '/s/{token}')]
+	public function linkShare(string $token): TemplateResponse {
+		Util::addScript(Application::APP_ID, 'tables-main');
```
blizzz: @enjeck may keep this if new functionality goes into the main script, or otherwise we can change this if course and load a different one.

The share token is being provided via initialState a few lines below.

lib/Controller/PageController.php
```diff
@@ -72,4 +72,25 @@ public function context(int $contextId): TemplateResponse {
...
+		$this->loadStyles();
+
+		$shareToken = new ShareToken($token);
+		$this->initialState->provideInitialState('shareToken', (string)$shareToken);
+
+		return new TemplateResponse(Application::APP_ID, 'main', [], TemplateResponse::RENDER_AS_GUEST);
```
enjeck: should this be `RENDER_AS_GUEST`? Or `RENDER_AS_PUBLIC`?

lib/Controller/PageController.php
```diff
@@ -72,4 +72,25 @@ public function context(int $contextId): TemplateResponse {
...
+	#[PublicPage]
+	#[NoCSRFRequired]
+	#[OpenAPI(scope: OpenAPI::SCOPE_IGNORE)]
+	#[FrontpageRoute(verb: 'GET', url: '/s/{token}')]
+	public function linkShare(string $token): TemplateResponse {
+		Util::addScript(Application::APP_ID, 'tables-main');
+		$this->loadStyles();
+
+		$shareToken = new ShareToken($token);
+		$this->initialState->provideInitialState('shareToken', (string)$shareToken);
```
enjeck: To properly load text editor so that rich text cells display properly, we need to add:
```
		if (class_exists(LoadEditor::class)) {
			$this->eventDispatcher->dispatchTyped(new LoadEditor());
		}
```

I already did at https://github.com/nextcloud/tables/pull/2236/changes#diff-70e5309fcd3311d771b3db9c93490ea270fd5894769093765ec37edb68e5dd9b

lib/Controller/ApiPublicColumnsController.php
```diff
@@ -0,0 +1,86 @@
...
+use OCP\AppFramework\Http\Attribute\OpenAPI;
+use OCP\AppFramework\Http\Attribute\PublicPage;
+use OCP\AppFramework\Http\DataResponse;
+use OCP\IL10N;
+use OCP\IRequest;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @psalm-import-type TablesPublicColumn from ResponseDefinitions
+ */
```
enjeck, line 33: we don't use `#[AssertShareToken]` here too?

lib/Controller/PublicRowOCSController.php
```diff
@@ -0,0 +1,91 @@
...
+	 * 200: Rows are returned
+	 * 400: Invalid request parameters
+	 * 403: No permissions
+	 * 404: Not found
+	 * 500: Internal error
+	 */
+	#[PublicPage]
+	#[AssertShareToken]
+	#[ApiRoute(verb: 'GET', url: '/api/2/public/{token}/rows', requirements: ['token' => '[a-zA-Z0-9]{16}'])]
+	#[OpenAPI]
```
enjeck, line 64: Since these are public, I imagine they are more prone to scraping and DoS? Do we want to throttle them?

lib/Service/RowService.php
```diff
@@ -66,6 +67,20 @@ public function formatRows(array $rows): array {
...
 	}
 
+	/**
+	 * @param Row2[] $rows
+	 * @return TablesPublicRow[]
+	 */
+	public function formatRowsForPublicShare(array $rows): array {
+		return array_map(static function (Row2 $row): array {
+			$rowData = $row->jsonSerialize();
+			unset($rowData['tableId']);
```
enjeck: Why do we unset the `tableId` but not the column/row ids? What makes tableId more riskier

lib/Controller/PageController.php
blizzz: You're totally right.

lib/Controller/ApiPublicColumnsController.php
blizzz: good catch!

lib/Service/RowService.php
blizzz: If we can prevent sending the row and column ids, we should do that. IIRC we need the column IDs as column meta data and rows are separate resourced and need to be linked together. Maybe we can hold back the row Id. 

But if we allow to have deep/anchor links to the specific table or view row, we have to provide it i think.

The basic idea is to leak as little internal data as possible. Even if table IDs are known, without authentication and access you should not be able to do anything about them. As they are being sequential at the moment, having them withdrawn is not a big protection either. 

That said, I am open to leave this item in place and send it along with the payload.

blizzz:
@enjeck does this help? 0078f8d40afff2bbef0774c263b7d601bcd6a6f7

enjeck (approved):
cool

PR #2212, opened 2025-12-09
feat: lazy initialization for editor setup

### LINKED ISSUE #2174: Inline rich text rendering has performance issues
@enjeck This was reported for our hub 26 winter table.

Every rich text is rendering using the text editor which can take some time, but should not be blocking i think.

I've also noticed that the text app does a request for each editor there to fetch task types:

Image

We should prevent fetching those requests and afterwards see if more optimization is needed

enjeck:
fixes #2174 

When there are many rich text cells, the page lags and while each editor is being loaded. This improves performance by only loading the cell when it's scrolled into the user's view

### DIFF
--- a/src/shared/components/ncEditor/NcEditor.vue
+++ b/src/shared/components/ncEditor/NcEditor.vue
@@ -65,7 +65,6 @@ export default {
 			localValue: '',
 			observer: null,
 			initialized: false,
-			idleHandle: null,
 		}
 	},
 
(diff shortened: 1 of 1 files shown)


lib/Controller/PageController.php
```diff
@@ -43,7 +43,6 @@ public function index(): TemplateResponse {
 		Util::addStyle(Application::APP_ID, 'grid');
 		Util::addStyle(Application::APP_ID, 'modal');
 		Util::addStyle(Application::APP_ID, 'tiptap');
-		Util::addStyle(Application::APP_ID, 'tables-style');
```
enjeck, line 46: unrelated to this feature, just removing since not needed

src/shared/components/ncEditor/NcEditor.vue
```diff
@@ -96,16 +99,56 @@ export default {
...
+		this?.editor?.destroy?.()
 	},
 
 	methods: {
 		t,
+		setupLazyInitialization() {
+			if (this.initialized) return
+
+			// Prefer initializing when the editor wrapper enters the viewport
+			if ('IntersectionObserver' in window) {
```
juliusknorr: Should be fine to assume that intersection observer is widely available https://caniuse.com/intersectionobserver

PR #2226, opened 2025-12-14
[main] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (1 unfixable, 22 only fixable manually using --force)


blizzz:
had to push two commits to fix the node version in both cypress workflows

PR #2236, opened 2025-12-28
feat: Public sharing via link

enjeck:
part of https://github.com/nextcloud/tables/issues/67

### TODO
- [x] Ensure password protected shares work one the backend implements that https://github.com/nextcloud/tables/pull/2211
- [x] Show the table/view name, emoji and description once I confirm we get it from the backend

### DIFF
--- a/lib/Controller/PublicSharePageController.php
+++ b/lib/Controller/PublicSharePageController.php
@@ -31,7 +31,7 @@ use OCP\Util;
 
 #[OpenAPI(scope: OpenAPI::SCOPE_IGNORE)]
 class PublicSharePageController extends AuthPublicShareController {
-	private Share $share;
+	private ?Share $share = null;
 	private ?ShareToken $shareToken = null;
 
 	public function __construct(
--- a/src/modules/sidebar/partials/SharingEntryLink.vue
+++ b/src/modules/sidebar/partials/SharingEntryLink.vue
@@ -283,7 +283,7 @@ export default {
 	}
 
 	&__avatar {
-		margin-right: 10px;
+		margin-inline-end: 10px;
 		flex-shrink: 0;
 	}
 
--- a/css/error.css
+++ b/css/error.css
@@ -18,9 +18,7 @@
   color: var(--color-main-text);
   background-color: var(--color-main-background-blur);
   padding: calc(3 * var(--default-grid-baseline));
-  border-radius: var(--border-radius-container);
   box-shadow: 0 0 10px var(--color-box-shadow);
-  display: inline-block;
   -webkit-backdrop-filter: var(--filter-background-blur);
   backdrop-filter: var(--filter-background-blur);
   display: flex;
--- a/src/modules/main/partials/editViewPartials/SelectedViewColumns.vue
+++ b/src/modules/main/partials/editViewPartials/SelectedViewColumns.vue
@@ -285,7 +285,7 @@ export default {
 }
 
 .row-elements.actions {
-	margin-left: auto;
+	margin-inline-start: auto;
 }
 
 .column-entry:hover .row-elements.move, .column-entry:focus-within .row-elements.move, .column-entry:hover .row-elements.actions, .column-entry:focus-within .row-elements.actions {
@@ -298,7 +298,7 @@ export default {
 
 .mandatory-indicator {
   color: var(--color-error);
-  margin-left: 4px;
+  margin-inline-start: 4px;
   font-size: 16px;
   line-height: 1;
 }
--- a/src/modules/sidebar/partials/SharingLinkList.vue
+++ b/src/modules/sidebar/partials/SharingLinkList.vue
@@ -77,6 +77,7 @@ export default {
 		font-size: 16px;
 	}
 }
+
 li {
     list-style-type: none;
 }
--- a/src/shared/components/ActivityEntry.vue
+++ b/src/shared/components/ActivityEntry.vue
@@ -118,17 +118,17 @@ export default {
 			flex-grow: 0;
 		}
 		.activity--subject {
-			margin-left: 10px;
+			margin-inline-start: 10px;
 		}
 		.activity--message {
-			margin-left: var(--default-clickable-area);
+			margin-inline-start: var(--default-clickable-area);
 			color: var(--color-text-light);
 			margin-bottom: 10px;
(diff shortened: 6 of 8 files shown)


src/modules/main/sections/PublicElement.vue
enjeck, line 1: This is for displaying both views and tables, despite the name. I should probably rename `view` to `element`

lib/Controller/PageController.php
```diff
@@ -78,13 +84,30 @@ public function context(int $contextId): TemplateResponse {
...
+
+			return new TemplateResponse(
+				Application::APP_ID,
+				'error',
+				['message' => $this->l->t('This share does not exist or is no longer available')],
+				TemplateResponse::RENDER_AS_PUBLIC
+			);
+		}
+
 		Util::addScript(Application::APP_ID, 'tables-main');
```
enjeck: I'll try using addStyle here to have custom styles, instead of the inline styles in error.php

src/App.vue
```diff
@@ -46,6 +49,12 @@ export default {
 	},
 	computed: {
 		...mapState(useTablesStore, ['isLoadingSomething', 'activeView', 'activeTable', 'activeContext']),
+		isPublicShare() {
+			return !!this.shareToken
+		},
+		shareToken() {
+			return loadState('tables', 'shareToken', false)
```
juliusknorr: loadState is not reactive so this does not need to be a computed property but can just go to data

To prepare for more advanced navigation on public pages, we could also use the token from the vue-router route, but may be a step to far already

src/modules/sidebar/partials/SharingEntryLink.vue
```diff
@@ -0,0 +1,345 @@
...
+						:checked.sync="usePassword"
+						@update:chk="onTogglePassword">
+						{{ t('tables', 'Password protection') }}
+					</NcActionCheckbox>
+				</div>
+
+				<div v-if="usePassword" class="sharing-entry-link__form-row">
+					<NcActionInput
+						:value.sync="password"
+						type="text"
```
juliusknorr: Should this be rather a type password? I think even in the files app that is not shown in plain text

juliusknorr (commented):
Went over the code and generally seems good, a few things to address still:

- We should probably not allow creating a share with a password when the password field is empty
- Design needs a bit of polishing so that the rows span the full width Image
- Editing should be blocked, you can currently inline edit, and requests are sent but they fail as expected
- Set password should probably be not checked by default (aligned with how it works in files)

lib/Controller/PageController.php
juliusknorr: Unrelated to this change but I think "tables-style" is not existent anymore. We can remove this and save a 404 request for a styles file that does not exist anymore. Can be seen in the browser development console

enjeck: Removed in another open ticket

enjeck:
Now looks like this:
Screenshot 2026-01-06 at 05 43 46

enjeck:
> The share link is not created correctly, at least when NC is installed in a subdir, see the duplicated `nextcloud` section:
> 
> `https://example.org/nextcloud/nextcloud/index.php/apps/tables/s/rXPTiveyDC5t4SID`
> 

Should be fixed now @blizzz

lib/Controller/ApiColumnsController.php
```diff
@@ -4,9 +4,11 @@
  * SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
+
```
blizzz, line 7: I think all the changes in this file are not necessary?

lib/Controller/PageController.php
```diff
@@ -32,6 +33,7 @@ public function __construct(
 		protected IInitialState $initialState,
 		protected ShareService $shareService,
 		protected NodeService $nodeService,
+		protected IL10N $l,
```
blizzz: We don't need it and its use clause here anymore

lib/Controller/PublicSharePageController.php
```diff
@@ -45,15 +45,23 @@ public function __construct(
 		private IEventDispatcher $eventDispatcher,
 	) {
 		parent::__construct($appName, $request, $session, $urlGenerator);
-		$this->shareToken = new ShareToken($this->getToken());
-		$this->share = $this->shareService->findByToken($this->shareToken);
+		$token = $request->getParam('token');
+		if (is_string($token) && $token !== '') {
+			$this->shareToken = new ShareToken($token);
+			try {
```
blizzz: Can we not have this try block? Or why do we need it? Properties should not be nullable ideally.

lib/Controller/PublicSharePageController.php
```diff
@@ -91,11 +99,13 @@ protected function getPasswordHash(): ?string {
 	}
 
 	public function isValidToken(): bool {
-		$this->shareToken;
-		return true;
+		return $this->share !== null;
```
blizzz, line 98: Ah, yeah, my code was a leftover after refactoring. Before it would fetch it, and succeed, or fail. If we need this than the nullable  approach is ok for now.

lib/ResponseDefinitions.php
```diff
@@ -112,6 +112,15 @@
...
  *
+ * @psalm-type TablesPublicNode = array{
+ *  title: string,
+ *  emoji: string,
+ *  description: string,
+ *  createdAt: string,
+ *  lastEditAt: string,
+ *  rowsCount: int,
+ * }
+ *
```
blizzz: goes away, openapi ci will complain when it is not used in the spec.

src/shared/utils/shareUtils.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+/**
+ * Checks if a share is a public link share
+ * @param {object} share
```
blizzz: there is some jsdoc warning ☝️ 🤷

blizzz (approved):
Works! Some cosmetic remarks. I skimmed front-end code only tbh.

PR #2238, opened 2025-12-28
Perf: Do not recreate `Cell` entity just for formatting value

Koc:
I've found that we are creating `Cell` entity just to call few methods of it. We can avoid of doing that and speedup `/row/table/{tableId}` endpoint up to 2x. For a table with 2k rows I've got a latency decrease from 1.15s to 0.568s with my PR.

| :heavy_minus_sign: Before | :heavy_plus_sign: After |
|--------|--------|
| image | image |

### DIFF
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -679,12 +679,10 @@ class Row2Mapper {
 
 			$column = $this->columnMapper->find($rowData['column_id']);
 			$columnType = $column->getType();
-			$cellClassName = 'OCA\Tables\Db\RowCell' . ucfirst($columnType);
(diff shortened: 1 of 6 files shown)


Koc:
I haven't profiled it deeply, but guess this caused by instance creation and calls tons of setters.

IMHO it's totally fine to operate with arrays here, as this is data access layer which produces read models (which are objects BTW). Part of the code which is responsible for write/mutations still operates entities.

@blizzz check referenced PRs to see more impressive speedup :smiley:

enjeck:
> I'd want to have a closer look. I am not happy resorting to a wild, unmaintainable array. Instantiation should not matter, the setters maybe.

Looks good to me, but I'll wait for @blizzz to approve

blizzz (approved):
With some stomach aches, but don't want to be the bottleneck. Changes are tenable.

PR #2242, opened 2025-12-28
Perf: Optimize performance of the get rows endpoints

### LINKED ISSUE #1490: Cannot open a table with 25k rows
### Steps to reproduce

1. Have a table with 25k rows (e.g. via CSV import)
2. Go to tables app
3. Open this table

### Expected behavior

The table is being loaded and values are being shown. Pagination is available.

### Actual behavior

`Could not load rows. Unknown error.` appears as error popup.

On the backend, an Exception is thrown: `An exception occurred while executing a query: SQLSTATE[HY000]: General error: 7 number of parameters must be between 0 and 65535`

### Tables app version

main

### Browser

_No response_

### Client operating system

_No response_

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

PostgreSQL

### Additional info

Actually the number of parameters is smaller, actually 25k row ids, plus ten column ids. However the query contains UNIONS and the row IDs are inserted 5 times in the total query – this would make sense again.

Excerpt from the log (omitting the 25k ids…):

```json
{
  "reqId": "jgnUNqgUUWi2A9hU5S7d",
  "level": 3,
  "time": "2024-12-04T16:39:10+00:00",
  "remoteAddr": "127.0.0.1",
  "user": "master",
  "app": "tables",
  "method": "GET",
  "url": "/master/index.php/apps/tables/row/table/830",
  "message": "An exception occurred while executing a query: SQLSTATE[HY000]: General error: 7 number of parameters must be between 0 and 65535",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64; rv:133.0) Gecko/20100101 Firefox/133.0",
  "version": "31.0.0.5",
  "exception": {
    "Exception": "OC\\DB\\Exceptions\\DbalException",
    "Message": "An exception occurred while executing a query: SQLSTATE[HY000]: General error: 7 number of parameters must be between 0 and 65535",
    "Code": 7,
    "Trace": [
      {
        "file": "/srv/http/nextcloud/master/lib/private/DB/ConnectionAdapter.php",
        "line": 53,
        "function": "wrap",
        "class": "OC\\DB\\Exceptions\\DbalException",
        "type": "::",
        "args": [
          {
            "__class__": "Doctrine\\DBAL\\Exception\\DriverException"
          }
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/apps-repos/tables/lib/Db/Row2Mapper.php",
        "line": 227,
        "function": "executeQuery",
        "class": "OC\\DB\\ConnectionAdapter",
        "type": "->",
        "args": [
          "SELECT `row_id`, `column_id`, `created_by`, `created_at`, `t1`.`last_edit_by`, `t1`.`last_edit_at`, `value`, `table_id`, `value_type` FROM (SELECT `row_id`, `column_id`, `last_edit_at`, `last_edit_by`, CAST(`value` AS TEXT) AS `value`, NULL AS `value_type` FROM `*PREFIX*tables_row_cells_text` WHERE (`column_id` IN (:columnIds)) AND (`row_id` IN (:rowsIds)) UNION ALL SELECT `row_id`, `column_id`, `last_edit_at`, `last_edit_by`, CAST(`value` AS TEXT) AS `value`, NULL AS `value_type` FROM `*PREFIX*tables_row_cells_number` WHERE (`column_id` IN (:columnIds)) AND (`row_id` IN (:rowsIds))  UNION ALL SELECT `row_id`, `column_id`, `last_edit_at`, `last_edit_by`, CAST(`value` AS TEXT) AS `value`, NULL AS `value_type` FROM `*PREFIX*tables_row_cells_datetime` WHERE (`column_id` IN (:columnIds)) AND (`row_id` IN (:rowsIds))  UNION ALL SELECT `row_id`, `column_id`, `last_edit_at`, `last_edit_by`, CAST(`value` AS TEXT) AS `value`, NULL AS `value_type` FROM `*PREFIX*tables_row_cells_selection` WHERE (`column_id` IN (:columnIds)) AND (`row_id` IN (:rowsIds))  UNION ALL SELECT `row_id`, `column_id`, `last_edit_at`, `last_edit_by`, CAST(`value` AS TEXT) AS `value`, CAST(`value_type` AS TEXT) AS `value_type` FROM `*PREFIX*tables_row_cells_usergroup` WHERE (`column_id` IN (:columnIds)) AND (`row_id` IN (:rowsIds)) ) `t1` INNER JOIN `*PREFIX*tables_row_sleeves` `rs` ON rs.id = t1.row_id",
          {
            "columnIds": [
              …
            ],
            "rowsIds": [
              …
            ]
          },
          {
            "columnIds": 101,
            "rowsIds": 101
          }
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/apps-repos/tables/lib/Db/Row2Mapper.php",
        "line": 180,
        "function": "getRows",
        "class": "OCA\\Tables\\Db\\Row2Mapper",
        "type": "->",
        "args": [
          [
            …
          ],
          [
            …
          ]
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/apps-repos/tables/lib/Service/RowService.php",
        "line": 86,
        "function": "findAll",
        "class": "OCA\\Tables\\Db\\Row2Mapper",
        "type": "->",
        "args": [
          [
            …
          ]
      },
      {
        "file": "/srv/http/nextcloud/master/apps-repos/tables/lib/Controller/RowController.php",
        "line": 45,
        "function": "findAllByTable",
        "class": "OCA\\Tables\\Service\\RowService",
        "type": "->",
        "args": [
          830,
          "master"
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/apps-repos/tables/lib/Controller/Errors.php",
        "line": 22,
        "function": "OCA\\Tables\\Controller\\{closure}",
        "class": "OCA\\Tables\\Controller\\RowController",
        "type": "->",
        "args": [
          "*** sensitive parameters replaced ***"
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/apps-repos/tables/lib/Controller/RowController.php",
        "line": 44,
        "function": "handleError",
        "class": "OCA\\Tables\\Controller\\RowController",
        "type": "->",
        "args": [
          {
            "__class__": "Closure"
          }
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 208,
        "function": "index",
        "class": "OCA\\Tables\\Controller\\RowController",
        "type": "->",
        "args": [
          830
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 114,
        "function": "executeController",
        "class": "OC\\AppFramework\\Http\\Dispatcher",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Tables\\Controller\\RowController"
          },
          "index"
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/lib/private/AppFramework/App.php",
        "line": 161,
        "function": "dispatch",
        "class": "OC\\AppFramework\\Http\\Dispatcher",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Tables\\Controller\\RowController"
          },
          "index"
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/lib/private/Route/Router.php",
        "line": 306,
        "function": "main",
        "class": "OC\\AppFramework\\App",
        "type": "::",
        "args": [
          "OCA\\Tables\\Controller\\RowController",
          "index",
          {
            "__class__": "OC\\AppFramework\\DependencyInjection\\DIContainer"
          },
          {
            "tableId": "830",
            "_route": "tables.row.index"
          }
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/lib/base.php",
        "line": 1012,
        "function": "match",
        "class": "OC\\Route\\Router",
        "type": "->",
        "args": [
          "/apps/tables/row/table/830"
        ]
      },
      {
        "file": "/srv/http/nextcloud/master/index.php",
        "line": 24,
        "function": "handleRequest",
        "class": "OC",
        "type": "::",
        "args": []
      }
    ],
    …
}
```

Koc:
This PR built on top of #2238.
Closes #1490.

It fixes loading of the large tables. Right now it is not possible to load a table with 30k rows and 8 columns due to an error:

```
"Previous":{"Exception":"Doctrine\\DBAL\\Exception\\DriverException","Message":
"An exception occurred while executing a query: SQLSTATE[HY000]: 
General error: 7 number of parameters must be between 0 and 65535",
```

PR contains 2 commits: fix itself and performance improvement.

I've measured latency for a various scenarios for `/row/table/{tableId}` endpoint:
| Scenario | master | #2238 | 1st commit from this PR | 2nd commit from this PR |
|--------|--------|--------|--------|--------|
| 8 columns x 2k rows | 0.68s | 0.31s | 0.31s | 0.17s |
| 8 columns x 10k rows | 2.86s | 1.21s | 1.21s | 0.44s |
| 8 columns x 60k rows | :lady_beetle: sql error | :lady_beetle: sql error | 6.97s | 2.27s |

📹 [Video](https://github.com/nextcloud/tables/pull/2242#issuecomment-3806357240) for performance comparison

Next step: move filtering/sorting/pagination to BE side and speedup FE. But this is completely separate topic which will be implemented in upcoming PRs.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -26,7 +26,7 @@ Share your tables and views with users and groups within your cloud.
 Have a good time and manage whatever you want.
 
 ]]></description>
-	<version>2.2.1</version>
+	<version>2.3.0-dev.0</version>
 	<licence>AGPL-3.0-or-later</licence>
 	<author>Nextcloud GmbH and Nextcloud contributors</author>
 	<namespace>Tables</namespace>
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -26,7 +26,7 @@ Share your tables and views with users and groups within your cloud.
 Have a good time and manage whatever you want.
 
 ]]></description>
-	<version>2.2.0</version>
+	<version>2.2.1</version>
 	<licence>AGPL-3.0-or-later</licence>
 	<author>Nextcloud GmbH and Nextcloud contributors</author>
 	<namespace>Tables</namespace>
@@ -58,6 +58,7 @@ Have a good time and manage whatever you want.
 		<post-migration>
 			<step>OCA\Tables\Migration\NewDbStructureRepairStep</step>
 			<step>OCA\Tables\Migration\DbRowSleeveSequence</step>
+			<step>OCA\Tables\Migration\CacheSleeveCells</step>
 		</post-migration>
 	</repair-steps>
 	<commands>
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -28,9 +28,9 @@ use Psr\Log\LoggerInterface;
 use Throwable;
 
 class Row2Mapper {
-	use TTransactional;
-
 	private const DB_CHUNK_SIZE = 1_000;
+
+	use TTransactional;
 
 	private RowSleeveMapper $rowSleeveMapper;
 	private ?string $userId;
@@ -124,10 +124,6 @@ class Row2Mapper {
 		return $rowSleeve->getTableId();
 	}
 
-	public function setUserId(string $userId): void {
-		$this->userId = $userId;
-	}
-
 	/**
 	 * @return int[]
 	 * @throws InternalError
@@ -203,14 +199,27 @@ class Row2Mapper {
 			return [];
 		}
 
-		$allRows = [];
-		foreach (array_chunk($rowIds, self::DB_CHUNK_SIZE) as $rowIdChunk) {
-			$allRows[] = $this->getRowsChunk($rowIdChunk, $columnIds);
+		$columnTypesCount = count($this->columnsHelper->columns);
+		if ($columnTypesCount === 0) {
+			return [];
 		}
-		return array_merge(...$allRows);
+
+		$columnsCount = count($columnIds);
+		$maxParamsPerType = floor(self::DB_CHUNK_SIZE / $columnTypesCount) ;
+		$calculatedChunkSize = (int)($maxParamsPerType - $columnsCount);
+		$chunkSize = max(1, $calculatedChunkSize);
+
+		$rowIdChunks = array_chunk($rowIds, $chunkSize);
+		$chunks = [];
+		foreach ($rowIdChunks as $chunkedRowIds) {
+			$chunks[] = $this->getRowsChunk($chunkedRowIds, $columnIds);
+		}
+
+		return array_merge(...$chunks);
 	}
 
 	/**
+	 * Builds and executes the UNION ALL query for a specific chunk of rows.
 	 * @param array $rowIds
 	 * @param array $columnIds
 	 * @return Row2[]
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -10,6 +10,8 @@ namespace OCA\Tables\Db;
 use DateTime;
 use DateTimeImmutable;
 use OCA\Tables\Constants\UsergroupType;
+use OCA\Tables\Db\RowLoader\CachedRowLoader;
+use OCA\Tables\Db\RowLoader\NormalizedRowLoader;
 use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Helper\ColumnsHelper;
@@ -18,18 +20,12 @@ use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\MultipleObjectsReturnedException;
 use OCP\AppFramework\Db\TTransactional;
 use OCP\DB\Exception;
-use OCP\DB\IResult;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
-use OCP\Server;
-use Psr\Container\ContainerExceptionInterface;
-use Psr\Container\NotFoundExceptionInterface;
 use Psr\Log\LoggerInterface;
 use Throwable;
 
 class Row2Mapper {
-	private const DB_CHUNK_SIZE = 1_000;
-
 	use TTransactional;
 
 	private RowSleeveMapper $rowSleeveMapper;
@@ -40,8 +36,15 @@ class Row2Mapper {
 	protected ColumnMapper $columnMapper;
 
 	private ColumnsHelper $columnsHelper;
+	/**
+	 * @var array<RowLoader\RowLoader::LOADER_*, RowLoader\RowLoader>
+	 */
+	private array $rowLoaders;
 
-	public function __construct(?string $userId, IDBConnection $db, LoggerInterface $logger, UserHelper $userHelper, RowSleeveMapper $rowSleeveMapper, ColumnsHelper $columnsHelper, ColumnMapper $columnMapper) {
+	public function __construct(?string $userId, IDBConnection $db, LoggerInterface $logger, UserHelper $userHelper, RowSleeveMapper $rowSleeveMapper, ColumnsHelper $columnsHelper, ColumnMapper $columnMapper,
+		NormalizedRowLoader $normalizedRowLoader,
+		CachedRowLoader $cachedRowLoader,
+	) {
 		$this->rowSleeveMapper = $rowSleeveMapper;
 		$this->userId = $userId;
 		$this->db = $db;
@@ -49,6 +52,10 @@ class Row2Mapper {
 		$this->userHelper = $userHelper;
 		$this->columnsHelper = $columnsHelper;
 		$this->columnMapper = $columnMapper;
+		$this->rowLoaders = [
+			RowLoader\RowLoader::LOADER_NORMALIZED => $normalizedRowLoader,
+			RowLoader\RowLoader::LOADER_CACHED => $cachedRowLoader,
+		];
 	}
 
 	/**
@@ -60,7 +67,7 @@ class Row2Mapper {
 		$this->db->beginTransaction();
 		try {
 			foreach ($this->columnsHelper->columns as $columnType) {
-				$this->getCellMapperFromType($columnType)->deleteAllForRow($row->getId());
+				$this->columnsHelper->getCellMapperFromType($columnType)->deleteAllForRow($row->getId());
 			}
 			$this->rowSleeveMapper->deleteById($row->getId());
 			$this->db->commit();
@@ -124,6 +131,10 @@ class Row2Mapper {
 		return $rowSleeve->getTableId();
 	}
 
+	public function setUserId(string $userId): void {
+		$this->userId = $userId;
+	}
+
 	/**
 	 * @return int[]
 	 * @throws InternalError
@@ -178,7 +189,7 @@ class Row2Mapper {
 			$wantedRowIdsArray = $this->getWantedRowIds($userId, $tableId, $filter, $sort, $limit, $offset);
 
 			// Get rows without SQL sorting
-			$rows = $this->getRows($wantedRowIdsArray, $showColumnIds);
+			$rows = $this->getRows($wantedRowIdsArray, $showColumnIds, RowLoader\RowLoader::LOADER_CACHED);
 
 			// Sort rows in PHP to preserve the order from getWantedRowIds
 			return $this->sortRowsByIds($rows, $wantedRowIdsArray);
@@ -191,92 +202,26 @@ class Row2Mapper {
 	/**
 	 * @param array $rowIds
 	 * @param array $columnIds
+	 * @param RowLoader\RowLoader::LOADER_* $loader
 	 * @return Row2[]
 	 * @throws InternalError
 	 */
-	private function getRows(array $rowIds, array $columnIds): array {
+	private function getRows(array $rowIds, array $columnIds, string $loader = RowLoader\RowLoader::LOADER_NORMALIZED): array {
 		if (empty($rowIds) || empty($columnIds)) {
 			return [];
 		}
 
-		$columnTypesCount = count($this->columnsHelper->columns);
-		if ($columnTypesCount === 0) {
-			return [];
+		$columns = $mappers = [];
+		foreach ($columnIds as $columnId) {
+			$columns[$columnId] = $this->columnMapper->find($columnId);
+			$mappers[$columnId] = $this->getCellMapper($columns[$columnId]);
 		}
 
-		$columnsCount = count($columnIds);
-		$maxParamsPerType = floor(self::DB_CHUNK_SIZE / $columnTypesCount) ;
-		$calculatedChunkSize = (int)($maxParamsPerType - $columnsCount);
-		$chunkSize = max(1, $calculatedChunkSize);
-
-		$rowIdChunks = array_chunk($rowIds, $chunkSize);
-		$chunks = [];
-		foreach ($rowIdChunks as $chunkedRowIds) {
-			$chunks[] = $this->getRowsChunk($chunkedRowIds, $columnIds);
+		$rows = [];
+		foreach ($this->rowLoaders[$loader]->getRows($rowIds, $columns, $mappers) as $row) {
+			$rows[] = $row;
 		}
-
-		return array_merge(...$chunks);
-	}
-
-	/**
-	 * Builds and executes the UNION ALL query for a specific chunk of rows.
-	 * @param array $rowIds
-	 * @param array $columnIds
-	 * @return Row2[]
-	 * @throws InternalError
-	 */
-	private function getRowsChunk(array $rowIds, array $columnIds): array {
-		$qb = $this->db->getQueryBuilder();
-
-		$qbSqlForColumnTypes = null;
-		foreach ($this->columnsHelper->columns as $columnType) {
-			$qbTmp = $this->db->getQueryBuilder();
-			$qbTmp->select('row_id', 'column_id', 'last_edit_at', 'last_edit_by')
-				->selectAlias($qb->expr()->castColumn('value', IQueryBuilder::PARAM_STR), 'value');
-
-			// This is not ideal but I cannot think of a good way to abstract this away into the mapper right now
-			// Ideally we dynamically construct this query depending on what additional selects the column type requires
-			// however the union requires us to match the exact number of selects for each column type
-			if ($columnType === Column::TYPE_USERGROUP) {
-				$qbTmp->selectAlias($qb->expr()->castColumn('value_type', IQueryBuilder::PARAM_STR), 'value_type');
-			} else {
-				$qbTmp->selectAlias($qbTmp->createFunction('NULL'), 'value_type');
-			}
-
-			$qbTmp
-				->from('tables_row_cells_' . $columnType)
-				->where($qb->expr()->in('column_id', $qb->createNamedParameter($columnIds, IQueryBuilder::PARAM_INT_ARRAY, ':columnIds')))
-				->andWhere($qb->expr()->in('row_id', $qb->createNamedParameter($rowIds, IQueryBuilder::PARAM_INT_ARRAY, ':rowsIds')));
-
-			if ($qbSqlForColumnTypes) {
-				$qbSqlForColumnTypes .= ' UNION ALL ' . $qbTmp->getSQL() . ' ';
-			} else {
-				$qbSqlForColumnTypes = '(' . $qbTmp->getSQL();
-			}
-		}
-		$qbSqlForColumnTypes .= ')';
-
-		$qb->select('row_id', 'column_id', 'created_by', 'created_at', 't1.last_edit_by', 't1.last_edit_at', 'value', 'table_id')
-			// Also should be more generic (see above)
-			->addSelect('value_type')
-			->from($qb->createFunction($qbSqlForColumnTypes), 't1')
-			->innerJoin('t1', 'tables_row_sleeves', 'rs', 'rs.id = t1.row_id');
-
-		try {
-			$result = $qb->executeQuery();
-		} catch (Exception $e) {
-			$this->logger->error($e->getMessage(), ['exception' => $e]);
-			throw new InternalError(get_class($this) . ' - ' . __FUNCTION__ . ': ' . $e->getMessage(), );
-		}
-
-		try {
-			$sleeves = $this->rowSleeveMapper->findMultiple($rowIds);
-		} catch (Exception $e) {
-			$this->logger->error($e->getMessage(), ['exception' => $e]);
-			throw new InternalError(get_class($this) . ' - ' . __FUNCTION__ . ': ' . $e->getMessage());
-		}
-
-		return $this->parseEntities($result, $sleeves);
+		return $rows;
 	}
 
 	/**
@@ -666,61 +611,6 @@ class Row2Mapper {
 	}
 
 	/**
-	 * @param IResult $result
-	 * @param RowSleeve[] $sleeves
-	 * @return Row2[]
-	 * @throws InternalError
-	 */
-	private function parseEntities(IResult $result, array $sleeves): array {
-		$rows = [];
-		foreach ($sleeves as $sleeve) {
-			$rows[$sleeve->getId()] = new Row2();
-			$rows[$sleeve->getId()]->setId($sleeve->getId());
-			$rows[$sleeve->getId()]->setCreatedBy($sleeve->getCreatedBy());
-			$rows[$sleeve->getId()]->setCreatedAt($sleeve->getCreatedAt());
-			$rows[$sleeve->getId()]->setLastEditBy($sleeve->getLastEditBy());
-			$rows[$sleeve->getId()]->setLastEditAt($sleeve->getLastEditAt());
-			$rows[$sleeve->getId()]->setTableId($sleeve->getTableId());
-		}
-
-		$rowValues = [];
-		$keyToColumnId = [];
-		$keyToRowId = [];
-		$cellMapperCache = [];
-
-		while ($rowData = $result->fetch()) {
-			if (!isset($rowData['row_id'], $rows[$rowData['row_id']])) {
-				break;
-			}
-
-			$column = $this->columnMapper->find($rowData['column_id']);
-			$columnType = $column->getType();
-			if (!isset($cellMapperCache[$columnType])) {
-				$cellMapperCache[$columnType] = $this->getCellMapperFromType($columnType);
-			}
-			$value = $cellMapperCache[$columnType]->formatRowData($column, $rowData);
-			$compositeKey = (string)$rowData['row_id'] . ',' . (string)$rowData['column_id'];
-			if ($cellMapperCache[$columnType]->hasMultipleValues()) {
-				if (array_key_exists($compositeKey, $rowValues)) {
-					$rowValues[$compositeKey][] = $value;
-				} else {
-					$rowValues[$compositeKey] = [$value];
-				}
-			} else {
-				$rowValues[$compositeKey] = $value;
-			}
-			$keyToColumnId[$compositeKey] = $rowData['column_id'];
-			$keyToRowId[$compositeKey] = $rowData['row_id'];
-		}
-
-		foreach ($rowValues as $compositeKey => $value) {
-			$rows[$keyToRowId[$compositeKey]]->addCell($keyToColumnId[$compositeKey], $value);
-		}
-
-		return array_values($rows);
-	}
-
-	/**
 	 * @throws InternalError
 	 */
 	public function isRowInViewPresent(int $rowId, View $view, string $userId): bool {
@@ -748,9 +638,12 @@ class Row2Mapper {
 		}
 
 		// write all cells to its db-table
+		$cachedCells = [];
 		foreach ($row->getData() as $cell) {
-			$this->insertCell($rowSleeve->getId(), $cell['columnId'], $cell['value'], $rowSleeve->getLastEditAt(), $rowSleeve->getLastEditBy());
+			$cachedCells[$cell['columnId']] = $this->insertCell($rowSleeve->getId(), $cell['columnId'], $cell['value'], $rowSleeve->getLastEditAt(), $rowSleeve->getLastEditBy());
 		}
+		$rowSleeve->setCachedCellsArray($cachedCells);
+		$this->rowSleeveMapper->update($rowSleeve);
 
 		return $row;
 	}
@@ -784,9 +677,12 @@ class Row2Mapper {
 		$this->columnMapper->preloadColumns(array_column($changedCells, 'columnId'));
 
 		// write all changed cells to its db-table
+		$cachedCells = $sleeve->getCachedCellsArray();
 		foreach ($changedCells as $cell) {
-			$this->insertOrUpdateCell($sleeve->getId(), $cell['columnId'], $cell['value']);
+			$cachedCells[$cell['columnId']] = $this->insertOrUpdateCell($sleeve->getId(), $cell['columnId'], $cell['value']);
 		}
+		$sleeve->setCachedCellsArray($cachedCells);
+		$this->rowSleeveMapper->update($sleeve);
 
 		return $row;
 	}
@@ -838,9 +734,11 @@ class Row2Mapper {
 	/**
 	 * Insert a cell to its specific db-table
 	 *
+	 * @return array<string, mixed> normalized cell data
+	 *
 	 * @throws InternalError
 	 */
-	private function insertCell(int $rowId, int $columnId, $value, ?string $lastEditAt = null, ?string $lastEditBy = null): void {
+	private function insertCell(int $rowId, int $columnId, $value, ?string $lastEditAt = null, ?string $lastEditBy = null): array {
 		try {
 			$column = $this->columnMapper->find($columnId);
 		} catch (DoesNotExistException $e) {
@@ -850,7 +748,7 @@ class Row2Mapper {
 
 		// insert new cell
 		$cellMapper = $this->getCellMapper($column);
-
+		$cachedCell = [];
 		try {
 			$cellClassName = 'OCA\Tables\Db\RowCell' . ucfirst($column->getType());
 			if ($cellMapper->hasMultipleValues()) {
@@ -862,6 +760,7 @@ class Row2Mapper {
 					$this->updateMetaData($cell, false, $lastEditAt, $lastEditBy);
 					$cellMapper->applyDataToEntity($column, $cell, $val);
 					$cellMapper->insert($cell);
+					$cachedCell[] = $cellMapper->toArray($cell);
 				}
 			} else {
 				/** @var RowCellSuper $cell */
@@ -871,11 +770,14 @@ class Row2Mapper {
 				$this->updateMetaData($cell, false, $lastEditAt, $lastEditBy);
 				$cellMapper->applyDataToEntity($column, $cell, $value);
 				$cellMapper->insert($cell);
+				$cachedCell = $cellMapper->toArray($cell);
 			}
 		} catch (Exception $e) {
 			$this->logger->error($e->getMessage(), ['exception' => $e]);
 			throw new InternalError('Failed to insert column: ' . $e->getMessage(), 0, $e);
 		}
+
+		return $cachedCell;
 	}
 
 	/**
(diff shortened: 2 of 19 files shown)


enjeck (commented):
Testing locally:
10k rows  before vs after I don't notice any speedups. How are you measuring the timing?
And 100k rows fails for both

lib/Db/RowLoader/CachedRowLoader.php
```diff
@@ -0,0 +1,88 @@
...
+			$id = (int)$sleeve['id'];
+			$row = new Row2();
+			$row->setId($id);
+			$row->setCreatedBy($sleeve['created_by']);
+			$row->setCreatedAt($sleeve['created_at']);
+			$row->setLastEditBy($sleeve['last_edit_by']);
+			$row->setLastEditAt($sleeve['last_edit_at']);
+			$row->setTableId((int)$sleeve['table_id']);
+
+			$cachedCells = json_decode($sleeve['cached_cells'] ?? '{}', true);
```
enjeck, line 72: Are we falling back to the normal way if nothing is returned from cached_cells somewhere else?

lib/Db/Row2Mapper.php
```diff
@@ -189,61 +198,27 @@ public function findAll(array $showColumnIds, int $tableId, ?int $limit = null,
...
 
-		try {
-			$result = $qb->executeQuery();
-		} catch (Exception $e) {
-			$this->logger->error($e->getMessage(), ['exception' => $e]);
-			throw new InternalError(get_class($this) . ' - ' . __FUNCTION__ . ': ' . $e->getMessage(), );
+		$columns = $mappers = [];
+		foreach ($columnIds as $columnId) {
+			$columns[$columnId] = $this->columnMapper->find($columnId);
+			$mappers[$columnId] = $this->columnsHelper->getCellMapperFromType($columns[$columnId]->getType());
```
enjeck: Looks like we run into the n+1 problem?

lib/Db/RowLoader/CachedRowLoader.php
Koc: no, it's not needed as we are caching all rows during migration, see `OCA\Tables\Migration\CacheSleeveCells` and we're updating `cached_cells` on any cell edit. So, I have no any use-case when `cached_cells` can be missing/outdated.

lib/Db/Row2Mapper.php
Koc: Not sure that this is a big problem here, as [we're calling](https://github.com/nextcloud/tables/blob/main/lib/Db/Row2Mapper.php#L174-L179) `ColumnMapper::preloadColumns()` before `Row2Mapper::getRows()` call and it [memoizes](https://github.com/nextcloud/tables/blob/main/lib/Db/ColumnMapper.php#L189-L192) loaded columns

lib/Migration/Version001010Date20251229000000.php
```diff
@@ -0,0 +1,41 @@
...
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 * @throws Exception
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('tables_row_sleeves');
+		if (!$table->hasColumn('tables_row_sleeves')) {
```
enjeck: I guess you meant this?

```suggestion
		if (!$table->hasColumn('cached_cells')) {
```

lib/Db/Row2Mapper.php
enjeck, line 1: This import path writes cells directly via the cell mapper, bypassing `Row2Mapper`, so `tables_row_sleeves.cached_cells` is never populated for imported rows?

lib/Db/RowLoader/CachedRowLoader.php
enjeck: `RowService::importRow` and `TablesMigrator::importRowCells` write sleeves and cells directly without setting `cached_cells`, and the repair step won't re-run (the `cachingSleeveCellsComplete` flag is already set), so those rows come back blank, i think. Is it better to fallback to NormalizedRowLoader?

lib/Migration/Version2230Date20260906000000.php
enjeck, line 1: we should update the migration name. It's been so long since this was started, new migrations have been added now

lib/Db/Row2Mapper.php
```diff
@@ -768,9 +670,12 @@ public function update(Row2 $row, ?string $userId = null): Row2 {
 		$this->columnMapper->preloadColumns(array_column($changedCells, 'columnId'));
 
 		// write all changed cells to its db-table
+		$cachedCells = $sleeve->getCachedCellsArray();
 		foreach ($changedCells as $cell) {
-			$this->insertOrUpdateCell($sleeve->getId(), $cell['columnId'], $cell['value']);
+			$cachedCells[$cell['columnId']] = $this->insertOrUpdateCell($sleeve->getId(), $cell['columnId'], $cell['value']);
 		}
+		$sleeve->setCachedCellsArray($cachedCells);
+		$this->rowSleeveMapper->update($sleeve);
```
enjeck, line 691: `update()` writes the sleeve twice per row edit. Can we fold the metadata change and the `cached_cells` update into a single `update()` call.

lib/Db/RowSleeveMapper.php
```diff
@@ -40,7 +40,7 @@ public function find(int $id): RowSleeve {
 
 	/**
 	 * @param int[] $ids
-	 * @return RowSleeve[]
```
enjeck, line 46: confused why raw data is better performance

lib/Migration/CacheSleeveCells.php
```diff
@@ -0,0 +1,105 @@
...
+
+	/**
+	 * @return int[]
+	 */
+	public function getTableIds(): array {
+		return $this->db->getQueryBuilder()
+			->select('id')
+			->from('tables_tables')
+			->orderBy('id')
+			->setMaxResults(PHP_INT_MAX)
```
enjeck: not sure why we need this

lib/Migration/CacheSleeveCells.php
```diff
@@ -0,0 +1,105 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
```
enjeck: should be `2026`

lib/Db/RowSleeveMapper.php
Koc: I guess the same reason like in #2238

lib/Migration/Version2230Date20260906000000.php
Koc: updated. Also bumped version from 2.2 to 2.3

lib/Db/RowLoader/CachedRowLoader.php
Koc: working on it now...

lib/Db/Row2Mapper.php
Koc: I didn't get the context. Is it the same as https://github.com/nextcloud/tables/pull/2242#discussion_r3441186645 ? Otherwise can you provide affected method names?

lib/Db/Row2Mapper.php
Koc: I did some changes here, testing now.

TBH would be nice to wrap all this db changes into a transaction. But this is completely separate topic

lib/Db/Row2Mapper.php
Koc: I did import via this flow and it caches everything fine with latest changes

image

Koc:
@enjeck @blizzz I did all requested changes, re-tested affected flows manually, we have green pipeline again. Please re-review one more time

lib/Db/RowLoader/CachedRowLoader.php
Koc: should be fixed now

appinfo/info.xml
```diff
@@ -26,7 +26,7 @@ Share your tables and views with users and groups within your cloud.
 Have a good time and manage whatever you want.
 
 ]]></description>
-	<version>2.2.0</version>
+	<version>2.3.0</version>
```
blizzz:
```suggestion
	<version>2.3.0-dev.0</version>
```

Would go for a pre-release number, unless we'd immediately release afterwards.

lib/Db/RowLoader/CachedRowLoader.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
```
blizzz, line 2:
```suggestion

declare(strict_types=1);

```

lib/Migration/Version2230Date20260906000000.php
blizzz: Actually the migration name needs to be so that it sorts beyond the last existing one: 

Image

Otherwise it would not be executed.

Yeah, we introduced a mistakes in the migration names 😔

lib/Migration/Version002030Date20260620000000.php
```diff
@@ -0,0 +1,41 @@
...
+	 * @return null|ISchemaWrapper
+	 * @throws Exception
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('tables_row_sleeves');
+		if (!$table->hasColumn('cached_cells')) {
+			$table->addColumn('cached_cells', Types::JSON, [
```
blizzz: We should avoid JSON types, as they work poorly across different DB vendors and caused trouble in the past. Use `Text` instead. `Types::JSON` is also officially deprecated for these reasons since NC 33.

lib/Migration/CacheSleeveCells.php
```diff
@@ -0,0 +1,104 @@
...
+							if ($cellMapper->hasMultipleValues()) {
+								$cachedCells[$column->getId()][] = $cellMapper->toArray($cell);
+							} else {
+								$cachedCells[$column->getId()] = $cellMapper->toArray($cell);
+							}
+						}
+					}
+
+					$sleeve = $this->rowSleeveMapper->find($rowId);
+					$sleeve->setCachedCellsArray($cachedCells);
```
blizzz, line 67: Correct me if I am wrong. This stores all cell values of a how table in a single row? On MySQL this would be limited to whopping 4GB, but generally, this sounds risky and could lead to transferal of a huge amount of data on really big tables. What do you think?

blizzz: Also as a post-migration step, this might take a real long time to accomplish on instanced with a big amount of data.

lib/UserMigration/TablesMigrator.php
```diff
@@ -335,6 +340,39 @@ private function importRowCells(array $data, array $rowIdMap, array $columnIdMap
...
+				$this->rowCellUsergroupMapper,
+			];
+
+			foreach ($cellMappers as $mapper) {
+				try {
+					$cells = $mapper->findAllForRow($newRowId);
+					foreach ($cells as $cell) {
+						$cachedCells[$cell->getColumnId()] = $mapper->toArray($cell);
+					}
+				} catch (\Exception $e) {
```
blizzz: It's a too broad catch, isn't it?

lib/UserMigration/TablesMigrator.php
```diff
@@ -335,6 +340,39 @@ private function importRowCells(array $data, array $rowIdMap, array $columnIdMap
 		}
 	}
 
+	private function rebuildCachedCells(array $rowIdMap): void {
+		foreach ($rowIdMap as $newRowId) {
+			$sleeve = $this->rowSleeveMapper->find($newRowId);
+			$cachedCells = [];
+
+			/** @var RowCellMapperSuper[] $cellMappers */
+			$cellMappers = [
```
blizzz: doesn't need to be within the loop?

lib/UserMigration/TablesMigrator.php
```diff
@@ -335,6 +340,39 @@ private function importRowCells(array $data, array $rowIdMap, array $columnIdMap
 		}
 	}
 
+	private function rebuildCachedCells(array $rowIdMap): void {
+		foreach ($rowIdMap as $newRowId) {
```
blizzz, line 353: Especially if this is running in user facing scripts, this may run into a timeout on big tables and fail.

lib/Db/RowSleeveMapper.php
```diff
@@ -49,14 +49,16 @@ public function findMultiple(array $ids): array {
 		$qb->select(
 			$sleeveAlias . '.id',
 			$sleeveAlias . '.table_id',
+			$sleeveAlias . '.cached_cells',
 			$sleeveAlias . '.created_by',
```
blizzz, line 56: always?

blizzz (changes requested):
I have a few questions and scaling concerns (also nitpicks added 😇 )

lib/Migration/Version2230Date20260906000000.php
Koc: fixed

lib/Db/RowLoader/CachedRowLoader.php
Koc: fixed here and in other files

lib/UserMigration/TablesMigrator.php
Koc: replaced with db exception

lib/UserMigration/TablesMigrator.php
Koc: sorry, I didn't get. What do you mean?

lib/Db/RowSleeveMapper.php
Koc: that's ok, we have 2 scenarios:
- load multiple rows in optimized way
  - for this case we need to load cached cells
- load single way in legacy EAV way
  - this is either read of a single row which will not affect performance
  - either this is load for update which is also fine as we can update this field, so we need to get actual value for it

lib/Migration/CacheSleeveCells.php
Koc: > This stores all cell values of a how table in a single row? 

yeah. This means that loading of the cells no need to dynamic query with union 

> On MySQL this would be limited to whopping 4GB, but generally, this sounds risky and could lead to transferal of a huge amount of data on really big tables. 

from my PoV 4Gb is more than enough. Regarding traffic - I'm not sure that this is a big deal as we're loading same amout of data from plenty of tables using `UNION`

> Also as a post-migration step, this might take a real long time to accomplish on instanced with a big amount of data.

I guess that this is a reason why post-migration steps exists. And this is one-time operation during update.  But I'm fine to any other solution, just say how to you see that (e.g. we can opt-in this feature, dispatch async job to cache cell values for the table).

lib/UserMigration/TablesMigrator.php
blizzz: The `$cellMappers` can be defined ones in front of the `foreach` loop.

lib/Migration/CacheSleeveCells.php
blizzz: > from my PoV 4Gb is more than enough.

I also expect it is in 99% of cases or more. And currently, we are not able to deal with sizes that are way smaller. So from that point, ok for now.

> Regarding traffic - I'm not sure that this is a big deal as we're loading same amout of data from plenty of tables using UNION

As long as we are indeed not loading more than that it is fine. If we potentially load more (comment https://github.com/nextcloud/tables/pull/2242#discussion_r3459561212 was about it), then we may increase the footprint too much. And it has to share the bandwidth with all other apps and components as well, so it could help clogging.

That's a real scenario already :upside_down_face: 

> I guess that this is a reason why post-migration steps exists. And this is one-time operation during update. But I'm fine to any other solution, just say how to you see that (e.g. we can opt-in this feature, dispatch async job to cache cell values for the table).

The post-migration step still runs live and sequentially, hence is blocking. The very least would be outputting some progress information, as otherwise the admin may think the process hangs.

Possible alternatives are `live-migration` where the repair step is done in production, startet as a regualar background job.

Better might be the expensive repair step that runs only during maintenance hours, and is also background-job triggered. https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/repair.html#expensive-repair-steps

Both require that it is not critical that the field is filled for a while without interruption of service.

enjeck (commented):
if all comments are addressed, can be merged after conflict is fixed?

Koc:
@enjeck I don't see any blockers. I did the rebase, fixed all conflicts, tested manually few CRUD operations and upgrade - everything works for me

lib/Migration/CacheSleeveCells.php
Koc: @blizzz I've moved caching of already existent cells to `live-migration`. Also we fallback to old way of rows loading till caching not completed.

blizzz:
@Koc loading views gives me a 500, likely related to them containing a meta column.

lib/Db/Row2Mapper.php
```diff
@@ -827,9 +745,20 @@ public function update(Row2 $row, ?string $userId = null): Row2 {
...
+				$sleeve->getId(),
+				$this->columnMapper->findAllByTable($sleeve->getTableId())
+			);
+		}
 		foreach ($changedCells as $cell) {
-			$this->insertOrUpdateCell($sleeve->getId(), $cell['columnId'], $cell['value']);
+			$cachedCells[$cell['columnId']] = $this->insertOrUpdateCell($sleeve->getId(), $cell['columnId'], $cell['value']);
 		}
+		$sleeve->setCachedCellsArray($cachedCells);
+		$this->rowSleeveMapper->update($sleeve);
```
blizzz: since we have two updates that are related, they should be bundled in a transaction, as you pointed out. `OCP\AppFramework\Db\TTransactional->atomic()`  can be used for convenience.

Also valid for insert logic, not just update.

blizzz:
<details><summary>Other yet unverified findings; unfold to read.</summary>

## Potentially Blocking

### B1 — Lost update on concurrent edits

`Row2Mapper::update()` is a read-modify-write of the whole blob: read at line 748, cell writes in
between, write back at 760-761. No lock, no version check (`QBMapper::update()` is a plain
`UPDATE … WHERE id = ?`).

Nextcloud pins the primary connection to **READ COMMITTED**
(`OC\DB\SetTransactionIsolationLevel::postConnect()`), so a transaction alone does **not** fix this:

```
T1  BEGIN; SELECT cached_cells → {5:"a", 6:"x"}
T2  BEGIN; SELECT cached_cells → {5:"a", 6:"x"}
T1  write cell col 5; UPDATE … SET cached_cells = {5:"b", 6:"x"}; COMMIT
T2  write cell col 6; UPDATE … SET cached_cells = {5:"a", 6:"y"}; COMMIT
```

Final cache `{5:"a", 6:"y"}` — T1's change to column 5 is gone from the cache while
`tables_row_cells_text` correctly holds `"b"`. Both committed, neither errored. MySQL REPEATABLE
READ does not help either (non-locking snapshot reads); only SERIALIZABLE would, and NC does not
run it.

Fix: re-read the sleeve with `IQueryBuilder::forUpdate()` inside the transaction before line 748
(`@since 33.0.0`, matches the app's `min-version`). Needs a `RowSleeveMapper::findForUpdate()`.
Alternative: optimistic version column with retry on 0 affected rows.

Collaborative editing is the point of this app, so this is reachable in normal use.

### B2 — No reconciliation path

Once the cache diverges there is no way back. `CacheSleeveCells` only picks up rows
`WHERE cached_cells IS NULL`, which a stale row is not. There is no `occ` command to rebuild.
Given B1 and M4/M5 below, ship a rebuild command with this.

---

## Architecture — worth settling before merging

**The join in the old query is dead weight, and removing it may be most of the win.**
`NormalizedRowLoader::getRowsChunk()` selects `created_by, created_at, t1.last_edit_by,
t1.last_edit_at, table_id` and joins `tables_row_sleeves` — but `parseResult()` never reads any of
them. All row metadata comes from the separate `findMultiple($rowIds)` call, and every
`formatRowData()` implementation only touches `value` / `value_type`. That join forces the UNION to
be materialized into a temp table and joined per cell row. There is already an index on
`(column_id, row_id)` (`Version000700Date20230916000000.php:82`).

Dropping the join — or going further and issuing five plain indexed
`SELECT row_id, column_id, value FROM tables_row_cells_X WHERE column_id IN (…) AND row_id IN (…)`
queries merged in PHP, with no `CAST` and no UNION — would be a fraction of this diff with no
schema change, no dual-write, no migration and no consistency risk. Worth benchmarking before
accepting a denormalized copy of every cell in the database.

If the cache is kept: it roughly doubles row-data storage, and `CachedRowLoader` transfers the
entire row from the DB even for a view showing 2 of 50 columns (filtered in PHP at
`CachedRowLoader.php:59-73`) — a bandwidth/memory regression exactly for the wide tables this
targets. Per AGENTS.md this warrants a ticket for maintainer alignment first.

---

## Medium

### M1 — Fresh installs stay on the slow path until the next app update

`live-migration` steps are only queued by `AppManager::upgradeApp()` (`AppManager.php:1156`).
`Installer::installAppLastSteps()` runs only `install` / `pre-migration` / `post-migration`
(`Installer.php:545,551,573`) and then sets `installed_version` to the current version.

A fresh install therefore never flips `cachingSleeveCellsComplete` and `getRowLoader()`
(`Row2Mapper.php:203`) keeps returning `LOADER_NORMALIZED`. Writes are unaffected —
`insert()`/`update()` always populate `cached_cells` — so nothing is wrong, the install just runs
slow while it accumulates data. It self-heals at the next release the admin installs; indefinite if
the version is pinned.

Fix: also list the step under `<install>` in `info.xml`, or flip the flag when nothing is pending.

Separately, confirm which server version introduced `live-migration` support. The app declares
`min-version="33"`; the server commit is from Aug 2026 and only `master`/36 was available to check,
so it is clearly not in 33/34. Either raise `min-version` or handle its absence.

### M2 — The backfill is an N+1 in a loop (AGENTS.md forbids this)

Three call sites, all through `ColumnsHelper::getCachedCellsForRow()`
(`lib/Helper/ColumnsHelper.php:141-155`), which runs one query per column and builds the query
builder inside the loop (`RowCellMapperSuper::findManyByRowAndColumn()`,
`lib/Db/RowCellMapperSuper.php:129-136`):

- **Backfill** — `CacheSleeveCells.php:59-63`, with `find()` + `update()` per row at lines 74-85.
  Cost is `rows × (columns + 2)`: a 100k-row × 20-column table is ~2.2M queries in one
  `BackgroundRepair` run. It is at least resumable (the `IS NULL` filter survives a kill).
- **Every edit of a row whose cache reads empty** — `Row2Mapper.php:748-756` (see M3).
- **User migration import** — `TablesMigrator.php:351`, per row, inside the import transaction.

Fix for all three: one query per cell table over a batch of row IDs (`WHERE row_id IN (…)`), which
is what the `(column_id, row_id)` index is for.

### M3 — The rebuild also fires on ordinary empty rows

`Row2Mapper.php:749` treats `$cachedCells === []` as "not yet migrated", but
`getCachedCellsArray()` returns `[]` for both `NULL` and `"[]"` — i.e. any genuinely empty row.
Filling the first cell of a new row triggers `findAllByTable()` plus one query per column, every
time, forever. Distinguish `null` from `[]`.

### M4 — Rows the backfill cannot reach become silently empty

`CacheSleeveCells::getTableIds()` snapshots `tables_tables` at the start and only backfills sleeves
whose `table_id` matches, then sets the flag unconditionally (line 68). Any sleeve with a
NULL/orphaned `table_id` is skipped but still read through the cached loader, where
`json_decode($sleeve['cached_cells'] ?? '{}')` (`CachedRowLoader.php:72`) yields nothing and the row
renders with no cells.

Fix: `CachedRowLoader` should fall back to the normalized loader when `cached_cells IS NULL` rather
than returning an empty row.

### M5 — `importRow()` writes `'[]'` before the cells exist

`RowService.php:1026` inserts `cached_cells = json_encode([])`, which makes those rows invisible to
the backfill's `IS NULL` filter. `TablesMigrator::rebuildCachedCells()` covers it, but it catches
`\Throwable` and only logs (`TablesMigrator.php:354`), so a failure leaves a permanently-wrong row
with no way back. Insert `NULL` instead.

### M6 — `find()` silently uses the slow loader

`Row2Mapper::find()` (line 89) calls `getRows()` without a loader argument, so it takes the
`LOADER_NORMALIZED` default. If deliberate, say so in a comment; as written the default parameter
reads like an oversight. It also means post-write API responses come from the normalized tables
while the grid comes from the cache — which is exactly what hides B1 from testing.

### M7 — Normalized chunk size drops from 1000 to ≤200 rows

`NormalizedRowLoader::getRows()` computes `floor(1000/5) - count($columns)`. Correct fix for the
placeholder explosion on `main` (the named `:rowsIds` is expanded once per UNION branch), but it is
5× more round trips in the fallback path — the only path until the background job runs, and the
permanent one in M1. Also, with ≥200 columns `$chunkSize` clamps to 1 and `5 × (200+1)` still
exceeds the 1000 limit.

### M8 — Test fixture does not match the production format

`DatabaseTestCase::updateCachedCells()` writes `['value' => …]` for every column type, but usergroup
columns are `hasMultipleValues()` and production stores a list of `['value', 'value_type']` pairs.
`CachedRowLoader`'s `array_map` branch (lines 65-67) is never exercised against realistic data.

---

## Minor

- `RowCellMapperSuper::findAllForRow()` (line 142) is dead code — never called. AGENTS.md forbids it.
- `CacheSleeveCells` declares `strict_types=1` and passes `fetchAll(\PDO::FETCH_COLUMN)` results
  straight into `cacheCellsForRow(int $rowId)`. Whether PDO returns `int` or `string` is
  driver/emulation dependent; `Row2Mapper` gets away with the same pattern only because it has no
  `strict_types`. Add an explicit `(int)`.
- `findManyByRowAndColumn()` documents `@throws MultipleObjectsReturnedException|DoesNotExistException`;
  `findEntities()` throws neither.
- `get_class($this)` replaces `static::class` in several new/edited error messages, inconsistent with
  the rest of `Row2Mapper`.
- `RowSleeveMapper::findMultiple()` no longer calls `closeCursor()`.
- `setCachedCellsArray(array $cachedCells):void` — missing space before the return type.
- Migration named `Version2230Date20260906000000` while `info.xml` goes to `2.3.2`; the AGENTS.md
  convention gives `Version2320…`. Sorts fine after `Version2210`, so cosmetic.
- Commit subjects (`Perf: Optimize performance…`) are not Conventional Commits as AGENTS.md
  requires — should be `perf(rows): …`.
- `cached_cells` is added to `RowSleeve::jsonSerialize()`. `RowSleeve` is not in
  `ResponseDefinitions.php` or any controller, so no OpenAPI regeneration is needed — but the full
  row payload now serializes anywhere a sleeve is dumped (e.g. logs).

</details>

Koc:
@blizzz views loading works fine for me for both cached/non-cached scenarios. Can you provide more details (error/stacktrace) or just push fix?

image


Also I've fixed extra code review comments except B2, M2, M6, M7 and some minior

lib/Db/Row2Mapper.php
Koc: added `atomic()` + `SELECT FOR UPDATE` for databases that support it

blizzz:
Looking more closely, the stack trace is strange. No such issues on main though. Would not rule out that there could be an artifiact on my dev setup.

<details><summary>Stacktrace</summary>
<p>

```json
{
  "reqId": "BHMc5EZhqoX5MrtyYIpW",
  "level": 3,
  "time": "2026-09-16T21:08:39+00:00",
  "remoteAddr": "127.0.0.1",
  "user": "master",
  "app": "tables",
  "method": "GET",
  "url": "/index.php/apps/tables/row/view/872",
  "scriptName": "/index.php",
  "message": "An internal error or exception occurred: OCA\\Tables\\Db\\Row2Mapper - findAll: Did expect one result but found none when executing: query \"SELECT * FROM `*PREFIX*tables_columns` WHERE `id` = :dcValue1\"; ",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64; rv:155.0) Gecko/20100101 Firefox/155.0",
  "version": "36.0.0.0",
  "exception": {
    "Exception": "OCA\\Tables\\Errors\\InternalError",
    "Message": "OCA\\Tables\\Db\\Row2Mapper - findAll: Did expect one result but found none when executing: query \"SELECT * FROM `*PREFIX*tables_columns` WHERE `id` = :dcValue1\"; ",
    "Code": 0,
    "Trace": [
      {
        "file": "/srv/http/nextcloud/apps-repos/tables/lib/Service/RowService.php",
        "line": 134,
        "function": "findAll",
        "class": "OCA\\Tables\\Db\\Row2Mapper",
        "type": "->",
        "args": [
          [
            -1,
            -5,
            5356,
            5360
          ],
          1465,
          null,
          null,
          [],
          [
            {
              "columnId": -5,
              "mode": "DESC"
            }
          ],
          "master"
        ]
      },
      {
        "file": "/srv/http/nextcloud/apps-repos/tables/lib/Controller/RowController.php",
        "line": 57,
        "function": "findAllByView",
        "class": "OCA\\Tables\\Service\\RowService",
        "type": "->",
        "args": [
          872,
          "master"
        ]
      },
      {
        "file": "/srv/http/nextcloud/apps-repos/tables/lib/Controller/Errors.php",
        "line": 23,
        "function": "{closure:OCA\\Tables\\Controller\\RowController::indexView():52}",
        "class": "OCA\\Tables\\Controller\\RowController",
        "type": "->",
        "args": [
          "*** sensitive parameters replaced ***"
        ]
      },
      {
        "file": "/srv/http/nextcloud/apps-repos/tables/lib/Controller/RowController.php",
        "line": 52,
        "function": "handleError",
        "class": "OCA\\Tables\\Controller\\RowController",
        "type": "->",
        "args": [
          {
            "__class__": "Closure"
          }
        ]
      },
      {
        "file": "/srv/http/nextcloud/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 198,
        "function": "indexView",
        "class": "OCA\\Tables\\Controller\\RowController",
        "type": "->",
        "args": [
          872
        ]
      },
      {
        "file": "/srv/http/nextcloud/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 86,
        "function": "executeController",
        "class": "OC\\AppFramework\\Http\\Dispatcher",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Tables\\Controller\\RowController"
          },
          "indexView"
        ]
      },
      {
        "file": "/srv/http/nextcloud/lib/private/AppFramework/App.php",
        "line": 138,
        "function": "dispatch",
        "class": "OC\\AppFramework\\Http\\Dispatcher",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Tables\\Controller\\RowController"
          },
          "indexView"
        ]
      },
      {
        "file": "/srv/http/nextcloud/lib/private/Route/Router.php",
        "line": 324,
        "function": "main",
        "class": "OC\\AppFramework\\App",
        "type": "::",
        "args": [
          "OCA\\Tables\\Controller\\RowController",
          "indexView",
          {
            "__class__": "OC\\AppFramework\\DependencyInjection\\DIContainer"
          },
          {
            "_route": "tables.row.indexview",
            "viewId": "872"
          }
        ]
      },
      {
        "file": "/srv/http/nextcloud/lib/OC.php",
        "line": 1232,
        "function": "match",
        "class": "OC\\Route\\Router",
        "type": "->",
        "args": [
          "/apps/tables/row/view/872"
        ]
      },
      {
        "file": "/srv/http/nextcloud/index.php",
        "line": 28,
        "function": "handleRequest",
        "class": "OC",
        "type": "::",
        "args": []
      },
      {
        "file": "/srv/http/nextcloud/lib/OC.php",
        "line": 1397,
        "function": "{closure:/srv/http/nextcloud/index.php:25}",
        "args": [
          "*** sensitive parameters replaced ***"
        ]
      },
      {
        "file": "/srv/http/nextcloud/index.php",
        "line": 25,
        "function": "handleRequests",
        "class": "OC",
        "type": "::",
        "args": [
          {
            "__class__": "Closure"
          }
        ]
      }
    ],
    "File": "/srv/http/nextcloud/apps-repos/tables/lib/Db/Row2Mapper.php",
    "Line": 195,
    "message": "An internal error or exception occurred: OCA\\Tables\\Db\\Row2Mapper - findAll: Did expect one result but found none when executing: query \"SELECT * FROM `*PREFIX*tables_columns` WHERE `id` = :dcValue1\"; ",
    "exception": "{\"class\":\"OCA\\Tables\\Errors\\InternalError\",\"message\":\"OCA\\Tables\\Db\\Row2Mapper - findAll: Did expect one result but found none when executing: query \\\"SELECT * FROM `*PREFIX*tables_columns` WHERE `id` = :dcValue1\\\"; \",\"code\":0,\"file\":\"/srv/http/nextcloud/apps-repos/tables/lib/Db/Row2Mapper.php:195\",\"trace\":\"#0 /srv/http/nextcloud/apps-repos/tables/lib/Service/RowService.php(134): OCA\\Tables\\Db\\Row2Mapper->findAll(Array, 1465, NULL, NULL, Array, Array, '...')\\n#1 /srv/http/nextcloud/apps-repos/tables/lib/Controller/RowController.php(57): OCA\\Tables\\Service\\RowService->findAllByView(872, '...')\\n#2 /srv/http/nextcloud/apps-repos/tables/lib/Controller/Errors.php(23): OCA\\Tables\\Controller\\RowController->{closure:OCA\\Tables\\Controller\\RowController::indexView():52}()\\n#3 /srv/http/nextcloud/apps-repos/tables/lib/Controller/RowController.php(52): OCA\\Tables\\Controller\\RowController->handleError(Object(Closure))\\n#4 /srv/http/nextcloud/lib/private/AppFramework/Http/Dispatcher.php(198): OCA\\Tables\\Controller\\RowController->indexView(872)\\n#5 /srv/http/nextcloud/lib/private/AppFramework/Http/Dispatcher.php(86): OC\\AppFramework\\Http\\Dispatcher->executeController(Object(OCA\\Tables\\Controller\\RowController), '...')\\n#6 /srv/http/nextcloud/lib/private/AppFramework/App.php(138): OC\\AppFramework\\Http\\Dispatcher->dispatch(Object(OCA\\Tables\\Controller\\RowController), '...')\\n#7 /srv/http/nextcloud/lib/private/Route/Router.php(324): OC\\AppFramework\\App::main('...', '...', Object(OC\\AppFramework\\DependencyInjection\\DIContainer), Array)\\n#8 /srv/http/nextcloud/lib/OC.php(1232): OC\\Route\\Router->match('...')\\n#9 /srv/http/nextcloud/index.php(28): OC::handleRequest()\\n#10 /srv/http/nextcloud/lib/OC.php(1397): {closure:/srv/http/nextcloud/index.php:25}()\\n#11 /srv/http/nextcloud/index.php(25): OC::handleRequests(Object(Closure))\\n#12 {main}\"}",
    "CustomMessage": "An internal error or exception occurred: OCA\\Tables\\Db\\Row2Mapper - findAll: Did expect one result but found none when executing: query \"SELECT * FROM `*PREFIX*tables_columns` WHERE `id` = :dcValue1\"; "
  }
}
```

</p>
</details>

blizzz:
P.S.: Indeed the query has an empty result set: 

```
> select * from oc_tables_columns where id = 1465;
…
(0 rows)
```

But the table loads fine, and another view, without the ID meta column, loads fine.

blizzz:
PPS.:

`select * from oc_tables_columns where id = 5356;`

results in the column data, but

`select * from oc_tables_columns where id = -5;`

results in zero results, hence the error.

Koc:
@blizzz reproduced and fixed :tada: 

image

Issue #2237, opened 2025-12-28
v2 API: Adding new row to table via request is hard to figure out


OrakMoya:
### Is your feature request related to a problem? Please describe.

The OCS API description for route `/ocs/v2.php/apps/tables/api/2/{nodeCollection}/{nodeId}/rows` simply states that new row data is `{ "data": "string" } `, going into no detail.

The v1 api for getting rows returns rows in this structure:
```json
"data": [
    {
        "columnId": int,
        "value": string
    },
    ...
]
```

This leads me to assume that the API for storing a new row expects the same structure. It doesn't, failing with this error in Nextcloud logs: `Column with id 0 is not part of table with id 7`

Visiting `lib/Controller/RowOCSController.php`, the `create()` method has a PHPDoc indicating: `$data An array containing the column identifiers and their values`. This is misleading as a PHP array is not the same as a JavaScript array, which the method isn't expecting in it's decoded JSON.

The code that extracts the column IDs and their values currently looks like this:
```php
// lib/Controller/RowOCSController.php:74
$newRowData = new RowDataInput();
foreach ($data as $key => $value) {
	$newRowData->add((int)$key, $value);
}
```

This route actually expects data as a PHP array that has keys as the column IDs and the value at key with the value of the array. This means that the JSON representing the data "array" is not an array but a JavaScript object:
```json
"data": {
    "columnId": "value"
}
```

### Describe the solution you'd like

Expand the docs to specify the correct data structure:

```
...
data: A PHP array, or a JSON object, containing column IDs as keys and column values as value at key.
...

Request body
Example value
"data": {
    "columnId1": "value",
    "columnId2": "value"
}
```

Also, return the error in the error message of the response. Currently, it fails with `An unexpected error occurred. More details can be found in the logs. Please reach out to your administration.` It'd be nicer to return a `Column with id .. not found`

### Describe alternatives you've considered

Rewrite the method to accept data in this format:
```json
"data": [
    {
        "columnId": int,
        "value": string
    },
    ...
]
```
e.g.:
```php
$newRowData = new RowDataInput();
foreach ($data as $columnData) {
	$newRowData->add((int)$columnData["columnId"], $columnData["value"]);
}
```

or, even better, to avoid numeric IDs (maybe on another route):
```json
"data": [
    {
        "columnName": string,
        "value": string
    },
    ...
]
```

Both variants above also play nicer with strict typing.

[...]

Issue #1840, opened 2025-06-10
API structure improvement: expose columns through a technical name/slug/alias


ericfischereu:
### Is your feature request related to a problem? Please describe.

When developing workflows with Nextcloud Flow and Tables, there's always the need for pretty complex filters since the values of a row are returned as array.

### Describe the solution you'd like

Values should be returned as JSON fields. It might also be beneficial to allow the user to add a technical name to a column, which is then the field name in the json.

### Describe alternatives you've considered

Using the filters, but that results in pretty complex data extraction (working with column IDs and filters)

blizzz:
Do you use the data from an event or from an API endpoint? Which?

ericfischereu:
Mainly from the webhook event. But it should also be possible to get the data with `/index.php/apps/tables/api/1/tables/{tableId}/rows`

Right, I think the row data is not being returned when calling this endpoint, but it should be from my point of view.

nimishavijay:
Talked about this in the design team call @samin-z and while a technical name sounds good as a short term fix, there could be more efficient solutions UX-wise:

1. Move technical items into "Advanced" section so the creation process doesn’t get more daunting. These advanced settings should be: Technical name, Column width. (And individual type fields like "Maximum text length" etc for other types would be nice to catch as well)
2. Automatically prefill the "Technical name" field with a slug of the name and the ID which can be copied. This can change based on renames though and is volatile, so probably shouldn’t be done at all.
3. Don’t expose a concept of "Technical name" at all but directly reference via IDs under the hood. Like how you don’t reference files by ID but get a file picker. ("Copy column URL" could still be available read-only/for copying, like in Airtable)

In the short term we can go with solution 1, long-term should be solution 3 :) What do you think?

blizzz:
> Don’t expose a concept of "Technical name" at all but directly reference via IDs under the hood. Like how you don’t reference files by ID but get a file picker. ("Copy column URL" could still be available read-only/for copying, like in Airtable)

This is exactly what this concept has to solve. For admin users who want to connect Tables to a windmill flow, numeric IDs are bad experience, they tell you nothing about the reference. Hence the approach to have a stable label that can be recognized and unterstood in their scripts.

samin-z:
@nimishavijay 
[...]

PR #2246, opened 2026-01-08
Feature/2237 add column name

samin-z:
This PR is a fix for [#1840](https://github.com/nextcloud/tables/issues/1840) and #2237.
# Description:
- User asked for more info of the column name to be also returned in response when user add new row so which each cell belongs to which column.
- Return more clear error message when user calls for non-existing row.

# Issues fixed:
- add new value called columnName to response when saving new row.
- return clear error message when user add none-existing row id.

### DIFF
--- a/lib/Controller/RowOCSController.php
+++ b/lib/Controller/RowOCSController.php
@@ -77,7 +77,7 @@ class RowOCSController extends AOCSController {
 		}
 
 		try {
-			return new DataResponse($this->rowService->create($tableId, $viewId, $newRowData)->jsonSerialize());
+			return new DataResponse($this->rowService->create($tableId, $viewId, $newRowData)->toResponseArray());
 		} catch (BadRequestError $e) {
 			return $this->handleBadRequestError($e);
 		} catch (NotFoundError $e) {
--- a/lib/Controller/RowOCSController.php
+++ b/lib/Controller/RowOCSController.php
@@ -77,7 +77,7 @@ class RowOCSController extends AOCSController {
 		}
 
 		try {
-			return new DataResponse($this->rowService->create($tableId, $viewId, $newRowData)->toResponseArray());
+			return new DataResponse($this->rowService->create($tableId, $viewId, $newRowData)->jsonSerialize());
 		} catch (BadRequestError $e) {
 			return $this->handleBadRequestError($e);
 		} catch (NotFoundError $e) {
--- a/lib/Db/Row2.php
+++ b/lib/Db/Row2.php
@@ -145,7 +145,15 @@ class Row2 implements JsonSerializable {
 	}
 
 	/**
-	 * @psalm-return TablesRow
+	 * @return array{
+	 *   id: int|null,
+	 *   tableId: int|null,
+	 *   createdBy: string|null,
+	 *   createdAt: string|null,
+	 *   lastEditBy: string|null,
+	 *   lastEditAt: string|null,
+	 *   data: list<array<array-key, mixed>>
+	 * }
 	 */
 	public function jsonSerialize(): array {
 		$data = [];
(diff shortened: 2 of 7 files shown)


lib/Controller/RowOCSController.php
blizzz, line 48: The added field should be reflected in the response definition of TablesRow, cf. https://github.com/nextcloud/tables/blob/main/lib/ResponseDefinitions.php#L76-L84

lib/Db/Row2.php
```diff
@@ -148,6 +159,31 @@ public function jsonSerialize(): array {
 		];
 	}
 
+	/**
+	 * return merged response-only metadata into the data cells.
+	 */
```
blizzz: The method names says this already, no need to duplicate :) 

What we want is however a `@return` annotation with the array shape. Or `@psalm-return` as done at `jsonSerialize()`, which itself would need to be adjusted afterwards.

Did you double check, cannot we only extend `jsonSerialize()` instead of adding a similar method?

lib/Db/Row2.php
```diff
@@ -23,6 +23,7 @@ class Row2 implements JsonSerializable {
 	private ?string $lastEditBy = null;
 	private ?string $lastEditAt = null;
 	private ?array $data = [];
+	private array $cellMeta = [];
```
blizzz: Should need to be added to a `VIRTUAL_PROPERTIES` const, otherwise the Mapper will complain. Have a look at https://github.com/nextcloud/tables/blob/main/lib/Db/Column.php#L161C2-L161C114 for comparison.

Also, the name is too broad and not specific (same problem with `data` already…), but right now I am also missing a better suggestion.

lib/Service/RowService.php
```diff
@@ -208,6 +208,28 @@ public function create(?int $tableId, ?int $viewId, RowDataInput|array $data): R
 			throw new InternalError('Cannot create row without table or view in context');
 		}
 
+		$fullRowData = [];
+		$columnNames = [];
+
+		foreach ($columns as $c) {
```
blizzz:
```suggestion
		foreach ($columns as $column) {
```

Please don't save bytes where it is not necessary :)

lib/Service/RowService.php
```diff
@@ -208,6 +208,28 @@ public function create(?int $tableId, ?int $viewId, RowDataInput|array $data): R
...
+			$columnNames[$c->getId()] = $c->getTitle();
+		}
+
+		$rows = $data instanceof RowDataInput ? iterator_to_array($data) : $data;
+
+		foreach ($rows as $r) {
```
blizzz: same

lib/Service/RowService.php
```diff
@@ -220,6 +242,12 @@ public function create(?int $tableId, ?int $viewId, RowDataInput|array $data): R
 		$row2->setData($data);
 		try {
 			$insertedRow = $this->row2Mapper->insert($row2);
+			// attached columnname to returned row for the response only
+			foreach ($fullRowData as $meta) {
+				if (isset($meta['columnId']) && array_key_exists('columnName', $meta)) {
+					$insertedRow->addCellMeta((int)$meta['columnId'], ['columnName' => $meta['columnName']]);
```
blizzz: can we generate it in Row2 instead? `data` is populated there. Having all logic in Row2 would capsulate it there and not leak logic to other classes around.

blizzz (changes requested):
Thanks for the effort @samin-z! I have left a few remarks, what do you think?

lib/Db/Row2.php
samin-z: ill rename it cellMetadata as the array holds metadata for each cell.

lib/Db/Row2.php
samin-z: i was a bit hesitant on using 'jsonSerialize' as i thought i might interfere in another place, but did more testing and it's fine.

samin-z:
this ticket can be closed after merging this ticket (https://github.com/nextcloud/tables/issues/1840)

enjeck:
No longer needed, iiuc

PR #2248, opened 2026-01-09
feat: add relation column type

### LINKED ISSUE #172: Link Tables by Key
It would be amazing if we coul activate a Key-Field in a table and re-use that key field in a second table.

Here is an example use-case:

- Table A: List of Movies
-- Fields: Name, Year, Category, etc. 
-- Key Field: Name

- Table B: Comments
-- Fields: Name (From Table A), Rating, Comment

That would be amazing because, than you could add a Link in Table A that says "show comments" that links to Table B that is Pre-Filtered (https://github.com/nextcloud/tables/issues/171) for the Movie that you want to see the comments for.

Long term goal would be to pre-aggregate the informations from Table B in Table A. So instead of having a column with a link that says "show comments" we could have a column that says "average rating" or "number of comments".

Koc:
This PR introduces a basic relation type. It allows to connect few tables together. We will store row id as a column value and display selected column as a label for this values. Closes #172.

## :heavy_check_mark: Implemented
<details><summary>:mag: Column create/edit</summary>
<p>
image
</p>
</details> 

<details><summary>:mag: Row create/edit via popup/inline</summary>
<p>
image
</p>
</details> 

<details><summary>:mag: Row display in table/view + handle relation row removal</summary>
<p>
image

image
</p>
</details> 

<details><summary>:mag: Filter for view</summary>
<p>
image 
</p>
</details> 

<details><summary>:mag: Filter for rows</summary>
<p>
image

image

</p>
</details> 

<details><summary>:mag: Import/Export</summary>
<p>

image

image

</p>
</details> 

## :construction: Pending (help welcomed):
- [x] Behat tests
- [ ] Cypress tests?

## :negative_squared_cross_mark: What is out of scope:
- Lookup columns to add an extra columns from the related table to the current one. It will be handled separately in upcoming [PR](https://github.com/nextcloud/tables/pull/2271).


src/shared/components/ncTable/partials/columnTypePartials/forms/RelationForm.vue
```diff
@@ -0,0 +1,171 @@
...
+			this.loadingColumns = true
+			try {
+				const dataStore = useDataStore()
+				const columns = await dataStore.getColumnsFromBE({
+					tableId: this.customSettings.relationType === 'table' ? this.customSettings.targetId : null,
+					viewId: this.customSettings.relationType === 'view' ? this.customSettings.targetId : null,
+				})
+				this.availableLabelColumns = columns
+					.filter(column =>
+						column instanceof NumberColumn || column instanceof TextLineColumn,
```
enjeck, line 140: why restrict to just these 2 column types?

enjeck: it seems like the relation should easily generalise to all column types

lib/Service/RelationService.php
```diff
@@ -0,0 +1,214 @@
...
+			if ($isView) {
+				$view = $this->viewMapper->find($targetId);
+				$rows = $this->row2Mapper->findAll(
+					[$targetColumn->getId()],
+					$view->getTableId(),
+					null,
+					null,
+					$view->getFilterArray(),
+					$view->getSortArray(),
+					$this->userId
```
enjeck, line 176: Do we ever consider permissions such that the user must have access to the table/view before they can link to it?

lib/Db/RowCellRelationMapper.php
```diff
@@ -0,0 +1,40 @@
...
+	/**
+	 * @inheritDoc
+	 */
+	public function getDbParamType() {
+		return IQueryBuilder::PARAM_INT;
+	}
+
+	public function formatRowData(Column $column, array $row) {
+		$value = $row['value'];
+		return (int)$value;
```
enjeck: int(null) is 0. Is 0 a valid id? Could this cause troube?

src/shared/components/ncTable/partials/TableCellRelation.vue
```diff
@@ -0,0 +1,188 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<div class="cell-relation">
+		<div v-if="!isEditing" class="non-edit-mode" @click="handleStartEditing">
+			<div v-if="isDeleted">
+				<span class="deleted">{{ value }}</span>
+				<span class="cursor-help" :title="t('tables', 'This relation is not exists anymore.')">&nbsp;⚠️</span>
```
enjeck:
```suggestion
				<span class="cursor-help" :title="t('tables', 'This relation does not exist anymore.')">&nbsp;⚠️</span>
```

enjeck (commented):
Thanks, it works quite well already. A few comments for now. I have only tested the creation of columns and rows so far

src/shared/components/ncTable/partials/columnTypePartials/forms/RelationForm.vue
Koc: In theory we can, but I'm not sure that this is good idea. This is the column which we will display in the relation dropdown. So, this column should be unique in order to properly select correct relation row (otherwise you will select wrong rows).

What sense to display select/multi-select/rating here? I doubt that those type of columns will be unique.  

image
image

lib/Service/RelationService.php
Koc: No, and this is a "feature by design". The main idea is that admin creates tables for dictionaries, e.g. `users, departments, offices, etc`, then creates table that references to this dictionaries, e.g. `vacations`. And he can share only `vacations` to the end users.

Otherwise he should to share all dictionary tables as well. And users will have tons of irrelevant tables in their shares. Or admin can forget to share dictionary tables and users will suffer because table with relations not displays all columns.

We can't share table partially per column right now. Let's preserve same approach for new column types as well.

lib/Db/RowCellRelationMapper.php
Koc: heh, the funny thing that this method not used at all. I've just removed it

Koc:
pipeline failures looks like are unrelated. Please re-review @enjeck @blizzz

lib/Service/RelationService.php
enjeck: We have to be careful not to leak data. At the moment, suppose Bob shares a table X witih Alice. Then Alice creates a table Y with a column that links to table X. Then Bob later decides to revoke the share from Alice. But Alice is still able to continue reading the linked relation column, despite not having the permission for that. Plus, Alice gets a 403 when they try to edit that column (which is to be expected since they no longer have permissions, but maybe a better error message than `Could not load columns. Request not allowed` helps?)

src/shared/components/ncTable/mixins/filter.js
```diff
@@ -49,19 +49,20 @@ export const FilterIds = {
 	IsLowerThan: 'is-lower-than',
 	IsLowerThanOrEqual: 'is-lower-than-or-equal',
 	IsEmpty: 'is-empty',
+	IsNotEmpty: 'is-not-empty',
```
enjeck: Creating a view with this filter doesn't work, as the backend filter needs to modified too

src/shared/components/ncTable/partials/columnTypePartials/forms/RelationForm.vue
enjeck: Ok, thanks.

enjeck (commented):
would be nice to have tests to avoid regressions in the future

lib/Service/RelationService.php
Koc: I got what you mean, but we open only single column that selected in "Value selection label".

In our organization we have hundreds of dictionary-tables. Imagine what mess can can have our users that will manage tables that rely on this dictionary-tables.

I can add some `NcNoteCard` with a notice that field will be exposed. WDYT?

Koc:
@Aveyron-RetD here we go, without tooltip

image

src/shared/components/ncTable/mixins/filter.js
Koc: Let's handle it separately https://github.com/nextcloud/tables/pull/2572

lib/Service/RelationService.php
enjeck: > I can add some NcNoteCard with a notice that field will be exposed. WDYT?

okay, for the MVP so we can merge this now, adding this is okay. Thanks!

We've discussed internally some ideas for what the follow-up for this would look like:
- we require that a column that is being referenced is accessible by a table manager.
- when a table ore or view, that contains such a column, is being unshared with someone, we check whether this this column is included in a reference. And then we re-eavluate the situation: check whether the owner or any other manager still has access to that column. If not, disable the relation and show an error.
- check should also happen with ownership transfer and demoting table managers.

Koc: image

enjeck: Sorry, but can we remove the notecard? I just realized that it does not address the Alice/Bob revoke case because the referenced-table owner never sees it. So it's basically useless. We'll just handle things properly in the follow up
Sorry for the confusion

tests/integration/features/APIv2.feature
```diff
@@ -121,6 +121,21 @@ Feature: APIv2
...
       | ug column renamed |
 
+  @api2relation
+  Scenario: Create relation columns
+    Given table "Employees" with emoji "👥" exists for user "participant1-v2" as "employees" via v2
+    And table "Vacations" with emoji "🏖️" exists for user "participant1-v2" as "vacations" via v2
+    Then column from main type "text" for node type "table" and node name "employees" exists with name "employee-name" and following properties via v2
+      | title   | Employee Name |
+      | subtype | line          |
+    Then column from main type "relation" for node type "table" and node name "vacations" exists with name "employee-relation" and following properties via v2
```
enjeck: We don't have a `/api/2/columns/relation` route though, so i don't think this works 🤔

lib/Service/RelationService.php
Koc: Removed

Koc:
hey @enjeck @blizzz!

I've added Behat tests, please check. Is that enough to merge PR?

enjeck (approved):
Thank you so much!

xbluemonkx:
H,
I have seen that this is part of the new release. - May I ask what exactly this offers now?

Is it possible to define "link" columns (columns that are avaible in to tables so if someone adds and entry in table A it is possible to use it in table B as a link?

Or is it more like: if i manually enter the same value in table A and table B it can be used as a link?

I am really curious about what / how this actually works from a user perspective

Koc:
@xbluemonkx mate, sorry, but I didn't got what you mean. You can check screenshots from the PR description or even checkout latest version of the module, install it locally and test by yourself (and answer to your question :wink:)

Aveyron-RetD:
@xbluemonkx  i believe this is this feature #2271 ,that you want which will bring you lookup columns. 
In this feature, there is only text field which works.

What you can do for the moment with this feature is for example an ordering table from an inventory table. You can have a table with all your references (A) and another table where you can order (B) when you create a relation column in table (B) you reference to a text field column in table A for example (Inventory name).  

So by editing table A, you create new references, and they will appear as selection choices in table B.  
The only downside they could be restricting your searches because the text field could be not enough for the user to recognize the inventory and this is where #2271 will come handy. It will fetch other results in table A on the same row and prefill cells depending on the text field chosen. 

Am i right @Koc  ?

Koc:
yeah, you are pretty good explained how it works, thanx!

PR #2271, opened 2026-01-19
feat: add relation lookup column type

Koc:
This is continuation of #2248 that allows to display more column from relation table.
There are 2 tables: `employees` and `vacations`. We can add `vacation.requester` **relation** column that referenced to the row id of the `employee` and display `employee.name`. But what if we need to display more column from the related table? e.g. `employee.department`. That's where **relation lookup** can help: select already existent relation and choose what extra column should be displayed.

## :framed_picture: Screenshots

image <br />

image

## :spiral_notepad: TODO
- [x] Merge https://github.com/nextcloud/tables/pull/2763 first

### DIFF
--- a/src/modules/modals/CreateRow.vue
+++ b/src/modules/modals/CreateRow.vue
@@ -9,15 +9,15 @@
 		data-cy="createRowModal"
 		@closing="actionCancel">
 		<div class="modal__content" @keydown="onKeydown">
-			<div v-for="column in nonMetaColumns" :key="column.id" :data-cy="column.title">
+			<div v-for="column in nonMetaColumns" :key="column.getValueColumnId()" :data-cy="column.title">
 				<ColumnFormComponent
-					v-model:value="row[column.id]"
+					v-model:value="row[column.getValueColumnId()]"
 					:column="column" />
-				<NcNoteCard v-if="isMandatory(column) && !isValueValidForColumn(row[column.id], column)"
+				<NcNoteCard v-if="isMandatory(column) && !isValueValidForColumn(row[column.getValueColumnId()], column)"
 					type="error">
 					{{ t('tables', '"{columnTitle}" should not be empty', { columnTitle: column.title }) }}
 				</NcNoteCard>
-				<NcNoteCard v-if="row[column.id] && column.type === 'text-link' && !isValidUrlProtocol(row[column.id])"
+				<NcNoteCard v-if="row[column.getValueColumnId()] && column.type === 'text-link' && !isValidUrlProtocol(row[column.getValueColumnId()])"
 					type="error">
 					{{ t('tables', 'Invalid protocol. Allowed: {allowed}', {allowed: allowedProtocols.join(', ')}) }}
 				</NcNoteCard>
@@ -104,7 +104,7 @@ export default {
 			return this.checkMandatoryFields(this.row)
 		},
 		hasInvalidUrlProtocol() {
-			return this.nonMetaColumns.some(col => col.type === 'text-link' && !this.isValidUrlProtocol(this.row[col.id]))
+			return this.nonMetaColumns.some(col => col.type === 'text-link' && !this.isValidUrlProtocol(this.row[col.getValueColumnId()]))
 		},
 		dialogTitle() {
 			return this.isFormMode ? t('tables', 'Fill form') : t('tables', 'Create row')
--- a/src/shared/components/ncTable/mixins/exportTableMixin.js
+++ b/src/shared/components/ncTable/mixins/exportTableMixin.js
@@ -5,9 +5,11 @@
 import moment from '@nextcloud/moment'
 import Papa from 'papaparse'
 import generalHelper from '../../../mixins/generalHelper.js'
+import { ColumnTypes } from './columnHandler.js'
 import {
 	TYPE_META_ID, TYPE_META_CREATED_BY, TYPE_META_CREATED_AT, TYPE_META_UPDATED_BY, TYPE_META_UPDATED_AT,
 } from '../../../../shared/constants.ts'
+import { useDataStore } from '../../../../store/data.js'
 
 export default {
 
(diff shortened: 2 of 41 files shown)


enjeck (commented):
Even with the screenshots, I'm not sure what we're adding and what the usecase is?

Koc:
@enjeck I've updated description:

> There are 2 tables: `employees` and `vacations`. We can add `vacation.requester` **relation** column that referenced to `employee.name`. But what if we need to display more column from the related table? e.g. `employee.department`. That's where **relation lookup** can help: select already existent relation and choose what extra column should be displayed.

src/modules/modals/CreateRow.vue
```diff
@@ -12,12 +12,12 @@
 			<div v-for="column in nonMetaColumns" :key="column.id" :data-cy="column.title">
 				<ColumnFormComponent
 					:column="column"
-					:value.sync="row[column.id]" />
-				<NcNoteCard v-if="isMandatory(column) && !isValueValidForColumn(row[column.id], column)"
+					:value.sync="row[getValueColumnId(column)]" />
```
Koc: let's just call `column.getValueColumnId()` directly

src/shared/components/ncTable/mixins/exportTableMixin.js
```diff
@@ -4,29 +4,71 @@
...
 					// if a normal column
 					if (column.id >= 0) {
-						const set = row.data ? row.data.find(d => d.columnId === column.id) || '' : null
-						rowData[column.title] = set ? column.getValueString(set) : ''
+						if (column.type === ColumnTypes.RelationLookup) {
+							rowData[column.title] = this.getRelationLookupValue(column, row, dataStore)
+						} else {
+							const set = row.data ? row.data.find(d => d.columnId === column.id) || '' : null
+							rowData[column.title] = set ? column.getValueString(set) : ''
+						}
```
Koc, line 72: try to omit if/else and move functionality to the column class

Koc:
added Behat tests

enjeck:
> That's where relation lookup can help: select already existent relation and choose what extra column should be displayed.

Having a relation column, and a lookup column is extremely confusing. Even with this, I had to think extra hard to understand how they're different. If I'm struggling to understand as the dev, I doubts users will easily get it (or maybe it's just me, lol). 

 I would need others to check if they understand this just from the description and images. Maybe @nextcloud/designers need to weigh in

Koc:
well, that's pretty similar to Excel's [`VLOOKUP`](https://support.microsoft.com/en-us/excel/functions/vlookup-function) function. **But I'm open to changes here**, let's wait for our designers

src/shared/components/ncTable/partials/TableCell.vue
```diff
@@ -0,0 +1,83 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<component :is="cellComponent"
```
Koc, line 6: replaced with a simple function

kra-mo:
I certainly think the names should be changed at least. @marcoambrosini, ideas?

Let's also just say that Excel's ancient statistics-oriented DSL is maybe not the best language to draw inspiration from when it comes to design ;)

marcoambrosini:
the "Lookup" term is used for this in both Baserow and Airtable. For the relational ones they use the wording "Link to table". I would do the same  

Maybe this can help: https://baserow.io/user-docs/lookup-field

src/shared/components/ncTable/partials/columnTypePartials/forms/RelationLookupForm.vue
```diff
@@ -0,0 +1,158 @@
...
+				{{ t('tables', 'Relation Column') }}
+			</div>
+			<div class="fix-col-4">
+				<NcSelect v-model="customSettings.relationColumnId"
+					:options="relationColumns"
+					:reduce="(option) => option.id"
+					:loading="loadingRelationColumns"
+					:aria-label-combobox="t('tables', 'Select relation column')"
+					required
+					@input="onRelationColumnChange" />
```
samin-z, line 18: after picking a relation column, target cplumn stays empty 'No results'. this select use @input, but NcSelect needs @update:model-value, with that change the target columns load correctly.

src/shared/components/ncTable/partials/columnTypePartials/forms/RelationLookupForm.vue
```diff
@@ -0,0 +1,158 @@
...
+				{{ t('tables', 'Target Column') }}
+			</div>
+			<div class="fix-col-4">
+				<NcSelect v-model="customSettings.targetColumnId"
+					:options="targetColumns"
+					:reduce="(option) => option.id"
+					:loading="loadingTargetColumns"
+					:aria-label-combobox="t('tables', 'Select target column')"
+					required
+					@input="onTargetColumnChange" />
```
samin-z, line 33: same as previous comment.

src/shared/components/ncTable/partials/TableCell.vue
```diff
@@ -0,0 +1,83 @@
...
+	[ColumnTypes.TextLink]: TableCellLink,
+	[ColumnTypes.TextRich]: TableCellTextRich,
+	[ColumnTypes.Number]: TableCellNumber,
+	[ColumnTypes.NumberStars]: TableCellStars,
+	[ColumnTypes.NumberProgress]: TableCellProgress,
+	[ColumnTypes.Selection]: TableCellSelection,
+	[ColumnTypes.SelectionMulti]: TableCellMultiSelection,
+	[ColumnTypes.SelectionCheck]: TableCellYesNo,
+	[ColumnTypes.Relation]: TableCellRelation,
+	[ColumnTypes.RelationLookup]: TableCellRelationLookup,
```
samin-z, line 43: Lookup cells still open as an empty text editor. RelationLookup is register here, but TableRow uses getTableCellComponent from TableCell.js, which doesnt include it and falls back to TableCellHtml. you should add the same mapping in TableCell.js.

samin-z (changes requested):

PR #2279, opened 2026-01-23
fix: Merge views instead of resetting to avoid race condition

### LINKED ISSUE #2180: Permission problem. User should see the view but he doesn't
### Steps to reproduce

I have some table and view with some (same) permissions:

Image

Image

I also have a user in groupe with this permissions:

Image

User can see this table and view:

Image

But if i add this user to another group (It doesn't matter which one) hi can't see view anymore:

Image

Image

But the fun part is: if i remove user from this additional group and add it again sometimes i can see view. Bo the other time i can't. Sometimes even if user is only in group with permissions hi can't see the view. But the other time he can see...

Now i can't see any pattern to be honest... It looks quite random to me at this moment... Sometimes i see the view and sometimes i don't. Without any clear reason.

This only applies to views, NOT to tables.

### Expected behavior

The user should see the view if he is in the group with permissions.

### Actual behavior

The user can't see the view if he is in the group with permissions.

### Tables app version

1.0.1

### Browser

Chrome 142.0.7444.176

### Client operating system

Windows 11

### Operating system

Ubuntu

### Web server

Apache

### PHP engine version

PHP 8.3

### Database

MariaDB

### Additional info

Ubuntu: 24.04.2 LTS
Nextcloud: 31.0.9

Kaijudo:
Fixes #2180

### DIFF
--- a/src/store/store.js
+++ b/src/store/store.js
@@ -153,9 +153,14 @@ export const useTablesStore = defineStore('store', {
 			try {
 				const res = await axios.get(generateUrl('/apps/tables/table'))
 				this.setTables(res.data)
-				this.views = []
 				res.data.forEach(table => {
-					if (table.views) this.views = this.views.concat(table.views)
+					if (table.views) {
+						table.views.forEach(view => {
+							if (this.views.filter(v => v.id === view.id).length === 0) {
+								this.views.push(view)
+							}
+						})
+					}
 				})
 			} catch (e) {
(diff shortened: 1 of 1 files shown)


blizzz:
@Kaijudo thanks, this makes sense. Could you please sign-off your commit so the DCO checks passes?

blizzz (approved):

Kaijudo:
@blizzz I updated my commit to include the DCO

PR #2284, opened 2026-01-25
refactor: Refactor business column classes

Koc:
1. instanciate a column business in a single place `ColumnsHelper::getColumnBusiness()` instead of duplication:
   - `ImportService::parseValueByColumnType()`
   - `RowService::getColumnBusiness()`
2. memoize column business instance per column type in a factory and per column id in a import to speedup performance a bit
3. `SuperBusiness` now implements `IColumnTypeBusiness`
4. make `$column` non-nullable in a column business classes and remove redundant checks. I can't find any use-case when it can be nullable

### DIFF
--- a/lib/Helper/ColumnsHelper.php
+++ b/lib/Helper/ColumnsHelper.php
@@ -9,6 +9,8 @@ namespace OCA\Tables\Helper;
 
 use OCA\Tables\Constants\UsergroupType;
 use OCA\Tables\Db\Column;
+use OCA\Tables\Service\ColumnTypes\IColumnTypeBusiness;
+use OCP\Server;
 
 class ColumnsHelper {
 
@@ -19,6 +21,11 @@ class ColumnsHelper {
 		Column::TYPE_SELECTION,
 		Column::TYPE_USERGROUP,
 	];
+
+	/**
+	 * @var array<string, IColumnTypeBusiness>
+	 */
+	private array $columnBusinesses = [];
 
 	public function __construct(
 		private UserHelper $userHelper,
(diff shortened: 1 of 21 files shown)


lib/Helper/ColumnsHelper.php
```diff
@@ -79,4 +86,18 @@ public function resolveSearchValue(string $placeholder, string $userId, ?Column
 			default: return $placeholder;
 		}
 	}
+
+	public function getColumnBusiness(Column $column): IColumnTypeBusiness {
```
blizzz:
```suggestion
	public function getColumnBusinessObject(Column $column): IColumnTypeBusiness {
```

nitpick, not blocking

lib/Service/ImportService.php
```diff
@@ -76,6 +79,7 @@ public function __construct(
 		TableService $tableService,
 		ViewService $viewService,
 		IUserManager $userManager,
+		private ColumnsHelper $columnsHelper,
```
blizzz:
```suggestion
		private readonly ColumnsHelper $columnsHelper,
```

lib/Service/RowService.php
```diff
@@ -231,6 +230,7 @@ public function create(?int $tableId, ?int $viewId, RowDataInput|array $data): R
 		$data = $this->cleanupAndValidateData($data, $columns, $tableId, $viewId);
 		$data = $this->enhanceWithViewDefaults($view, $data);
 
+		$tableId = $tableId ?? $view->getTableId();
```
blizzz: this is already done in line 227, we should not need to repeat it here.

lib/Service/RowService.php
```diff
@@ -468,7 +468,7 @@ private function isValueValidForMandatoryColumn($value, Column $column, IColumnT
 			$defaultValue = $column->getSelectionDefault();
 			return $defaultValue !== null && $defaultValue !== '' && $defaultValue !== '[]';
 		}
-		return $value !== [];
+		return $value !== null && $value !== '' && $value !== [];
```
blizzz: the check for null and emptry string happens in line 452, it does not need to be repeated again.

PR #2290, opened 2026-01-30
fix(test): Add cypress exception

enjeck:
Current tests are failing due to the exception `Cannot read properties of undefined (reading 'from')`


enjeck:
on stable branches, we hit an unsupported browser screen, so all tests fail:

Test column date and time -- Insert and test rows - default now (failed)

enjeck:
In [fix(test): Handle unsupported browser screen](https://github.com/nextcloud/tables/pull/2290/commits/5451aef1359f95090c93975cd17478a02ada89f5), made it so that we just click the continue button to proceed

PR #2291, opened 2026-01-30
chore: Update max server version

enjeck:


enjeck:
Just noticed @blizzz
[fix: getId is supposed to return string](https://github.com/nextcloud/tables/pull/2291/commits/d1fa49958c14e9bb58ac6bcaee76fa844b3399f4) and saw I also fixed at https://github.com/nextcloud/tables/pull/2290 😅

Issue #1424, opened 2024-10-17
Please add Nextcloud tables to the user_migration export


Markoise:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

**Which Nextcloud Version are you currently using:** (see administration page)
Nextcloud 30.0.0

**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
It would be nice if you could also export the tables and data from Nextcloud Tables to restore them when moving.

**Describe alternatives you've considered**

**Additional context**

PR #2292, opened 2026-01-30
#1424: Add Tables to user_migration

samin-z:
Add ability to export/import user tables.

### DIFF
--- a/lib/Db/RowCellBulkFetchTrait.php
+++ b/lib/Db/RowCellBulkFetchTrait.php
@@ -1,0 +1,40 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Tables\Db;
+
+use OCP\DB\QueryBuilder\IQueryBuilder;
+
+trait RowCellBulkFetchTrait {
+	/**
+	 * @param int[] $rowIds
+	 * @param int[] $columnIds
+	 *
+	 * @return array
+	 */
+	public function findAllByRowIdsAndColumnIds(array $rowIds, array $columnIds): array {
+		if (empty($rowIds) || empty($columnIds)) {
+			return [];
+		}
+
+		$chunkSize = 500;
+		$results = [];
+		foreach (array_chunk($rowIds, $chunkSize) as $rowIdChunk) {
+			$qb = $this->db->getQueryBuilder();
+			$qb->select('*')
+				->from($this->table)
+				->where($qb->expr()->in('row_id', $qb->createNamedParameter($rowIdChunk, IQueryBuilder::PARAM_INT_ARRAY)))
+				->andWhere($qb->expr()->in('column_id', $qb->createNamedParameter($columnIds, IQueryBuilder::PARAM_INT_ARRAY)));
+
+			$chunkResults = $this->findEntities($qb);
+			if (!empty($chunkResults)) {
+				$results = array_merge($results, $chunkResults);
+			}
+		}
+		return $results;
+	}
+}
--- a/lib/Db/ShareMapper.php
+++ b/lib/Db/ShareMapper.php
@@ -199,4 +199,36 @@ class ShareMapper extends QBMapper {
 			->andWhere($qb->expr()->eq('node_type', $qb->createNamedParameter($nodeType, IQueryBuilder::PARAM_STR)))
 			->executeStatement();
 	}
+
+	/**
+	 * @param int[] $tableIds
+	 * @param int[] $contextIds
+	 *
+	 * @return Share[]
+	 */
+	public function findAllSharesForTablesAndContexts(array $tableIds, array $contextIds): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->table);
+
+		$orX = $qb->expr()->orX();
+		if (!empty($tableIds)) {
+			$orX->add(
+				$qb->expr()->andX(
+					$qb->expr()->eq('node_type', $qb->createNamedParameter('table', IQueryBuilder::PARAM_STR)),
+					$qb->expr()->in('node_id', $qb->createNamedParameter($tableIds, IQueryBuilder::PARAM_INT_ARRAY))
+				)
+			);
+		}
+		if (!empty($contextIds)) {
+			$orX->add(
+				$qb->expr()->andX(
+					$qb->expr()->eq('node_type', $qb->createNamedParameter('context', IQueryBuilder::PARAM_STR)),
+					$qb->expr()->in('node_id', $qb->createNamedParameter($contextIds, IQueryBuilder::PARAM_INT_ARRAY))
+				)
+			);
+		}
+		$qb->where($orX);
+		return $this->findEntities($qb);
+	}
 }
--- a/lib/Service/ColumnService.php
+++ b/lib/Service/ColumnService.php
@@ -11,6 +11,7 @@ use DateTime;
 use Exception;
 use OCA\Tables\Db\Column;
 use OCA\Tables\Db\ColumnMapper;
+use OCA\Tables\Db\Table;
 use OCA\Tables\Db\TableMapper;
 use OCA\Tables\Db\View;
 use OCA\Tables\Dto\Column as ColumnDto;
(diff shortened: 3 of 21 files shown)


lib/Service/ColumnService.php
```diff
@@ -558,4 +559,55 @@ private function enhanceColumns(?array $columns, ?View $view = null): array {
...
+	 * 
+	 * @return Column
+	 * 
+	 * @throws InternalError
+	 */
+	public function importColumn(Table $table, array $column): Column
+	{
+		$item = new Column();
+		$item->setTableId($table->getId());
+		$item->setTitle($column['title']);
```
juliusknorr, line 574: I'm not 100% sure with the abstraction layers in the tables app but you could probably shorten this by using https://github.com/nextcloud/server/blob/master/lib/public/AppFramework/Db/Entity.php#L51-L62

```
$item = Column::fromRow($column);
```

Please verify though, setTableId may need separate handling still.

lib/Service/ContextService.php
```diff
@@ -508,4 +509,79 @@ protected function insertNodesFromArray(Context $context, array $nodes): void {
...
+
+		if (!$hasMatchingNode) {
+			return null;
+		}
+
+        $newContext = new Context();
+        $newContext->setName($context['name'] ?? '');
+        $newContext->setIcon($context['iconName'] ?? '');
+        $newContext->setDescription($context['description'] ?? '');
+        $newContext->setOwnerId($context['owner'] ?? '');
```
juliusknorr: If we import, do we set the old owner id? Or would be be needed to change this to the current user id of the importer instead? Same probably applies to other importers

lib/Service/ColumnService.php
samin-z: @juliusknorr will do.

lib/Service/ContextService.php
samin-z: @juliusknorr this was actually one of the final fixes im making, im going to actually check if the user exist and if not then use current logged in user, what do you think about that? or should we go straight with the current user?

juliusknorr: We should always use the user that is importing as the new owner (as I assume only your own tables/context are exported). The current logged in user may not work as I assume it is importing in a background job?

But I would guess there is a way to get the "target" user as part of the import process of the user_migration app

lib/TablesMigrator.php
```diff
@@ -0,0 +1,473 @@
...
+	public function __construct(
+		IL10N                     $l10n,
+		TableMapper               $tableMapper,
+		ColumnMapper              $columnMapper,
+		RowSleeveMapper           $rowSleeveMapper,
+		ViewMapper                $viewMapper,
+		ContextMapper             $contextMapper,
+		ShareMapper               $shareMapper,
+		ContextNodeRelationMapper $contextNodeRelationMapper,
+		FavoritesService          $favoritesService,
```
juliusknorr: Please use https://stitcher.io/blog/constructor-promotion-in-php-8

lib/Service/ContextService.php
samin-z: understandable, i'm currently working on that.

lib/Service/ColumnService.php
samin-z: i looked into it and its best not to use `Column::fromRow($column);`, the reason for that is the imported data structure may not exactly match the expected structure and fromRow does not handle type conversions, thus i need to use setters which defeats the whole purpose of using fromRow.

juliusknorr: All right, then let's keep it.

lib/Service/ContextService.php
samin-z: it's applied so the new owner is current user.

lib/Service/ShareService.php
```diff
@@ -578,4 +583,58 @@ public function findSharedWithUserIds(int $elementId, string $elementType): arra
...
+	 */
+	public function importShare(int $nodeId, array $share, string $userId): void {
+		$existingReceiver = $this->userManager->userExists($share['receiver']);
+
+		if (!$existingReceiver) {
+			return;
+		}
+
+		$nodeType = $share['nodeType'] ?? 'table';
+		$receiver = $existingReceiver;
```
benjaminfrueh: Shouldn't this still be $share['receiver'] here, $existingReceiver is a boolean, and for the share create() method a receiver string is expected.

lib/Db/RowCellBulkFetchTrait.php
```diff
@@ -0,0 +1,31 @@
...
+	 * @return array
+	 */
+	public function findAllByRowIdsAndColumnIds(array $rowIds, array $columnIds): array {
+		if (empty($rowIds) || empty($columnIds)) {
+			return [];
+		}
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->table)
+			->where($qb->expr()->in('row_id', $qb->createNamedParameter($rowIds, IQueryBuilder::PARAM_INT_ARRAY)))
```
benjaminfrueh: This could potentially cause problems with large row datasets. For example SQLite has a 999 parameter limit in the Nextcloud supported version 3.24+ and would fail with 1000 $rowIds. Maybe should consider selecting it in chunks, e. g. looping $rowIds with array_chunk.

lib/Db/ShareMapper.php
```diff
@@ -199,4 +199,37 @@ public function deleteByNode(int $nodeId, string $nodeType):void {
...
 			->executeStatement();
 	}
+
+	/**
+	 * @param int[] $tableIds
+	 * @param int[] $contextIds
+	 *
+	 * @return Share[]
+	 */
+	// add better function name
```
benjaminfrueh: To keep the it consistent with the other existing functions in ShareMapper, my suggestion for function name would be something like `findAllSharesForTablesAndContexts`. Open for discussion of course.

lib/Service/FavoritesService.php
```diff
@@ -115,4 +116,46 @@ private function checkAccessToNode(int $nodeType, int $nodeId): void {
...
 	}
 
+	/**
+	 * @param string|null $userId
+	 *
+	 * @return array<int, array<string, mixed>>
+	 *
+	 * @throws Exception
+	 */
+	public function findAll(?string $userId = null): array {
```
benjaminfrueh, line 127: To be consistent, services should use db mappers and not query the database directly. A `FavoritesMapper` in `OCA\Tables\Db` could be nice. But I saw that this service is already mixing concerns in other methods too.

Maybe this is out of scope of this PR to refactor, but it could improve testability and enable reuse across services.

lib/Service/RowService.php
```diff
@@ -843,4 +847,39 @@ private function getColumnBusiness(Column $column): IColumnTypeBusiness {
...
+
+	/**
+	 * @param Table $table
+	 * @param array $row
+	 *
+	 * @return Row2
+	 *
+	 * @throws InternalError
+	 */
+	public function importRow(Table $table, array $row): Row2 {
```
benjaminfrueh: This should ideally happen in the existing `RowSleeveMapper`so this service does not need the db connection. Also the `Row2` Object seems not needed here, as only the id from the return value is used in the `TablesMigrator`.

lib/TablesMigrator.php
```diff
@@ -0,0 +1,425 @@
...
+			$exportDestination->addFileContents(self::FILE_ROW_CELL_TEXT, json_encode($rowCellText));
+
+			$rowCellDateTime = $this->rowCellDatetimeMapper->findAllByRowIdsAndColumnIds($rowIds, $columnIds);
+			$exportDestination->addFileContents(self::FILE_ROW_CELL_DATETIME, json_encode($rowCellDateTime));
+
+			$rowCellUserGroup = $this->rowCellUsergroupMapper->findAllByRowIdsAndColumnIds($rowIds, $columnIds);
+			$exportDestination->addFileContents(self::FILE_ROW_CELL_USERGROUP, json_encode($rowCellUserGroup));
+
+		} catch (\Throwable $e) {
+			file_put_contents('/tmp/tables_export_error.log', $e->getMessage() . "\n" . $e->getTraceAsString(), FILE_APPEND);
```
benjaminfrueh: This is possibly just debug code leftover, should not write errors to /tmp/ here. Also it is rethrowing $e, I think, following the pattern of the other migrators, it should throw it as a `TableMigratorException` here which extends `OCP\UserMigration\UserMigrationException`

lib/TablesMigrator.php
```diff
@@ -0,0 +1,425 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Tables;
```
benjaminfrueh: Nitpick: The `TableMigrator` is currently directly in `lib/` which might be too prominent for an optional app-enabled feature. Consider moving it to `lib/UserMigration/` with namespace `OCA\Tables\UserMigration` to match the structure of the other apps implementing user migration.

lib/TablesMigrator.php
```diff
@@ -0,0 +1,425 @@
...
+
+	/**
+	 * @param array $tableIds
+	 *
+	 * @return array
+	 */
+	private function getAllViewsForTableIds(array $tableIds): array {
+		$views = [];
+		foreach ($tableIds as $tableId) {
+			$views = array_merge($views, $this->viewMapper->findAll($tableId));
```
benjaminfrueh: Nitpick: Using `array_merge()` in a loop is in general bad practice as it has O(n²) complexity as it copies all existing elements.  Consider using array_push with spread operator here instead, which appends without copying.

lib/Service/ShareService.php
samin-z: you're absolutly right, it was a last minute change i made and i missed it.

lib/Service/FavoritesService.php
samin-z: yeah exactly because of this reason i decided not to do it for now as it was exceeding the scope of this task and the time sensitivity of it, ill leave a comment to be update later.

lib/Service/RowService.php
samin-z: i did not use RowSleeveMapper because it only works with RowSleeve entities, this service and its consumer expect a Row2 object.

benjaminfrueh: So this is similiar to what insert() in the https://github.com/nextcloud/tables/blob/main/lib/Db/Row2Mapper.php#L696 does? Could maybe add a import() in the Row2Mapper and reuse the `createRowSleeveFromExistingData`() for it?

samin-z: when importing rows, only row metadata (like tableId, createdBy, createdAt...) at this stage are imported and data is not available in the import and it's inserted separately after the row is created.
this is different from the UI workflow, where cell data is included in the initial row creation request, hence why i can't use row2mapper here as well.

Issue #1061, opened 2024-05-05
More than 1000 expressions in a list are not allowed on Oracle


Markoise:
### Steps to reproduce

I have displayed a table that has over 2000 rows.

### Expected behavior

No error message

### Actual behavior

An error message appears in the Nextcloud log, which is probably caused by the table app. 

### Tables app version

0.7.1

### Browser

Brave 124

### Client operating system

Windows 11, Android

### Operating system

Debian

### Web server

Nginx

### PHP engine version

PHP 8.2

### Database

MariaDB

### Additional info

[...]

PR #2293, opened 2026-01-31
Perf: Optimize shares loading by eliminating N+1 queries

Koc:
Current implementation `SharesService::findTablesSharedWithMe()` generates repeatable queries for each view/table/group/circle.

This PR optimizes loading of that data by execution single `WHERE IN` query for each entity, so we have constant queries count.

Inspired by https://github.com/nextcloud/tables/pull/1749.

Next steps for optimization (separate PRs):
- `TableService::enhanceTable()`
- `ViewService::enhanceView()`

### DIFF
--- a/lib/Db/ShareMapper.php
+++ b/lib/Db/ShareMapper.php
@@ -100,14 +100,25 @@ class ShareMapper extends QBMapper {
 	 * @throws Exception
 	 */
 	public function findAllSharesFor(string $nodeType, array $receivers, string $userId, ?string $receiverType = 'user'): array {
-		$qb = $this->db->getQueryBuilder();
-		$qb->select('*')
-			->from($this->table)
-			->where($qb->expr()->in('receiver', $qb->createNamedParameter($receivers, IQueryBuilder::PARAM_STR_ARRAY)))
-			->andWhere($qb->expr()->neq('sender', $qb->createNamedParameter($userId, IQueryBuilder::PARAM_STR)))
-			->andWhere($qb->expr()->eq('node_type', $qb->createNamedParameter($nodeType, IQueryBuilder::PARAM_STR)))
-			->andWhere($qb->expr()->eq('receiver_type', $qb->createNamedParameter($receiverType, IQueryBuilder::PARAM_STR)));
-		return $this->findEntities($qb);
+		if (!$receivers) {
+			return [];
+		}
+
+		$chunks = [];
+		// deduct extra parameters (sender, node type, receiver type)
+		foreach (array_chunk($receivers, 1000 - 3) as $receiversChunk) {
+			$qb = $this->db->getQueryBuilder();
+			$qb->select('*')
+				->from($this->table)
+				->where($qb->expr()->in('receiver', $qb->createNamedParameter($receiversChunk, IQueryBuilder::PARAM_STR_ARRAY)))
+				->andWhere($qb->expr()->neq('sender', $qb->createNamedParameter($userId)))
+				->andWhere($qb->expr()->eq('node_type', $qb->createNamedParameter($nodeType)))
+				->andWhere($qb->expr()->eq('receiver_type', $qb->createNamedParameter($receiverType)));
+
+			$chunks[] = $this->findEntities($qb);
+		}
+
+		return array_merge(...$chunks);
 	}
 
 	/**
--- a/lib/Db/TableMapper.php
+++ b/lib/Db/TableMapper.php
@@ -70,16 +70,18 @@ class TableMapper extends QBMapper {
 			return $result;
 		}
 
-		$qb = $this->db->getQueryBuilder();
-		$qb->select('*')
-			->from($this->table)
-			->where($qb->expr()->in('id', $qb->createNamedParameter(array_keys($missing), IQueryBuilder::PARAM_INT_ARRAY)));
+		$missing = array_keys($missing);
+		foreach (array_chunk($missing, 1000) as $missingChunk) {
+			$qb = $this->db->getQueryBuilder();
+			$qb->select('*')
+				->from($this->table)
+				->where($qb->expr()->in('id', $qb->createNamedParameter($missingChunk, IQueryBuilder::PARAM_INT_ARRAY)));
 
-		$entities = $this->findEntities($qb);
-		foreach ($entities as $entity) {
-			$id = $entity->getId();
-			$this->cache[(string)$id] = $entity;
-			$result[$id] = $entity;
+			foreach ($this->findEntities($qb) as $entity) {
+				$id = $entity->getId();
+				$this->cache[(string)$id] = $entity;
+				$result[$id] = $entity;
+			}
 		}
 		return $result;
 	}
--- a/lib/Db/ViewMapper.php
+++ b/lib/Db/ViewMapper.php
@@ -70,17 +70,19 @@ class ViewMapper extends QBMapper {
 			return $result;
 		}
 
-		$qb = $this->db->getQueryBuilder();
-		$qb->select('v.*', 't.ownership')
-			->from($this->table, 'v')
-			->innerJoin('v', 'tables_tables', 't', 't.id = v.table_id')
-			->where($qb->expr()->in('v.id', $qb->createNamedParameter(array_keys($missing), IQueryBuilder::PARAM_INT_ARRAY)));
+		$missing = array_keys($missing);
+		foreach (array_chunk($missing, 1000) as $missingChunk) {
+			$qb = $this->db->getQueryBuilder();
+			$qb->select('v.*', 't.ownership')
+				->from($this->table, 'v')
+				->innerJoin('v', 'tables_tables', 't', 't.id = v.table_id')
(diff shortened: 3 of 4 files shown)


lib/Helper/CircleHelper.php
```diff
@@ -11,6 +11,7 @@
 use OCA\Circles\Model\Circle;
 use OCA\Circles\Model\Member;
 use OCA\Circles\Model\Probes\CircleProbe;
+use OCA\Tables\Errors\InternalError;
```
blizzz: i suppose that's not needed?

lib/Service/ShareService.php
```diff
@@ -203,52 +205,38 @@ public function findTablesSharedWithMe(?string $userId = null): array {
...
-			// get all views or tables that are shared with me by circle
-			if ($this->circleHelper->isCirclesEnabled()) {
-				$userCircles = $this->circleHelper->getUserCircles($userId);
-
-				foreach ($userCircles as $userCircle) {
-					$shares = $this->mapper->findAllSharesFor($elementType, $userCircle->getSingleId(), $userId, ShareReceiverType::CIRCLE);
-					$elementsSharedWithMe = array_merge($elementsSharedWithMe, $shares);
-				}
-			}
+			$userCircles = $this->circleHelper->getUserCircles($userId);
```
blizzz, line 229: we need the `isCirclesEnables` check, otherwise it will break when circles is not enabled. IIRC. Or did you verfiy this case?

lib/Db/ViewMapper.php
```diff
@@ -50,6 +50,41 @@ public function find(int $id): View {
...
 	}
 
+	/**
+	 * @param int[] $ids
+	 * @return array<int, View> indexed by view id
+	 */
+	public function findMany(array $ids): array {
+		$missing = [];
+		$result = [];
+		foreach ($ids as $id) {
```
blizzz, line 60: probably more theoretical than not, it must not be more than 1k $ids. On the safe side it would be be wrapped in array_chunk, for instance.

lib/Db/ViewMapper.php
```diff
@@ -50,6 +50,41 @@ public function find(int $id): View {
...
+
+		if (!$missing) {
+			return $result;
+		}
+
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('v.*', 't.ownership')
+			->from($this->table, 'v')
+			->innerJoin('v', 'tables_tables', 't', 't.id = v.table_id')
+			->where($qb->expr()->in('v.id', $qb->createNamedParameter(array_keys($missing), IQueryBuilder::PARAM_INT_ARRAY)));
```
blizzz: same here

lib/Db/TableMapper.php
```diff
@@ -50,6 +50,40 @@ public function find(int $id): Table {
...
 	}
 
+	/**
+	 * @param int[] $ids
+	 * @return array<int, Table> indexed by table id
+	 */
+	public function findMany(array $ids): array {
+		$missing = [];
+		$result = [];
+		foreach ($ids as $id) {
```
blizzz, line 60: same as https://github.com/nextcloud/tables/pull/2293/changes#diff-14a147e0e87c879c9a766b4dae24347823dfb815b9f7b7d1b6871c14268f88f2R60

lib/Db/ShareMapper.php
```diff
@@ -91,19 +91,19 @@ public function findShareForNode(int $nodeId, string $nodeType, string $receiver
...
 	 * @param string $userId
 	 * @param string|null $receiverType
 	 *
-	 * @return array
+	 * @return Share[]
 	 *
 	 * @throws Exception
 	 */
-	public function findAllSharesFor(string $nodeType, string $receiver, string $userId, ?string $receiverType = 'user'): array {
+	public function findAllSharesFor(string $nodeType, array $receivers, string $userId, ?string $receiverType = 'user'): array {
```
blizzz, line 102: To double check, did you verify all consumers are adjusted?

lib/Db/ViewMapper.php
Koc: yeah, in theory it's can be possible, see https://github.com/nextcloud/tables/pull/2242, but limit is 65535 for Postgres. Not sure, that it's real case scenario to have so many views. For rows definitely, but for views or tables... I'm not sure.

If you really want to add `array_chunk` here - just confirm and I will do it.

lib/Helper/CircleHelper.php
Koc: This is [needed](https://github.com/nextcloud/tables/blob/e6ee749b669954c992da08b12601af105517704f/lib/Helper/CircleHelper.php#L80) to fix `@throws` tag.

If you want - I can move that change in a separate PR. Just confirm and I will do

lib/Service/ShareService.php
Koc: For sure, and this check still [present](https://github.com/nextcloud/tables/blob/e6ee749b669954c992da08b12601af105517704f/lib/Helper/CircleHelper.php#L83-L85) in the mentioned `CircleHelper::getUserCircles()`. I've just removed duplicated check that has no sense

lib/Db/ShareMapper.php
Koc: I've manually tested next scenarios:
- shared table for user
- shared table for group
- shared table for collective
- shared view for user
- shared view for group
- shared view for collective

all scenarios pass checks ✔️. But feel free to checkout branch and test deeply

lib/Helper/CircleHelper.php
blizzz: Then it is OK for me, it looked like a forgotten line on first sight.

blizzz: On a second glance, `getUserCircles()` does not ever throw it. So it is superfluous there already. Nitpick :shrug:

lib/Db/ViewMapper.php
blizzz: The server will complain at 1000 for Oracle (cf. #1061). Smallest possible value wins there.

I am a bit torn about a defensive, more solid approach and little YAGNI. Views is more likely to reach that number. In the regular case, it won't really have much of an impact either. Let's do it with chunking.

lib/Db/ShareMapper.php
blizzz: Yes, looks good!

lib/Db/ViewMapper.php
Koc: added trait to determinate max available parameters for db vendor + chunking for shares/tables/views

lib/Helper/CircleHelper.php
Koc: removed import and `@throws` annotation

lib/AppInfo/Application.php
```diff
@@ -73,7 +73,7 @@ public function register(IRegistrationContext $context): void {
 			throw new Exception('Cannot include autoload. Did you run install dependencies using composer?');
 		}
 
-		$context->registerService(AuditLogServiceInterface::class, fn (IAppContainer $c) => $c->query(DefaultAuditLogService::class));
+		$context->registerService(AuditLogServiceInterface::class, fn (ContainerInterface $c) => $c->get(DefaultAuditLogService::class));
```
Koc, line 76: This change needed to pass [pipeline](https://github.com/nextcloud/tables/actions/runs/22025606410/job/63641652045?pr=2293)

upd: ah, newermind, I have no idea how to fix psalm, see https://github.com/nextcloud/server/pull/58261 for more details

blizzz: Fixed in https://github.com/nextcloud/tables/pull/2328

lib/Db/ViewMapper.php
blizzz: It would still run against https://github.com/nextcloud/server/blob/master/lib/private/DB/QueryBuilder/QueryBuilder.php#L225

We avoid special solutions for different DB engines, or move it as much up as possible in the stack, if there has to be some sort of handling. In this case the change would also be better moving up; the risk is that it becomes harder to spot such cases.

Koc: mate, sorry, I didn't get what to do here?

A. just hardcode 1000 for all DB engines
B. open separate PR to the server repo (we can wait years in that case)
C. something else?

Koc: BTW there is [similar PR](https://github.com/nextcloud/tables/pull/1749/changes) from you where you don't care about parameters count

blizzz: > mate, sorry, I didn't get what to do here?
> 
> A. just hardcode 1000 for all DB engines B. open separate PR to the server repo (we can wait years in that case) C. something else?

A is most straight forward

blizzz: > BTW there is [similar PR](https://github.com/nextcloud/tables/pull/1749/changes) from you where you don't care about parameters count

Mistake/bug of mine then :flushed:

Koc: @blizzz done, please check

PR #2322, opened 2026-02-12
fix(FileAction): have to use nc/files 4.0 to register file action in >= 33

blizzz:
The previous libraries version in 3.x is not compatible anymore, hence we have to cut ties.

To reproduce: 
1. have Tables enabled
2. Go to Files
3. Upload a CSV file
4. Open file actions
5. Find **Import into Tables** 

Screenshot_20260212_153615

:warning: Drops support in 2.x (unreleased) for NC <= 32.

### DIFF
--- a/src/file-actions.js
+++ b/src/file-actions.js
@@ -2,7 +2,7 @@
  * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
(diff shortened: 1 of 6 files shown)


src/file-actions.js
```diff
@@ -29,6 +29,7 @@ const fileAction = new FileAction({
 	},
 
 	exec: async (file) => {
+		console.debug(file)
```
benjaminfrueh: Should this debug statement be removed?

blizzz: Yes. I thought I did.

PR #2347, opened 2026-03-02
feat: improve pagination block

### LINKED ISSUE #2073: Embedded View has only data from first page
### Steps to reproduce

1.Use smart picker to add a 'view' to a page -- see full content mode.  View has two pages of data.
2.Search data.  Not all data for search appeared as it does in the full tables app.
3.Notice there is no way to switch pages.
4. Notice the data found is from first page and data not found is on second page.

### Expected behavior

Embedded view search sees all data.  Also see a pagination control.

### Actual behavior

Subset of data seen -- just first page, it seems.  

### Tables app version

1.0.0 beta 3

### Browser

Firefox

### Client operating system

Windows

### Operating system

Linux Mint

### Web server

Apache

### PHP engine version

PHP 8.3

### Database

MariaDB

### Additional info

Thanks for supporting this embedding feature!  Previously could not get it working at all.  Just updated to NC32rc3.  I want this feature because some people I volunteer with are old and this will make it easy for me to give them access via Collectives.

Koc:
Current pagination block has few drawbacks:
- it is not possible to tune how many items per page to display
- block located in the very bottom of the page on the left side. So it is very inconvenient for wide table with multiple rows (you need to scroll all the time)
- pagination not visible in the embedded views (closes #2073)

<details><summary> :mag:  Screenshoots</summary>
<p>

Initial view  
image

Large table scrolled to the bottom  
image

Large table scrolled to the right with few rows selected  
image

Embedded table into collective  
image
</p>
</details> 

So, this PR aims to fix all of them.

:warning: Styles are not perfect (I'm not CSS master, sorry :see_no_evil:) but it's better than it now. Feel free to improve anything or just pushing directly into the branch

### DIFF
--- a/src/shared/components/ncTable/NcTable.vue
+++ b/src/shared/components/ncTable/NcTable.vue
@@ -43,7 +43,7 @@ deselect-all-rows        -> unselect all rows, e.g. after deleting selected rows
 <template>
 	<div ref="table" class="NcTable" data-cy="ncTable">
 		<div class="options row" style="padding-right: calc(var(--default-grid-baseline) * 2);">
-			<Options :rows="rows" :columns="parsedColumns" :element-id="elementId" :is-view="isView"
+			<Options :rows="getSearchedAndFilteredAndSortedRows" :all-rows="rows" :columns="parsedColumns" :element-id="elementId" :is-view="isView"
 				:selected-rows="localSelectedRows" :show-options="parsedColumns.length !== 0"
 				:view-setting.sync="localViewSetting" :config="config" @create-row="$emit('create-row')"
 				@download-csv="data => downloadCsv(data, parsedColumns, downloadTitle)"
@@ -52,7 +52,7 @@ deselect-all-rows        -> unselect all rows, e.g. after deleting selected rows
 		</div>
 		<div class="custom-table row">
 			<CustomTable v-if="config.canReadRows || (config.canCreateRows && rows.length > 0)" :columns="parsedColumns"
-				:rows="rows" :is-view="isView" :element-id="elementId" :view-setting.sync="localViewSetting"
+				:rows="getSearchedAndFilteredAndSortedRows" :is-view="isView" :element-id="elementId" :view-setting.sync="localViewSetting"
 				:config="config" @create-row="$emit('create-row')"
 				@edit-row="rowId => $emit('edit-row', rowId)"
 				@create-column="$emit('create-column')"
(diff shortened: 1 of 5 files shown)


Aveyron-RetD:
Maybe this could solve ? #2073

src/shared/components/ncTable/NcTable.vue
```diff
@@ -220,6 +230,153 @@ export default {
...
+			}
+
+			for (const row of this.rows) {
+				if (!row || !row.data) {
+					continue
+				}
+
+				if (debug) {
+					console.debug('new row ===============================================', row)
+				}
```
enjeck, line 273: we don't need this

src/shared/components/ncTable/NcTable.vue
enjeck, line 1: select a row first, then apply search/filter so that selected row disappears, then use the “selected rows” export action. There's an error:

```
vue.runtime.esm.js:3065 TypeError: Cannot read properties of undefined (reading 'id')
    at exportTableMixin.js:24:31
    at Array.forEach (<anonymous>)
    at VueComponent.downloadCsv (exportTableMixin.js:23:9)
    at download-csv (NcTable.vue:430:1)
    at invokeWithErrorHandling (vue.runtime.esm.js:3033:30)
    at VueComponent.invoker (vue.runtime.esm.js:1832:20)
    at invokeWithErrorHandling (vue.runtime.esm.js:3033:30)
    at Vue2.$emit (vue.runtime.esm.js:3732:17)
    at VueComponent.exportCsv (Options.vue:157:1)
```

I guess cuz selectedRows is still populated, but Options.vue now resolves rows only from the filtered list, so hidden selected IDs become undefined

src/shared/components/ncTable/NcTable.vue
enjeck: and other debug

Koc: that was added before me (3 years ago). I've just moved already existent code from one file to another. But if you want - I can remove it, not a big deal

image

enjeck: whoops, didn't realize!

src/shared/components/ncTable/NcTable.vue
Koc: good catch, thanx!

Fixed by introduction `all-rows` prop to `Options` component.

enjeck (approved):
Cool,  Thanks! Please check merge conflict

Koc:
how is that happened? I haven't any conflicts just 8 minutes ago :open_mouth:

Koc:
resolved conflicts that was caused by merging of the pinned columns PR.

Pin columns works fine, pagination as well.

image

PR #2348, opened 2026-03-02
Client-side only enforcement of textMaxLength

samin-z:
Backend enforcement of textMaxLength, Min and Max values for table cells.

### DIFF
--- a/lib/Service/RowService.php
+++ b/lib/Service/RowService.php
@@ -224,11 +224,6 @@ class RowService extends SuperService {
 			throw new InternalError('Cannot create row without table or view in context');
 		}
 
-		$columnsById = [];
-		foreach ($columns as $column) {
-			$columnsById[$column->getId()] = $column;
-		}
-
 		$tableId = $tableId ?? $view->getTableId();
 
 		$data = $data instanceof RowDataInput ? $data : RowDataInput::fromArray($data);


blizzz (approved):

lib/Service/RowService.php
```diff
@@ -224,6 +224,16 @@ public function create(?int $tableId, ?int $viewId, RowDataInput|array $data): R
 			throw new InternalError('Cannot create row without table or view in context');
 		}
 
+		$columnsById = [];
+		foreach ($columns as $column) {
+			$columnsById[$column->getId()] = $column;
+		}
+
+		foreach ($data as $entry) {
+			$column = $columnsById[$entry['columnId']];
```
enjeck: What if it's an invalid `columnId`? Then `$columnsById[$entry['columnId']]` won't exist and `$column` null? Which breaks `validateColumnValueLimits()`?

enjeck: maybe moving `validateColumnValueLimits()` into `cleanupAndValidateData()` is more robust

enjeck (commented):
I have a concern

lib/Service/RowService.php
samin-z: @enjeck good point, cleaned up unnecessary checks.

PR #2357, opened 2026-03-04
Fix: fix export table to csv with deleted options

Koc:
## Reproducer:
1. Create a table with multi-select column (options: `a, b, c`)
2. Create a row with options `a, c`
3. Delete option `c`
4. Export table to csv

Expected result: table exported to csv without removed option
Actual result: there is an error in the console, csv file not created


benjaminfrueh (approved):

enjeck (approved):
thx

PR #2387, opened 2026-03-11
Add possibility filter select/multiselect columns with dropdown

Koc:
Right now it is not possible to to search select/multiselect columns using `OR` operator (contains "aaa" or "bbb") using plain text filters. This PR adds possibility to filter select/multiselect with droprowns instead of plain text. Also it can be useful for large tables when you don't know full list of possible dropdown values

## :mag: Preview
This is how it works now  
image

This is how it works with patch  
image  
image  
image

PR #2405, opened 2026-03-15
Enhancement: Extract method for getting filters for column

Koc:
During work on https://github.com/nextcloud/tables/pull/2387 I've realized that we have some code duplication. Let's extract it into separate function. This logic going to be extended in the mentioned PR:

```js
export function getFiltersForColumn(column, viewSetting) {
	if (viewSetting?.filter?.length > 0) {
		return viewSetting.filter.filter(filter => {
			if (filter.columnId !== column.id) {
				return false
			}
			if (filter.operator.id === FilterIds.ContainsItem) {
				return filter.value.length > 0
			}

			return true
		})
	}
	return []
}
```

### DIFF
--- a/src/shared/components/ncTable/mixins/filter.js
+++ b/src/shared/components/ncTable/mixins/filter.js
@@ -126,3 +126,10 @@ export const Filters = {
 		noSearchValue: true,
 	}),
 }
+
+export function getFiltersForColumn(column, viewSetting) {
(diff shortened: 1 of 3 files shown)


Koc:
@samin-z sadly your approval has not power here :crying_cat_face: 

image

samin-z:
> @samin-z sadly your approval has not power here 😿

@Koc you should be able to do it now :)

PR #2406, opened 2026-03-16
validate text column value limit only if the limit in the DB is a valid value

orandev:
On some Nextcloud platforms, there are columns of the tables app in the DB with text_max_length equal to -1. These columns cannot be filled in with the current code.
Issue on our platform with these columns started after upgrading to Tables app 1.0.5.
The error thrown is "An bad request was encountered: Value for column What to do exceeds maximum length of -1".

### DIFF
--- a/lib/Service/RowService.php
+++ b/lib/Service/RowService.php
@@ -877,7 +877,7 @@ class RowService extends SuperService {
 	 */
 	private function validateColumnValueLimits(Column $column, $value): void {
 		$textMaxLength = $column->getTextMaxLength();
-		if ($textMaxLength !== null && is_string($value) && mb_strlen($value) > $textMaxLength) {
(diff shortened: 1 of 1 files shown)


samin-z:
this issue has already been resolved with this taks
https://github.com/nextcloud/tables/issues/2413

orandev:
Hello @samin-z 
Could you backport https://github.com/nextcloud/tables/pull/2414 fix to v1.x.x please? I updated tables app to 1.0.9 and the issue is still there.

samin-z:
hello @orandev 
which version is effected?

orandev:
Hello @samin-z 
1.0.9 (and 1.0.5, I didn't test the other ones)
I'm on Nextcloud 32.

PR #2407, opened 2026-03-18
🖼️ Add persistent per-view layout modes (table / tiles / gallery)

### LINKED ISSUE #313: Add view mode: Tiles

-

Let the user select the columns to show up in a tile. Click on the tile will open a modal with all details.

_No response_

### Additional context

_No response_

Rello:
## Motivation

- Implement optional gallery and tile views, as available in SharePoint
- Both views are using the first column to get the picture url for the preview
- `Tile` is only showing the picture plus its title from the 2nd column. *Use case: Index pages*
- `Gallery` view is showing all additional columns from the current view. *Use case: Product catalogue or inventory*
- Provide persistent, per-view layout modes so users can choose between `table`, `tiles` and `gallery` and have that preference stored with the view.  
- Expose layout selection in the View settings modal and apply the persisted layout automatically when a view is opened, while keeping backward compatibility (missing/old values => `table`).

### Source data

Bildschirmfoto 2026-03-19 um 20 34 35

### Tile View

Bildschirmfoto 2026-03-19 um 20 52 47

### Gallery View

Bildschirmfoto 2026-03-19 um 20 53 05

### Settings modal

Bildschirmfoto 2026-03-19 um 20 54 28

## Sharepoint reference 

565593259-fda68f2f-2a04-4ace-b1f9-fb2efbc6c77a-2

### 🖼️ Screenshots

🌞 Light | 🌚 Dark
---|---
view-3|view-3
view-4|view-4
view-5|view-5
view-6|view-6
view-7|view-7
view-8|view-8

---

## 🧪 Andy: I tested:

* gallery/tiles view in the app as views and part of an app/context, text-only, image-only, image+text and opened as user share
* embedded widget in text file/app for tile/gallery
* public links tile/gallery
* import/export of the structure

## 📌 Remarks

* **Known limitations:** card layouts render text-only for anonymous viewers of a public link (`/core/preview` needs a session), and a federated receiver gets the layout but not the card sources (column ids are local), so it renders text-only there too.
* The footer pagination now also shows in the table layout — deliberate, not a side effect of the card layouts.

---

Also unifies file-preview link handling with the new card layout.

Previews now work for fileId/FILEID/OPENFILE links (the old matcher only took lowercase fileid — not the spelling the server writes). They no longer resolve for foreign hosts, which previously previewed the local file with that id, nor for ids buried in unrelated query values, since links are parsed rather than pattern-matched. Normal links are unchanged; attributes.fileId is now validated. Adds the missing test coverage.

Reverse proxies unaffected — overwrite.cli.url is CLI-only, so web links carry the browser's own host. Multiple trusted domains can differ, but the picker's attributes.fileId wins over parsing; only legacy values with a foreign host and no fileId fall back to the plain link widget.


jancborchardt (dismissed):
@marcoambrosini one for you, check it out. :)

@Rello do you have reference screenshots from Sharepoint as you mention?
List and Gallery view currently also look very similar (with Gallery in need of some design improvement).

marcoambrosini (commented):
Hi @Rello, apologies for the late review

I think this is a nice addition, a few comments from the design perspective:

- I don't think we should map the cards backgrounds and titles to the first columns of the table. Instead we could provide two select element in the settings below the view selection radio: "Background source" and " Title source", with the ability to select the right columns from there

- Also in settings: instead of these radio buttons, let’s use [our Nextcloud component `NcRadioGroup`](https://nextcloud-vue-components.netlify.app/#/Components/NcRadioGroup)
    
    So this:

Image

    
    Instead of this:
    
Image

    
- The aspect ratio of the cards in gallery view is not very graceful visually, and they just grow uncontrolled if we adding many fields to them. We cannot make them scrollable because they are already within a scroll container, the page, and that would result in nested scrolling. To solve this I think we should remove the column names and limit the aspect ratio of the image to a 3:2 landscape orientation instead of 1:1. We would get something like this:

Image

- In gallery view, I would left align the title so that it matches the alignment of the text underneath

- I’m not sure if it’s my screen, but it looks like the text in gallery view is blue. Let’s make sure it’s `--color-main-text`

Rello:
@marcoambrosini 

Bildschirmfoto 2026-04-13 um 22 43 46

Bildschirmfoto 2026-04-13 um 22 44 16

Bildschirmfoto 2026-04-13 um 22 44 33

marcoambrosini (approved):

lib/Migration/Version2400Date20260901000000.php
```diff
@@ -0,0 +1,52 @@
...
+				'notnull' => false,
+				'unsigned' => false,
+			]);
+		}
+		if (!$table->hasColumn('card_title_source')) {
+			$table->addColumn('card_title_source', Types::INTEGER, [
+				'notnull' => false,
+				'unsigned' => false,
+			]);
+		}
```
AIlkiv, line 41: @Rello 
I’m planning to introduce a Kanban layout, which will also require additional columns. This led me to consider an alternative approach: instead of continuously adding new dedicated columns for each feature, we could introduce a single settings column (stored as JSON text).

This settings field would encapsulate various configuration attributes, allowing us to store extensible metadata without inflating the schema.

I see this as a more scalable and maintainable solution, as it avoids excessive schema growth and simplifies future migrations.

Example:
```
settings =
{
layout: ...,
card_background_source: ...,
card_title_source: ...,
}
```

Rello: Hi, 

thats a good idea. But I would keep the dedicated *layout* column for performance. there, your kanban would then also go in. but to derive the pure layout of a view (where table might be the standard) it would be overhead to parse the settings json every time

AIlkiv: Hi.
Are you planning to replace the card_background_source and card_title_source columns with a single settings (JSON) column, or should I try to push that change into your PR?

enjeck (commented):
Just confirming that the state when there's no image is designer-approved? Here, for a tile view, I don't see why say "No image" instead of just text-based tiling like in https://media.geeksforgeeks.org/wp-content/uploads/20231218102144/Screenshot-2023-12-18-102058.png, for example. 

Image

cc @marcoambrosini

src/modules/modals/ViewSettings.vue
```diff
@@ -53,6 +53,48 @@
...
+								<NcRadioGroupButton :label="t('tables', 'Gallery')" value="gallery" data-cy="viewLayoutGallery" />
+							</NcRadioGroup>
+						</td>
+					</tr>
+				</tbody>
+			</table>
+			<div class="layout-source-settings">
+				<div class="layout-source-settings__item">
+					<div class="layout-source-settings__label">
+						{{ t('tables', 'Background source') }}
```
enjeck, line 74: What happens if this column is deleted after being set as the background source?

src/modules/modals/ViewSettings.vue
```diff
@@ -53,6 +53,48 @@
...
+					</div>
+					<NcSelect
```
enjeck, line 76: If a column isn't selected to be accessible in the view, it shouldn't be shown as the background or title source:

Image

enjeck: both in the frontend and validated in the backend too.

Rello:
@copilot resolve the merge conflicts in this pull request

lib/Migration/Version2400Date20260901000000.php
Rello: I am adding it. I just pushed the change

src/modules/modals/ViewSettings.vue
Rello: Bildschirmfoto 2026-04-29 um 11 37 52

Rello:
> Just confirming that the state when there's no image is designer-approved? Here, for a tile view, I don't see why say "No image" instead of just text-based tiling like in https://media.geeksforgeeks.org/wp-content/uploads/20231218102144/Screenshot-2023-12-18-102058.png, for example.
2BK1ZfxUS9CBdrnTws71-prkJ5-t8t_fhbs">
> cc @marcoambrosini

@enjeck what do you mean by this? I am not sure if I understand. what would you show instead of the "no image"?

jancborchardt (commented):
@Rello As @enjeck says, it is indeed weird to show "No image", especially as it underlays the text.

Instead, it would be better to show nothing as @enjeck proposed and just focus on the text.

FYI @marcoambrosini

Rello:
> Just confirming that the state when there's no image is designer-approved? Here, for a tile view, I don't see why say "No image" instead of just text-based tiling like in https://media.geeksforgeeks.org/wp-content/uploads/20231218102144/Screenshot-2023-12-18-102058.png, for example.
> 

Hello @enjeck , @jancborchardt 

it was corrected

Bildschirmfoto 2026-06-17 um 10 34 16

src/modules/modals/ViewSettings.vue
```diff
@@ -151,6 +197,48 @@ export default {
...
+		},
+		backgroundSourceValue: {
+			get() {
+				const stored = this.mutableView?.viewSettings?.cardBackgroundSource
+				// Use stored value only if the column still exists in the accessible options
+				if (stored !== null && stored !== undefined) {
+					const exists = this.cardSourceOptions.some(opt => opt.id === stored)
+					if (exists) return stored
+				}
+				return this.cardSourceOptions[0]?.id ?? null
```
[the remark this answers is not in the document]
Rello: done

lib/Controller/ApiTablesController.php
```diff
@@ -250,6 +251,7 @@ public function createFromScheme(string $title, string $emoji, string $descripti
 					array_merge($inputColumnsArray, [
 						'sort' => $newSort,
 						'filter' => $newFilter,
+						'layout' => $view['layout'] ?? null,
 					])
```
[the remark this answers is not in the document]
Rello: done

src/shared/components/ncTable/sections/CustomTable.vue
```diff
@@ -108,9 +194,22 @@ export default {
...
+			return Array.from(
+				{ length: this.totalPages },
+				(value, index) => 1 + index,
+			)
+		},
 		currentPageRows() {
 			return this.rows.slice((this.pageNumber - 1) * this.rowsPerPage, ((this.pageNumber - 1) * this.rowsPerPage) + this.rowsPerPage)
 		},
+		totalPages() {
+			return Math.ceil(this.rows.length / this.rowsPerPage)
```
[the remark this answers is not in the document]
Rello: not relevant to this PR

Rello:
@enjeck can you give it another look when possible? thank you

src/shared/components/ncTable/sections/CustomTable.vue
```diff
@@ -50,21 +50,99 @@
...
+					<template #icon>
+						<ChevronRightIcon :size="20" />
+					</template>
+				</NcButton>
+				<NcButton type="tertiary" :disabled="totalPages === 1 || pageNumber >= totalPages" :aria-label="t('tables', 'Go to last page')" @click="pageNumber = totalPages">
+					<template #icon>
+						<PageLastIcon :size="20" />
+					</template>
+				</NcButton>
+			</div>
```
enjeck: can we reuse `PaginationBlock.vue` ?

src/shared/components/ncTable/sections/CustomTable.vue
```diff
@@ -129,6 +229,12 @@ export default {
 		},
 	},
 
+	updated() {
+		if (this.pageNumber > this.totalPages || this.totalPages === 1) {
+			this.pageNumber = this.totalPages
+		}
+	},
```
enjeck: Mutating reactive state in `updated()` could cause re-render loops, and the `totalPages === 1` branch fires on every update. `PaginationBlock` handles this with a watcher

lib/Migration/Version2400Date20260901000000.php
enjeck, line 1: ideally renamed to later version, since new migration files have been added since this was started

lib/Service/ViewService.php
```diff
@@ -293,6 +297,29 @@ protected function assertInputColumnsAreValid(View $view, string $userId, Column
 		}
 	}
 
+	/**
+	 * Ensures that card view settings reference columns that are part of the view.
+	 * @throws InvalidArgumentException
+	 */
+	protected function assertCardSourceColumnsAreValid(View $view): void {
```
enjeck: `assertCardSourceColumnsAreValid()` blocks unrelated view updates.
`ViewService::update()` calls `assertCardSourceColumnsAreValid()` on every update. That method re-validates the view's stored `cardBackgroundSource` and `cardTitleSource` against the view's column set even when the current request does not touch `viewSettings` at all.
The result is that once a card source points at a column, any later update that removes that column from the view by changing columnSettings gets rejected even though the request never mentions the card settings. The caller is then forced to also rewrite viewSettings in the same request just to make an unrelated column change succeed.

To reproduce:

```
# 1. View with two columns; cardTitleSource = column 8
PUT /index.php/apps/tables/api/1/views/3
{"data":{"columnSettings":[{"columnId":7,"order":0},{"columnId":8,"order":1}],
         "layout":"gallery",
         "viewSettings":{"cardBackgroundSource":7,"cardTitleSource":8}}}
→ 200 OK

# 2. Remove column 8 from the view, nothing else changed
PUT /index.php/apps/tables/api/1/views/3
{"data":{"columnSettings":[{"columnId":7,"order":0}]}}
→ 400  {"message":"Invalid cardTitleSource column ID: 8"}

# 3. Identical column change, but also repoint card sources to column 7
PUT /index.php/apps/tables/api/1/views/3
{"data":{"columnSettings":[{"columnId":7,"order":0}],
         "viewSettings":{"cardBackgroundSource":7,"cardTitleSource":7}}}
→ 200 OK
```

The second and third requests send the same column edit, so the only thing that changes the outcome is whether `viewSettings` gets rewritten

enjeck (commented):
more comments:

enjeck:
@Rello i fixed merge conflicts

src/shared/components/ncTable/sections/CustomTable.vue
AndyScherzinger: resolved via: 8ac1935e

src/shared/components/ncTable/sections/CustomTable.vue
AndyScherzinger: resolved via: b6e90a8d

lib/Migration/Version2400Date20260901000000.php
AndyScherzinger: resolved via: beebfecd

lib/Service/ViewService.php
AndyScherzinger: resolved via: c65523c3

enjeck:
> > Just confirming that the state when there's no image is designer-approved? Here, for a tile view, I don't see why say "No image" instead of just text-based tiling like in https://media.geeksforgeeks.org/wp-content/uploads/20231218102144/Screenshot-2023-12-18-102058.png, for example.
> 
> Hello @enjeck , @jancborchardt
> 
> it was corrected
> 
> Bildschirmfoto 2026-06-17 um 10 34 16

I personally think the text should take the full width and height of the card, but will defer to @nextcloud/designers for the final say

AndyScherzinger:
## Visual issues in the card layouts

*Edited: re-measured per card. The earlier version of this comment named body truncation as a main cause; that was measured on the wrong view and is corrected below.*

Tested on a dev instance at `4ce97f30`, 150-row table, six views (tiles and gallery × images+text / images only / text only). Row labels are of deliberately varied length, 5 to 213 characters, and descriptions vary from empty to ~200 characters. Numbers are `getBoundingClientRect()` values taken in Chrome against the running instance.

The cards that read as broken on `#/view/6` are **1, 4, 5, 6, 7, 9, 10, 11, 14, 15, 16, 17, 19, 20**. Two independent defects account for exactly that set.

### 1. Cards do not fill their grid row, leaving dead space (12 of the 14)

`.card-layout` is a grid, so every `.layout-card` is stretched to the height of the tallest card in its row — 295px here. But `.layout-card` is `display: block`, and its children are an image wrapper pinned by `aspect-ratio: 3 / 2` (always 155px) plus a body of natural height. Whatever is left over is simply empty card at the bottom.

| card | image | body | card | unused |
|-----:|------:|-----:|-----:|-------:|
| 2, 3, 8, 12, 13, 18 | 155 | 138 | 295 | **2px** — fine |
| 1, 5, 6, 10, 11, 15, 16, 20 | 155 | 118 | 295 | **22px** |
| 4, 9, 14, 19 | 155 | 64 | 295 | **76px** |

The amount of empty space is set by how much metadata a row happens to have: two short values leave 22px, a single value leaves 76px. So cards with less content look progressively more broken next to their neighbours, which is the effect visible in the screenshots.

### 2. The title banner has no height limit (cards 7 and 17)

`.layout-card__title-banner` is `position: absolute; bottom: 0` inside the 155px image wrapper, with nothing capping its height. The 213-character label produces a **184px** banner — 29px taller than the image it lives in — so it grows past the top and `.layout-card { overflow: hidden }` slices the first line off, leaving the label starting mid-sentence.

This one is not gallery-specific: `#/view/3` and `#/view/5` (both `layout=tiles`) show the same `+29px` overflow on the same two cards. Views whose title column is empty escape it only because `getCardTitle()` then falls back to the short `Row {id}` string.

### Correction to the earlier version

`.layout-card__body { max-height: 168px; overflow: hidden }` does **not** truncate anything on `#/view/6` — measured `scrollHeight` equals rendered height on all 20 cards. The truncation I reported earlier was measured on `#/view/8`, where the view still carries the Image column so a long URL wraps into the metadata list. That is mostly an artifact of how the test view was set up, though the fixed `max-height` can still cut content silently when metadata is genuinely tall.

### Fix, verified live

Injecting this into the running page clears every reported card in all four affected views:

```css
.layout-card { display: flex; flex-direction: column; }
.layout-card__body { flex: 1 1 auto; max-height: none; }
.layout-card__title-banner {
    max-height: 100%;
    display: -webkit-box;
    -webkit-line-clamp: 2;
    -webkit-box-orient: vertical;
    overflow: hidden;
}
```

| view | before | after |
|------|--------|-------|
| 6 gallery images+text | gaps on 1,4,5,6,9,10,11,14,15,16,19,20 (max 76px); banner +29px on 7,17 | max gap 2px, no overflow |
| 8 gallery text only | gaps on 4,9,14,19 (max 32px); banner +29px on 7,17 | max gap 2px, no overflow |
| 3 tiles images+text | banner +29px on 7,17 | no overflow |
| 5 tiles text only | banner +29px on 7,17 | no overflow |

### Two smaller points

- With no `cardBackgroundSource` set, the image wrapper is still rendered, so text-only views carry a 155px empty grey block on every card.
- `:alt="getCardTitle(row)"` paints the whole title into the card as alt text when a referenced file is unreadable, on top of the banner that already shows it. `alt=""` would be safer, since it carries no extra information.

AndyScherzinger:
Follow-up to the visual issues above: all of them are addressed in d7862e6b, and the branch has been rebased onto current main.

| reported above | status |
|---|---|
| 1. cards leave dead space (12 of the 14 cards) | fixed |
| 2. title banner outgrows the image and is clipped (cards 7, 17) | fixed |
| 4. text-only views reserve an empty image block | fixed |
| 5. `alt` text painted into the card when a file cannot be read | fixed |

### What changed

**Dead space.** The card is now a flex column and the body absorbs the leftover row height, so a card with less metadata no longer ends in a gap. Measured on the same data: max unused height per card **76px → 2px**, and the twelve cards listed above are clean.

**Banner height.** Instead of a fixed line count, the banner gets a line budget computed from the rendered card size and kept in step by a `ResizeObserver`. It works out to three or four lines at usual widths and ellipsizes beyond that, so the `+29px` overflow on cards 7 and 17 is gone in all four affected views.

The observer also watches the title element, because the card container keeps its size when only the text size changes — without that the budget goes stale and the banner grows to cover the whole image at 200% text. A `MIN_TITLE_LINES` constant makes the share a preference rather than a hard cap, so an enlarged font shortens the title instead of dropping it to a single line. Measured at 200% text: two lines at 59% of the image, no overflow.

**Text-only cards.** Without a background column there is no image to cover, so the title may use the whole card. A tile then reads like a note and starts at the top; a gallery card keeps its title next to the metadata below it.

**Broken images.** The image keeps its `alt` attribute, and with it its accessible name — Chrome's accessibility tree still reports `role=image` with the full name and `ignored=false` — but the visual fallback is suppressed, so an unreadable file leaves only the broken image indicator rather than painting the title a second time on top of the banner.

### One further issue found while testing

The title did not meet WCAG AA. White text on the `rgba(0,0,0,0.4)` scrim measured **2.85:1** over a white image region, against the 4.5:1 required at this size and weight, and the result depended on which picture a user happened to choose. Sampling the actual pixels behind every title gave:

| scrim | worst case | AA | AAA |
|---|---|---|---|
| 0.40 (before) | 2.85 | no | no |
| **0.55 (now)** | **4.76** | yes | no |
| 0.70 | 8.52 | yes | yes |

The scrim is now 0.55, which puts the worst case at **4.76:1 in both colour schemes**. AAA would need 0.70, which darkens the image considerably, so that is left as a deliberate choice rather than applied. The accompanying text shadow is cosmetic only — contrast ratios ignore shadows, so it carries no part of AA.

Text-only cards were already fine at 12.2:1 (dark) and 13.59:1 (light), since they use theme tokens.

### Correction to the comment above

Point 3 there attributed part of the problem to `.layout-card__body { max-height: 168px }` truncating content. That measurement was taken on `#/view/8` rather than `#/view/6`; on the view under discussion nothing was being truncated. The `max-height` has been removed anyway, as the body now flexes.

AndyScherzinger:
@enjeck @jancborchardt I polished the PR and addressed all comments as good as possible, see the commit as well as the screenshot section at the end of the top PR description.

cc @Rello FYI for the changes I made, image overlay color has been slightly darkened to fullfill a WCAG AA large text requirement.

AndyScherzinger (approved):

lib/Service/StructureService.php
```diff
@@ -213,6 +213,13 @@ protected function prepareViewsData(array $views, array $columnsMap): array {
 					unset($views[$i]['columnSettings'][$j]['columnId']);
 				}
 			}
+			foreach (['cardBackgroundSource', 'cardTitleSource'] as $sourceKey) {
+				$sourceId = $view['viewSettings'][$sourceKey] ?? null;
+				if ($sourceId !== null && isset($columnsMap[$sourceId])) {
+					$views[$i]['viewSettings'][$sourceKey . 'Uuid'] = $columnsMap[$sourceId]['uuid'];
+					unset($views[$i]['viewSettings'][$sourceKey]);
```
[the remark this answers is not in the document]
AndyScherzinger: Good catch, fixed in e873fe73. A scheme is user supplied, so `viewSettings` can be absent or arrive as a JSON string, and the string case would have raised a TypeError on the offset access rather than a notice. The loop is now guarded with `isset(...) && is_array(...)`.

openapi.json
```diff
@@ -2043,6 +2072,12 @@
                                         "nullable": true,
                                         "default": null,
                                         "description": "Technical name for the view"
+                                    },
+                                    "layout": {
+                                        "type": "string",
+                                        "nullable": true,
+                                        "default": null,
+                                        "description": "Layout for the view with 'table', 'tiles', 'gallery' or null"
                                     }
```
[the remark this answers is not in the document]
AndyScherzinger: Fixed in e873fe73. The `@param` on `createView` is now `'table'|'tiles'|'gallery'|null` and `openapi.json` was regenerated, so the request schema carries `"enum": ["table","tiles","gallery"]` and the generated type is `"table" | "tiles" | "gallery" | null` instead of `string | null`. The update payload already had the enum; only the create endpoint was missing it.

lib/Service/ViewService.php
```diff
@@ -202,6 +204,11 @@ public function create(
 			$item->setEmoji($emoji);
 		}
 		$item->setDescription('');
+		try {
+			$item->setLayout(ViewUpdateInput::normalizeLayout($layout));
+		} catch (InvalidArgumentException $e) {
+			throw new BadRequestError($e->getMessage());
+		}
```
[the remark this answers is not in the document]
AndyScherzinger: Not doing this one, because I tried it and it breaks switching a view back to the table layout.

I made `normalizeLayout()` return `null` for `'table'` so the default had one stored representation. `ViewUpdateInput::updateDetail()` only yields a parameter when it is non-null, so a null layout reads as 'not provided' and the update is silently skipped. Measured before and after on a running instance:

    create layout=table   -> API reports 'table'
    update layout=gallery -> API reports 'gallery'
    update layout=table   -> API reports 'gallery'   <- update lost

Canonicalising the other way (always storing `'table'`) would need a migration for the rows that are already `null`, including the column default. Since `View::getLayoutNormalized()` maps both to `'table'` on read, the difference is not observable through the API, so the storage inconsistency is left as it is.

lib/Service/ViewService.php
```diff
@@ -322,6 +340,72 @@ protected function assertInputColumnsAreValid(View $view, string $userId, Column
...
+	 * @return list<int>
+	 */
+	private function getCardSourceCandidateIds(View $view, string $userId): array {
+		$viewColumnIds = $view->getColumnIds();
+		if (!empty($viewColumnIds)) {
+			return $viewColumnIds;
+		}
+
+		$columnService = \OCP\Server::get(ColumnService::class);
+		return array_map(static fn (Column $column) => $column->getId(), $columnService->findAllByManagedView($view, $userId));
```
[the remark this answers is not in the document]
AndyScherzinger: Not doing this one: injecting `ColumnService` here would be a circular dependency, since `ColumnService` already takes `ViewService` in its constructor (`lib/Service/ColumnService.php:50`).

That is why the pre-existing `assertInputColumnsAreValid()` a few lines above (line 330) already resolves `ColumnService` the same way. The new method follows the convention that is already in this file rather than introducing a second one. Breaking the cycle properly would mean extracting the column lookup into its own service, which is worth doing separately rather than in this PR.

lib/Service/ViewService.php
```diff
@@ -257,23 +264,34 @@ public function update(int $id, ViewUpdateInput $data, ?string $userId = null, b
...
 				if ($parameter === ViewUpdatableParameters::TECHNICAL_NAME) {
 					$this->assertTechnicalNameValid($value);
 				}
 
-				if ($value instanceof JsonSerializable) {
-					$insertableValue = json_encode($value);
-				}
+				$insertableValue = $value instanceof JsonSerializable
+					? json_encode($value)
+					: $value;
```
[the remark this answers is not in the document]
AndyScherzinger: Leaving this one out of this PR: the line predates it. `$insertableValue = json_encode($value)` is on main today (`lib/Service/ViewService.php:272`) and this PR only added another parameter that flows through the same loop.

The concern is fair though: a filter value carrying invalid UTF-8 would make `json_encode()` return `false` and the setter would persist that. It deserves `JSON_THROW_ON_ERROR` and a controlled `BadRequestError`, applied to every JsonSerializable parameter rather than just the new one, which is a change of its own and unrelated to layout modes.

PR #2427, opened 2026-03-30
feat(tests): Migrate some tests from Cypress to Playwright

enjeck:
This should hopefully fix the issues we've been facing with flaky Cypress tests. Also, we had been pinned to an older version of Cypress since newer versions were crashing, we don't need to worry about that anymore. Plus, these tests should be faster since tests run in parallel


cypress/e2e/view-filtering-selection-row-removal.cy.js
```diff
@@ -0,0 +1,163 @@
...
+		cy.get('[data-cy="createRowBtn"]').click()
+		cy.fillInValueTextLine('title', 'sevenths row')
+		cy.fillInValueSelection('selection', 'sel2')
+		cy.fillInValueSelectionMulti('multi selection', ['A', 'C', 'B', 'D'])
+		cy.get('[data-cy="createRowSaveButton"]').click()
+
+		cy.loadTable('View filtering test table')
+	})
+
+	it('Filter view remove row when it no longer matches filter', () => {
```
samin-z, line 80: 'Filter view remove row when it no longer matches filter' exist in Cypress test but was not migrated to 'view-filtering-selection.spec.ts', was there a specific reason this test couldnt be migrated?

enjeck: It just wasn't passing on the CI despite my attempts 🤷‍♀️

samin-z (approved):

PR #2428, opened 2026-03-30
#1840: API structure improvement: expose columns through a technical name/slug/alias

samin-z:
Feature ticket: #1840 

New response structure:
```js
data: [
    {
        columnId: 70,
        value: 11
    },
    {
        columnId: 72,
        value: 3
    },
    {
        columnId: 76,
        value: [
            {
                id: user1,
                type: 0,
                displayName: user1
            }
        ]
    }
],
dataByAlias: {
    column_70: {
        columnId: 70,
        value: 11
    },
    column_72: {
        columnId: 72,
        value: 3
    },
    test_customer: {
        columnId: 76,
        value: [
            {
                id: user1,
                type: 0,
                displayName: user1
            }
        ]
    }
}
```

now we have new field added to columns called 'technical_name' which if user does not add it as default will be 'column_{id}', 'dataByAlias' added on top of 'data' not to break the current structure.
Screenshot 2026-03-30 at 15 13 32

### DIFF
--- a/lib/Migration/Version1000Date20260327000000.php
+++ b/lib/Migration/Version1000Date20260327000000.php
@@ -1,0 +1,84 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Tables\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\DB\Types;
+use OCP\IDBConnection;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+use Override;
+
+class Version1000Date20260327000000 extends SimpleMigrationStep {
+	private IDBConnection $connection;
+
+	public function __construct(IDBConnection $connection) {
+		$this->connection = $connection;
+	}
+
+	#[Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('tables_columns')) {
+			return null;
+		}
+
+		$table = $schema->getTable('tables_columns');
+		if (!$table->hasColumn('technical_name')) {
+			$table->addColumn('technical_name', Types::STRING, [
+				'notnull' => false,
+				'length' => 200,
+			]);
+		}
+
+		if (!$table->hasIndex('tables_columns_table_tech_name_uq')) {
+			$table->addUniqueIndex(['table_id', 'technical_name'], 'tables_columns_table_tech_name_uq');
+		}
+
+		return $schema;
+	}
+
+	#[Override]
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		$selectQb = $this->connection->getQueryBuilder();
+		$selectQb->select('id')
+			->from('tables_columns')
+			->where(
+				$selectQb->expr()->orX(
+					$selectQb->expr()->isNull('technical_name'),
+					$selectQb->expr()->eq('technical_name', $selectQb->createNamedParameter('')),
+				)
+			);
+
+		$result = $selectQb->executeQuery();
+		$updatedCount = 0;
+
+		try {
+			while ($row = $result->fetchAssociative()) {
+				$columnId = (int)$row['id'];
+
+				$updateQb = $this->connection->getQueryBuilder();
+				$updateQb->update('tables_columns')
+					->set('technical_name', $updateQb->createNamedParameter('column_' . $columnId, IQueryBuilder::PARAM_STR))
+					->where($updateQb->expr()->eq('id', $updateQb->createNamedParameter($columnId, IQueryBuilder::PARAM_INT)));
+
+				$updatedCount += $updateQb->executeStatement();
+			}
+		} finally {
+			$result->closeCursor();
+		}
+
+		$output->info('Version1000Date20260327000000: backfilled technical_name for ' . $updatedCount . ' columns.');
+	}
+}
--- a/lib/ResponseDefinitions.php
+++ b/lib/ResponseDefinitions.php
@@ -21,6 +21,7 @@ namespace OCA\Tables;
  * 	ownership: string,
  * 	ownerDisplayName: string|null,
  * 	createdBy: string,
+ * 	dataByAlias: array<string, array{columnId: int, value: mixed}>,
  * 	createdAt: string,
  * 	lastEditBy: string,
  * 	lastEditAt: string,
@@ -91,6 +92,7 @@ namespace OCA\Tables;
  *    createdAt: string,
  *    lastEditAt: string,
  *    data: ?array{columnId: int, value: mixed},
+ *    dataByAlias: array<string, array{columnId: int, value: mixed}>,
  *  }
  *
  * @psalm-type TablesLinkShare = array{
@@ -118,6 +120,7 @@ namespace OCA\Tables;
  * @psalm-type TablesColumn = array{
  *  id: int,
  *  title: string,
+ *  technicalName: string,
  *  tableId: int,
  *  createdBy: string,
  *  createdByDisplayName: string,
@@ -163,6 +166,7 @@ namespace OCA\Tables;
  * @psalm-type TablesPublicColumn = array{
  *   id: int,
  *   title: string,
+ *   technicalName: string,
  *   createdAt: string,
  *   lastEditAt: string,
  *   type: string,
--- a/lib/ResponseDefinitions.php
+++ b/lib/ResponseDefinitions.php
@@ -85,6 +85,7 @@ namespace OCA\Tables;
  * 	lastEditBy: string,
  * 	lastEditAt: string,
  *  data: ?array{columnId: int, value: mixed},
+ *  dataByAlias: array<string, array{columnId: int, value: mixed}>,
  * }
  *
  * @psalm-type TablesPublicRow = array{
--- a/lib/Service/ColumnService.php
+++ b/lib/Service/ColumnService.php
@@ -308,6 +308,9 @@ class ColumnService extends SuperService {
 		$item = Column::fromDto($columnDto);
 		$item->setTitle($newTitle);
 		$item->setTableId($table->getId());
+		if ($item->getTechnicalName() !== null) {
+			$this->assertTechnicalNameUnique($table->getId(), $item->getTechnicalName());
+		}
 		$this->updateMetadata($item, $userId, true);
 
 		try {
@@ -315,6 +318,15 @@ class ColumnService extends SuperService {
 		} catch (\OCP\DB\Exception $e) {
 			$this->logger->error($e->getMessage(), ['exception' => $e]);
 			throw new InternalError(get_class($this) . ' - ' . __FUNCTION__ . ': ' . $e->getMessage());
+		}
+		if ($entity->getTechnicalName() === null || $entity->getTechnicalName() === '') {
+			$entity->setTechnicalName($this->buildDefaultTechnicalName($entity->getId()));
+			try {
+				$entity = $this->mapper->update($entity);
+			} catch (Exception $e) {
+				$this->logger->error($e->getMessage(), ['exception' => $e]);
+				throw new InternalError(get_class($this) . ' - ' . __FUNCTION__ . ': ' . $e->getMessage());
+			}
 		}
 		if (isset($view) && $view) {
 			// Add columns to view(s)
@@ -366,6 +378,10 @@ class ColumnService extends SuperService {
 
 			if ($title !== null) {
 				$item->setTitle($title);
+			}
+			if ($columnDto->getTechnicalName() !== null) {
+				$this->assertTechnicalNameUnique($item->getTableId(), $columnDto->getTechnicalName(), $item->getId());
+				$item->setTechnicalName($columnDto->getTechnicalName());
 			}
 			if ($columnDto->getType() !== null) {
 				$item->setType($columnDto->getType());
@@ -674,6 +690,7 @@ class ColumnService extends SuperService {
 		$item = new Column();
 		$item->setTableId($table->getId());
 		$item->setTitle($column['title']);
+		$item->setTechnicalName($column['technicalName'] ?? null);
 		$item->setCreatedBy($table->getOwnership());
 		$item->setCreatedAt($column['createdAt']);
 		$item->setLastEditBy($table->getOwnership());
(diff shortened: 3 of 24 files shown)


lib/Service/RowService.php
enjeck, line 1: Looking at https://github.com/nextcloud/tables/issues/1840, in order for this to be useful for workflows, we must expose it via AbstractRowEvent.php -> toPublicRow() ?

lib/Service/ColumnService.php
enjeck, line 1: I believe we're now doing 2 writes? Both `create()` and `importColumn()` create a column first and only then assign the fallback column_<id> name. If the second write fails, the request throws after the insert has already succeeded, so we leave behind a persisted column without the technicalName. Maybe we should have a transaction then?

src/modules/modals/EditColumn.vue
```diff
@@ -216,6 +227,18 @@ export default {
...
+			return normalized === '' ? null : normalized
+		},
+		isTechnicalNameValid() {
+			const technicalName = this.normalizeTechnicalName(this.editColumn.technicalName)
+			if (technicalName === null) {
+				return false
+			}
+
+			return /^[a-z][a-z0-9_]*$/.test(technicalName)
+		},
```
enjeck, line 236: used here and in `CreateColumn`, could be moved to a shared file and reused

src/modules/modals/EditColumn.vue
```diff
@@ -216,6 +227,18 @@ export default {
...
 		},
+		normalizeTechnicalName(technicalName) {
+			const normalized = technicalName?.trim()
+			return normalized === '' ? null : normalized
+		},
+		isTechnicalNameValid() {
+			const technicalName = this.normalizeTechnicalName(this.editColumn.technicalName)
+			if (technicalName === null) {
+				return false
+			}
```
enjeck: we return `false` for null technicalName when we edit but `true` for creating? Why?

enjeck (commented):
some comments

lib/Service/ColumnService.php
samin-z: there is no problem even if the update of technicalName fails a 'null' value for the column is valid. When 'technicalName' is null, the code in RowService handles it:
`$alias = $aliasByColumnId[$columnId] ?? ('column_' . $columnId);`
this causes the column to behave correctly with the 'column_{columnId}' fallback.
also when the user edits and saves the column, the 'null' will be updated with either their chosen value or the default 'column_{columnId}'.

src/modules/modals/EditColumn.vue
samin-z: that wa my mistake as i was going through it and forgot to update

Koc:
hey @samin-z!

Thank you so much for taking care of this. I've just opened similar PR for Views https://github.com/nextcloud/tables/pull/2547.

Few ideas that came to my mind when I worked on it:

1. as this technical name is some kind of "advanced" feature used mostly by developers - then I moved it under special "spoiler" in order to not to burden the UI for regular users
image

2. as changes to this field can broke API/integrations - then I added some notice on change
image

WDYT about adding same things to your PR?

samin-z:
i compared playwright fails against main, failures are in known flaky datetime/login/navigation tests and are reproducible outside this branch, none of the failures point to technical_name or dataByAlias code paths.

samin-z:
@nimishavijay 

now we have advanced setting to column edit/create

does this setup align with what design had in mind or exact design for "table view"?
Screenshot 2026-05-05 at 16 27 55

Screenshot 2026-05-05 at 16 28 05

lib/Service/RowService.php
```diff
@@ -671,7 +679,9 @@ public function updateSet(
 			);
 		}
 
-		return $this->filterRowResult($view ?? null, $updatedRow);
+		$updatedRow = $this->filterRowResult($view ?? null, $updatedRow);
+		$this->attachAliasPayload($updatedRow, $columns);
```
enjeck: Did you test out the webbook event to see what the payload is? Cuz from the code, it seems like above we dispatch the event with `new RowUpdatedEvent($updatedRow, $previousData)` before even attaching `dataByAlias`? So webhook results doesn't get the `dataByAlias` value?

lib/ResponseDefinitions.php
```diff
@@ -21,6 +21,7 @@
  * 	ownership: string,
  * 	ownerDisplayName: string|null,
  * 	createdBy: string,
+ * 	dataByAlias: array<string, array{columnId: int, value: mixed}>,
```
enjeck, line 24: should this be optional

lib/Service/ColumnService.php
```diff
@@ -681,10 +698,41 @@ public function importColumn(Table $table, array $column): int {
...
+		}
+
+		foreach ($columns as $column) {
+			if ($excludeCurrentColumnId !== null && $column->getId() === $excludeCurrentColumnId) {
+				continue;
+			}
+
+			if ($column->getTechnicalName() === $technicalName) {
+				throw new BadRequestError('Technical name must be unique in the table.');
+			}
```
enjeck: I think status 400 is better, not 500

enjeck (commented):
some comments

lib/Service/RowService.php
samin-z: you're right, fixed it in RowService by ensuring alias payload is attached before event dispatch.

lib/ResponseDefinitions.php
samin-z: i'de keep it as none optional, the row serializer always returns this key and it gets normalized to an array, so clients get a stable contract anyway.

lib/Service/ColumnService.php
samin-z: it is returning 400

lib/ResponseDefinitions.php
enjeck: I think this line means that every view response has a required `dataByAlias` field? But `View::jsonSerialize()` in lib/Db/View.php does not return `dataByAlias`. I'm confused

lib/Service/ColumnService.php
enjeck: Uh, I don't think so?

```
curl -L -u admin:admin -H 'OCS-ApiRequest: true' -H 'Accept: application/json' \
  -X POST \
  -d 'baseNodeId=1' -d 'baseNodeType=table' \
  -d 'title=Bad alias v2 repro' -d 'technicalName=Bad-Name' \
  http://nextcloud.local/ocs/v2.php/apps/tables/api/2/columns/text

{"ocs":{"meta":{"status":"failure","statuscode":996,"message":"Internal Server Error\nTechnical name must start with a letter and contain only lowercase letters, digits, and underscores."},"data":[]}}%  
```

lib/ResponseDefinitions.php
samin-z: dataByAlias field is always present in row api responses, the code that builds these responses ensures it regardless of technicalName. the confusion here is because View::jsonSerialize() is unrelated to row serialization. the type definition is correct and we do not need to make dataByAlias optional unless we expect future changes that could remove it.
you can check it locally and if there is a case which i missed, please let me know.

lib/Service/ColumnService.php
samin-z: i see the confusion, when adding new (none-unique) column it returns 400, but if a column technical name is updated it return 500, that's how i missed it. ill fix this.

nimishavijay:
Yes, this makes sense! would it be possible to make it collapsible and hide it by default? like the files sharing settings?




I also realize that this is moving towards a bigger redesign of the "edit column" modal so also pinging @marcoambrosini here

samin-z:
@nimishavijay i thought about it, but since it's a new feature it's possible users might miss it. so for better user accessibility and until users get used to it i thought to leave it out in the open and in later version to improve it, what do you think?

nimishavijay:
I would still vote for being able to collapse this to make the first look of the column creation simple, what do you think @marcoambrosini ? If you insist it be shown I would then create an overview issue about redesigning this modal in general to improve alignment, hiding some details, etc

samin-z:
@nimishavijay since for views (the one @enjeck mentioned #2547 ) it's also as you mentioned (toggled) ill apply this then to have the same behavior everywhere.

nimishavijay:
Makes sense, thank you!

samin-z:
updated the design to toggle the advanced settings
Screenshot 2026-06-24 at 17 30 55
Screenshot 2026-06-24 at 17 31 01

nimishavijay:
Looks good! We can also hide the "Advanced settings" header :)

src/modules/modals/EditColumn.vue
```diff
@@ -212,6 +223,10 @@ export default {
 			delete data.createdBy
 			delete data.lastEditAt
 			delete data.lastEditBy
+
+			data.technicalName = this.normalizeTechnicalName(data.technicalName)
+			data.customSettings = { width: data.customSettings.width }
```
enjeck: doesn't this discard other settings? Should probably be:


```suggestion
			data.customSettings = { ...data.customSettings, width: data.customSettings.width }
```

src/modules/modals/EditColumn.vue
```diff
@@ -212,6 +223,10 @@ export default {
...
+			console.debug('this column data will be send', data)
```
enjeck: we don't need to keep this

lib/Service/RowService.php
```diff
@@ -773,6 +784,7 @@ public function delete(int $id, ?int $viewId, string $userId, ?int $tableId = nu
 
 		try {
 			$deletedRow = $this->row2Mapper->delete($item);
+			$this->attachAliasPayload($item);
```
enjeck: Both attachAliasPayload calls here and below omit `$columns`, so each falls back to `loadColumnsForData`, causing a separate `columnMapper->findAll` query? So that's two extra column queries per delete? Why not this?
```suggestion
			$this->attachAliasPayload($item, $columns);
```

enjeck: Also, since we call `attachAliasPayload` twice because `filterRowResult` drops `dataByAlias`, could we have filterRowResult trim/rebuild the alias map so callers don't compute it twice?

tests/unit/Validation/ColumnDtoValidatorTest.php
```diff
@@ -0,0 +1,139 @@
...
+	public function testTechnicalNameOver200CharsIsRejected(): void {
+		$this->expectException(BadRequestError::class);
+		$this->expectExceptionMessageMatches('/maximum.*200/i');
+		$this->validator->validate($this->dto('test' . str_repeat('t', 197)));
+	}
+
+	/**
+	 * @dataProvider reservedNameProvider
+	 */
+	#[DataProvider('reservedNameProvider')]
```
enjeck: do we need both `@dataProvider` annotation and the `#[DataProvider]` attribute? Isn't one enough?

enjeck (commented):
I have more comments. Would also appreciate a review from @blizzz

samin-z:
> Looks good! We can also hide the "Advanced settings" header :)

@nimishavijay i added it to be the same design as view advanced settings, if it's better to hide it then ill aligh with @Koc

lib/Migration/Version1000Date20260327000000.php
```diff
@@ -0,0 +1,84 @@
...
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\DB\Types;
+use OCP\IDBConnection;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+use Override;
+
+class Version1000Date20260327000000 extends SimpleMigrationStep {
```
blizzz: Not sure that this Migration will run as it will be sorted before `Version2010Date20260414000000.php`.

lib/Service/ColumnService.php
```diff
@@ -308,6 +308,9 @@ public function create(
 		$item = Column::fromDto($columnDto);
 		$item->setTitle($newTitle);
 		$item->setTableId($table->getId());
+		if ($item->getTechnicalName() !== null) {
+			$this->assertTechnicalNameUnique($table->getId(), $item->getTechnicalName());
+		}
```
blizzz: The assertion would already be done when inserting to the database. Relying on a potential constraint violation from the DB would also save from race conditions here:  a conflicted name could be inserted between this check and the insertion afterwards.

lib/Service/RowService.php
```diff
@@ -882,11 +891,79 @@ private function filterRowResult(?View $view, Row2 $row): Row2 {
...
+		$row->filterDataByAliasByColumns($columnIds);
 
 		return $row;
 	}
 
+	/**
+	 * @param Row2[] $rows
+	 * @param Column[]|null $columns
+	 */
+	private function attachAliasPayloads(array $rows, ?array $columns = null): void {
```
blizzz:
```suggestion
	private function attachAliasPayloads(array $rows, array $columns): void {
```

I have not seen it ever being called without parameter, right?

lib/Service/RowService.php
```diff
@@ -882,11 +891,79 @@ private function filterRowResult(?View $view, Row2 $row): Row2 {
...
+		foreach ($rows as $row) {
+			$this->attachAliasPayload($row, $columns);
+		}
+	}
+
+	/**
+	 * @param Column[]|null $columns
+	 */
+	private function attachAliasPayload(Row2 $row, ?array $columns = null): void {
```
blizzz:
```suggestion
	private function attachAliasPayload(Row2 $row, array $columns): void {
```

As above, plus the logic could be trimmed down, no?

blizzz (changes requested):
some remarks, but nothing fundamental

lib/Migration/Version1000Date20260327000000.php
samin-z: yes, this migration will run. 1000 runs before 2010, and the two changes are independent so that order is safe.
but if you see as it more safe, i can rename the migration.

lib/Service/ColumnService.php
samin-z: i agree that DB should be the source of truth, i'd still keep the check for clarity for users so they know why it has failed. the unique index still enforcess it if two requests are applied.

enjeck (approved):
🚀

tests/unit/Validation/ColumnDtoValidatorTest.php
```diff
@@ -57,7 +56,6 @@ public function testTechnicalNameOver200CharsIsRejected(): void {
 	/**
 	 * @dataProvider reservedNameProvider
 	 */
-	#[DataProvider('reservedNameProvider')]
```
blizzz: should keep the Attributes rather than then doc annotation

blizzz (approved):

samin-z:
PHPUnit fails because it does not recognize #[DataProvider] attributes.

blizzz:
> PHPUnit fails because it does not recognize #[DataProvider] attributes.

Ah, they were introduced in phpunit 11, we're here on 9.6 still I just see. I assumed when they were in the code, we were already at a matching phpunit version.

benjaminfrueh (approved):

PR #2446, opened 2026-04-05
Fix: Make case-insensitive search for the usergroup and multi selection columns

Koc:
This is a follow-up for the #1798. It spreads same logic what we have in text line type to other column types. So, now everything is case insensitive.

It also fixes filters for usergroup single selection

## :radio_button: Single selection
| Before | After |
|--------|--------|
| image | image |

## :ballot_box_with_check: Multi selection
| Before | After |
|--------|--------|
| image | image |

## :bust_in_silhouette: User single selection
| Before | After |
|--------|--------|
| image | image |

## :busts_in_silhouette: User multi selection
| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/src/shared/components/ncTable/mixins/columnsTypes/selection.js
+++ b/src/shared/components/ncTable/mixins/columnsTypes/selection.js
@@ -52,15 +52,15 @@ export default class SelectionColumn extends AbstractSelectionColumn {
 
 	isFilterFound(cell, filter) {
 		const filterValue = filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value
-		const cellLabel = this.getLabel(cell.value)
+		const cellLabel = this.getLabel(cell.value)?.toLowerCase()
 		const filterMethod = {
 			[FilterIds.ContainsItem]() { return filterValue.map(option => option.id).includes(cell.value) },
-			[FilterIds.Contains]() { return cellLabel?.toLowerCase().includes(filterValue?.toLowerCase()) },
-			[FilterIds.DoesNotContain]() { return !cellLabel?.toLowerCase().includes(filterValue?.toLowerCase()) },
-			[FilterIds.BeginsWith]() { return cellLabel?.startsWith(filterValue) },
-			[FilterIds.EndsWith]() { return cellLabel?.endsWith(filterValue) },
-			[FilterIds.IsEqual]() { return cellLabel === filterValue },
-			[FilterIds.IsNotEqual]() { return cellLabel !== filterValue },
+			[FilterIds.Contains]() { return cellLabel?.includes(filterValue.toLowerCase()) },
+			[FilterIds.DoesNotContain]() { return !cellLabel?.includes(filterValue.toLowerCase()) },
+			[FilterIds.BeginsWith]() { return cellLabel?.startsWith(filterValue.toLowerCase()) },
+			[FilterIds.EndsWith]() { return cellLabel?.endsWith(filterValue.toLowerCase()) },
+			[FilterIds.IsEqual]() { return cellLabel === filterValue.toLowerCase() },
+			[FilterIds.IsNotEqual]() { return cellLabel !== filterValue.toLowerCase() },
 			[FilterIds.IsEmpty]() { return !cellLabel },
 		}[filter.operator.id]
 		return super.isFilterFound(filterMethod, cell)
--- a/src/shared/components/ncTable/mixins/columnsTypes/selectionMulti.js
+++ b/src/shared/components/ncTable/mixins/columnsTypes/selectionMulti.js
@@ -68,7 +68,7 @@ export default class SelectionMutliColumn extends AbstractSelectionColumn {
 
 	isFilterFound(cell, filter) {
 		const filterValue = filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value
-		const valueString = this.getValueString(cell)
+		const valueString = this.getValueString(cell)?.toLowerCase()
 
 		const filterMethod = {
 			[FilterIds.ContainsItem]() {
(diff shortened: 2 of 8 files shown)


src/shared/components/ncTable/mixins/columnsTypes/usergroup.js
```diff
@@ -29,7 +30,7 @@ export default class UsergroupColumn extends AbstractUsergroupColumn {
 				ret += ', ' + obj.id
 			}
 		})
-		return ret
+		return ret.toLowerCase()
```
enjeck, line 33: Is it intentional that the search is case insentive but the other filters aren't? 
 For the Is equal / Is not equal operators. The filter UI stores whatever casing the user typed, so a row containing `Alice` now becomes `alice` before comparison. That makes Is equal "Alice" fail and Is not equal "Alice" incorrectly pass?

src/shared/components/ncTable/mixins/columnsTypes/selectionMulti.js
```diff
@@ -71,8 +71,8 @@ export default class SelectionMutliColumn extends AbstractSelectionColumn {
 		const valueString = this.getValueString(cell)
 
 		const filterMethod = {
-			[FilterIds.Contains]() { return valueString?.includes(filterValue) },
-			[FilterIds.DoesNotContain]() { return !valueString?.includes(filterValue) },
+			[FilterIds.Contains]() { return valueString?.toLowerCase().includes(filterValue.toLowerCase()) },
+			[FilterIds.DoesNotContain]() { return !valueString?.toLowerCase().includes(filterValue.toLowerCase()) },
```
enjeck, line 75: At https://github.com/nextcloud/tables/blob/main/src/shared/components/ncTable/partials/TableHeaderColumnOptions.vue#L315, there’s a small check meant to stop adding the same Contains or Does not contain filter twice. I think that conflicts with this?

enjeck (commented):
Thanks, some comments:

src/shared/components/ncTable/mixins/columnsTypes/usergroup.js
Koc: now I've replicated same logic that we have for text line. Everything is case insensitive. For a long term perspective maybe we can add some toggle related to sensitivity.

src/shared/components/ncTable/mixins/columnsTypes/selectionMulti.js
Koc: Please check now

src/shared/components/ncTable/mixins/columnsTypes/selection.js
```diff
@@ -51,12 +51,12 @@ export default class SelectionColumn extends AbstractSelectionColumn {
 	}
 
 	isFilterFound(cell, filter) {
-		const filterValue = filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value
-		const cellLabel = this.getLabel(cell.value)
+		const filterValue = (filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value).toLowerCase()
+		const cellLabel = this.getLabel(cell.value)?.toLowerCase()
 		const filterMethod = {
 			[FilterIds.ContainsItem]() { return filterValue.map(option => option.id).includes(cell.value) },
```
enjeck, line 57: `ContainsItem` crashes on Selection / SelectionMulti

1. Create a table with a single-selection column and add a few rows.
2. On that column open the filter menu and pick the operator "Is one of" / "Contains item" (the one that shows option chips, FilterIds.ContainsItem).
3. Select one option and apply.
4. Fails to create view

src/shared/components/ncTable/mixins/columnsTypes/selection.js
```diff
@@ -51,12 +51,12 @@ export default class SelectionColumn extends AbstractSelectionColumn {
...
-			[FilterIds.Contains]() { return cellLabel?.toLowerCase().includes(filterValue?.toLowerCase()) },
-			[FilterIds.DoesNotContain]() { return !cellLabel?.toLowerCase().includes(filterValue?.toLowerCase()) },
+			[FilterIds.Contains]() { return cellLabel.includes(filterValue) },
```
enjeck: To be safe, since `cellLabel` can be undefined
```suggestion
			[FilterIds.Contains]() { return cellLabel?.includes(filterValue) },
```

enjeck (changes requested):

src/shared/components/ncTable/mixins/columnsTypes/selection.js
Koc: aah, good catch 🦅 

Fixed now and thanx for clear reproducer.

src/shared/components/ncTable/mixins/columnsTypes/textLink.js
```diff
@@ -52,7 +52,7 @@ export default class TextLinkColumn extends AbstractTextColumn {
 	}
 
 	isFilterFound(cell, filter) {
-		const filterValue = filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value
+		const filterValue = (filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value).toLowerCase()
 		const value = this.getValueFromCellValue(cell.value)
```
enjeck: Do we need to lowercase  `value` too? Otherwise, a cell with uppercase characters e.g "Example" doesn't match a search for "example"

```suggestion
		const filterValue = (filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value).toLowerCase()
		const value = this.getValueFromCellValue(cell.value).toLowerCase()
```

src/shared/components/ncTable/mixins/columnsTypes/textLong.js
```diff
@@ -22,11 +22,12 @@ export default class TextLongColumn extends AbstractTextColumn {
 	}
 
 	isFilterFound(cell, filter) {
-		const filterValue = filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value
+		const filterValue = (filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value).toLowerCase()
+		const cellValue = cell.value.toLowerCase()
```
enjeck: `cell.value` can be null here:

```suggestion
		const cellValue = cell.value?.toLowerCase()
```

enjeck (changes requested):

src/shared/components/ncTable/mixins/columnsTypes/textLink.js
Koc: fixed

src/shared/components/ncTable/mixins/columnsTypes/textLong.js
Koc: fixed

enjeck (approved):
thank you so much! 🚀

Issue #727, opened 2023-11-27
deleting one option of a multiple options column rather hides itself and other entries plus maybe reuses their former "base" for replacements and supplements


wsauerlich:
### Steps to reproduce

1. Create a table which has the multiple selection type in any column
2. Add a row with multiple options in a cell which is part of that column
3. Delete one of the options which were predefined in step 1 by deleting it from the selectable options in the edit menu
4. Add a new option in the same edit menu mentioned in step 3

### Expected behavior

The option should be instantly and permanently deleted in every row where this option has been chosen and the options in the edit menu aswell. A row with one or more remaining options should still show these. And when adding (another) option a "whole new" option should be created.

### Actual behavior

The chosen option remains visible until one clicks somewhere else and then back on the table or view, however the adding of an option is instantly visible. (Regardless whether of looking at a view or the "main page" of the table; showing the data and the views.)

As soon as one option is removed from the edit menu of the multiple selection column all cells in the rows that had chosen this option can no longer be edited and the remaining options within the cells are no longer visible.

When another option will be added later at the same "position", it will automatically be re-added to the rows, that lost an option before. Even if the option got a completly new name. The cells become editable and visible again, but it can cause confusion, because every new option is treated as a replacement for the deleted one and not a supplement for the bunch of all the other options.

### Tables app version

0.6.4

### Browser

Firefox 120.0

### Operating system

Debian

### Web server

Apache

### PHP engine version

PHP 8.1

### Database

MariaDB

enjeck:
I can reproduce this. Thanks for explaining the problem so detailly @wsauerlich

enjeck:
Making this note here so I don't forget:

> When another option will be added later at the same "position", it will automatically be re-added to the rows, that lost an option before. Even if the option got a completly new name. The cells become editable and visible again, but it can cause confusion, because every new option is treated as a replacement for the deleted one and not a supplement for the bunch of all the other options.

This happens because whenever we add a new a `selectionOption`, we use the length of the existing array. The id for each option is gotten from [`getNextId()`](https://github.com/nextcloud/tables/blob/main/src/shared/components/ncTable/partials/columnTypePartials/forms/SelectionMultiForm.vue#L101), which is based on the length of the existing options. So the `id`s for options are not unique, since an id that was deleted earlier can be later reinstated. 

[...]

PR #2450, opened 2026-04-06
Fix: Properly display removed options in selection columns

Koc:
## Reproducer
1. Create a table with select and multi select columns
2. Create rows, select values
3. Remove some options

### Expected result
We should see yellow triangle with exclamation mark that informs us about non-existent optipns

## :framed_picture: Screenshoots
| :heavy_minus_sign: Before | :heavy_plus_sign: After |
|--------|--------|
| image | image |

### DIFF
--- a/src/shared/components/ncTable/mixins/columnsTypes/selection.js
+++ b/src/shared/components/ncTable/mixins/columnsTypes/selection.js
@@ -20,13 +20,12 @@ export default class SelectionColumn extends AbstractSelectionColumn {
 	}
 
 	getLabel(id) {
-		const i = this.selectionOptions?.findIndex((obj) => obj.id === id)
-		return this.selectionOptions[i]?.label
+		return this.getOptionObject(id)?.label
 	}
 
-	isDeletedLabel(value) {
-		const i = this.selectionOptions?.findIndex((obj) => obj.id === value)
-		return !!this.selectionOptions[i]?.deleted
+	getOptionObject(id) {
+		if (id === null || id === undefined) return null
+		return this.selectionOptions?.find((obj) => obj.id === id) || { id, label: String(id), deleted: true }
 	}
 
 	sort(mode, nextSorts) {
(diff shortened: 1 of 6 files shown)


enjeck:
Related to https://github.com/nextcloud/tables/issues/727

enjeck (commented):
If deleting an option is supposed to actually remove that value from rows, then the best UX in my opinion is to reflect that cleanup directly: remaining selections stay visible, deleted ones disappear. I don't like keeping a ghost value around.

src/shared/components/ncTable/partials/TableCellSelection.vue
```diff
@@ -5,7 +5,12 @@
 <template>
 	<div class="cell-selection">
 		<div v-if="!isEditing" class="non-edit-mode" @click="handleStartEditing">
-			{{ column.getLabel(value) }}<span v-if="isDeleted()" :title="t('tables', 'This option is outdated.')">&nbsp;⚠️</span>
+			<template v-if="selectedOption.deleted">
```
enjeck, line 8: this throws an error if you:
1. Open a table that has rows.
2. Add a new single-selection column with no default value (or just leave any existing single-select cell empty).
3. View the table in grid mode.
```
TypeError: Cannot read properties of null (reading 'deleted')
    at VueComponent.render50 (TableCellSelection.vue:158:1)
```

Easy fix could be:
```suggestion
			<template v-if="selectedOption?.deleted">
```

src/shared/components/ncTable/mixins/columnsTypes/selection.js
```diff
@@ -20,13 +20,12 @@ export default class SelectionColumn extends AbstractSelectionColumn {
 	}
 
 	getLabel(id) {
-		const i = this.selectionOptions?.findIndex((obj) => obj.id === id)
-		return this.selectionOptions[i]?.label
+		return this.getOptionObject(id)?.label
```
enjeck, line 23: `getLabel` now returns `String(id)` instead of `undefined` for orphaned values, which changes filter behavior in `isFilterFound`. i suppose that for example, `IsEmpty` there will now treat a row holding a deleted option as non-empty? Same for Contains, BeginsWith etc
Is this intentional behavior?

enjeck (commented):
some comments:

PR #2452, opened 2026-04-06
Update URL validation for text-link column

luka-nextcloud:


enjeck (dismissed):
Some links like `javascript:https://openai.com`  are accepted. 

Also, in the frontend, we have `isValidUrlProtocol()` for validating the url, and it would be nice if we have similar logic for the backend too

blizzz (approved):

PR #2453, opened 2026-04-08
docs: Changed the occurrences of 'circle' to 'team' for translators (fixes #2249)

### LINKED ISSUE #2249: Change all occurrences of "circle" to "team"
Same for "circles" -> "teams"

Translators at Transifex asked for it.

abbasnadir:
Replace the occurrence of circle with team as requested by the issue.

### DIFF
--- a/src/shared/components/ncTable/partials/rowTypePartials/UsergroupForm.vue
+++ b/src/shared/components/ncTable/partials/rowTypePartials/UsergroupForm.vue
@@ -47,23 +47,23 @@ export default {
 		return {
 			selectUsers: this.column.usergroupSelectUsers,
 			selectGroups: this.column.usergroupSelectGroups,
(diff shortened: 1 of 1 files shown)


enjeck (changes requested):
Thanks for your contribution. Nextcloud’s translation system is powered by [Transifex](https://www.transifex.com/nextcloud/). To start translating sign up and enter a group. 
We cannot merge those changes directly because they would be reverted by the state of transifex on the next sync, so please contribute those to the translation system directly.

https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#adding-translations

For this case however, we would need to update https://github.com/nextcloud/tables/blob/main/lib/Helper/CircleHelper.php#L65 instead, and these en_GB files would be updated from there.

PR #2454, opened 2026-04-11
Direct leakage of backend exception details in general index endpoint

tomaioo:
## Summary

Security: Direct leakage of backend exception details in general index endpoint

## Problem

**Severity**: `Medium` | **File**: `lib/Controller/ApiGeneralController.php:L62`

The `index()` method catches `InternalError|Exception` and returns `$e->getMessage()` in the response body. This duplicates information-disclosure behavior and may reveal backend internals to authenticated users (or broader audiences depending on route exposure).

## Solution

Replace returned exception text with a generic message and a stable error code. Keep detailed error context in logs only. Consider reusing a centralized sanitizer in the error-handling layer to prevent regressions.

## Changes

- `lib/Controller/ApiGeneralController.php` (modified)

### DIFF
--- a/lib/Controller/ApiGeneralController.php
+++ b/lib/Controller/ApiGeneralController.php
@@ -47,7 +47,7 @@ class ApiGeneralController extends AOCSController {
 	 *
 	 * Tables and views incl. shares
 	 *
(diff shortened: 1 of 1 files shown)


enjeck:
Please see https://github.com/nextcloud/server/security/policy

PR #2457, opened 2026-04-11
fix: Auto-focus input field for selecting a value via the column header

AndyScherzinger:
Before this change when clicking on "Select value" in the column header menu the opened dialog didn't auto-focus on the input field for direct typing but forces users to do an extra mouse click or keyboard tabbing.

This way the field is automatically focused after launching / rendering the dialog and field.


enjeck (approved):
nice

Issue #2125, opened 2025-10-28
Make archiving tables/views a personal option


juliusknorr:
With growing number of tables it may become useful if archiving would be a personal flag that one could set on a table or view so you can hide tables that you are no longer interested in as a recipient. 

This would be in line with how Talk implemented archiving. 

We could still consider the table owner being able to archive for everyone on top.

marcoambrosini:
Thanks for your input @kizm0!
Indeed, in addition to archiving, it would be nice if shared tables were under their own collapsible group, what do you think @juliusknorr ?

juliusknorr:
I don't think sharing is a good categorization there, an incoming shared table may be equally important as a personal one. But might be my personal opinionated point of view

AndyScherzinger:
I'll be giving it a try, let's see 🤞

kizm0:
just for information in [collectives 4.6.0](https://github.com/nextcloud/collectives/releases/tag/v4.6.0) there was a nice feature added to choose between collectives and their included pages, this could be also a nice feature for tables to choose between tables and their views are listed as children.

PR #2476, opened 2026-04-12
🗃️ Make archiving tables/applications a personal option

AndyScherzinger:
* Resolves partially #2125

### 🖼️ Screenshots

Archived categories | Archive app | Archive table | Unarchive app | Unarchive table
---|---|---|---|---
2026-04-12 15_33_25-Tables - Nextcloud — Mozilla Firefox|2026-04-12 15_34_51-Tables - Nextcloud — Mozilla Firefox|2026-04-12 15_34_35-Tables - Nextcloud — Mozilla Firefox|2026-04-12 15_34_05-Tables - Nextcloud — Mozilla Firefox|2026-04-12 15_33_50-Tables - Nextcloud — Mozilla Firefox

The original ticket #2125 defined tables and views, while this PR tacked tables and apps. Given a clear view on how archived views should get displayed this could be added as a follow-up PR to not make this PR grow in review-size and complexity.

---

## Summary

* This change replaces the existing single-user global `archived` flag on tables with a two-layer per-user model: an owner-level flag on the entity and a new `tables_archive_user` table that stores individual user overrides, so archive state is resolved per-user at read time. When an owner archives a table or application, it is automatically treated as archived for every user with access, but any of those users may override this for themselves; an owner unarchiving resets the state for everyone by clearing all per-user overrides. 
* The existing `tables_tables.archived` column is repurposed from a simple global toggle into the owner-level signal, and the database migration converts all currently-archived tables into owner-archived entries to preserve existing data without disruption. 
* Application (context) archiving is introduced as a completely new capability end-to-end: covering database schema, entity, service, API endpoints, routes, and frontend navigation - while table archiving is substantially reworked from a simple owner toggle into the full per-user propagation system. 
* A new `ArchiveService` centralizes all business logic for both resource types, four new API endpoints are split across `ApiTablesController` and `ContextController` following existing naming conventions, and the frontend gains archive/unarchive actions in both navigation item menus alongside a new collapsible "Archived applications" section mirroring the already-existing "Archived tables" section.

### ⚠️ Beware

The database migration only covers the owner and direct user shares, not groups, etc. since they can't be resolved directly via SQL (think teams/groups). So in these cases archived tables would resurface until manually archived by the respective user. Also there is no cleanup for teams/groups, if a user is taken out of a team, the archive meta-info would not get deleted. Only simple but confusing way around it would be to prevent archiving/unarchiving for tables/apps where a user is neither the owner nor a user-share receiver. Expensive alternative would be a housekeeping job checking all these archive bits periodically, but that also sounds expensive with little use. Else we would need to consume a team/group changes event and check if some cleanup is needed. WDYT?

### 🏁 Checklist
 
- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


jancborchardt (dismissed):
Looks generally nice, 2 details
- [x] I’d say to use outline icons instead of filled for both cases, archiving and unarchiving
- [x] In the cases of "Archive app" and "Unarchive app", the sorting of actions in the menu is off. We always recommend the destructive action (Delete) to be the very last, and then the next most destructive like Archive could be directly above it.

lib/Db/UserArchive.php
```diff
@@ -0,0 +1,35 @@
...
+	public function __construct() {
+		$this->addType('id', 'integer');
+		$this->addType('node_type', 'integer');
+		$this->addType('node_id', 'integer');
+		$this->addType('archived', 'boolean');
+	}
+
+	public function isArchived(): bool {
+		return $this->archived;
+	}
```
ChristophWurst: Should not be necessary. Boolean getters can also be used as `isXXX` and not just `getXXX`

Ref https://github.com/nextcloud-deps/ocp/blob/85a2b59a1988278324e0b41bc78ad64e0d1f627e/OCP/AppFramework/Db/Entity.php#L177-L196

lib/Db/UserArchiveMapper.php
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		$results = [];
+		foreach (array_chunk($nodeIds, 997) as $chunk) {
+			$qb = $this->db->getQueryBuilder();
+			$qb->select('*')
+				->from($this->table)
+				->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId, IQueryBuilder::PARAM_STR)))
+				->andWhere($qb->expr()->eq('node_type', $qb->createNamedParameter($nodeType, IQueryBuilder::PARAM_INT)))
+				->andWhere($qb->expr()->in('node_id', $qb->createNamedParameter($chunk, IQueryBuilder::PARAM_INT_ARRAY)));
```
ChristophWurst: perf: create the query outside the loop with a parameter. Set the parameter inside the loop. That prevents rebuilding the query representation if this really loops a lot.

AndyScherzinger: DONE via https://github.com/nextcloud/tables/pull/2476/commits/cce6242bb4729c34677d5e55927b0b54a9a2e8aa

lib/Db/UserArchive.php
AndyScherzinger: Done via https://github.com/nextcloud/tables/pull/2476/commits/298dbca7a447dc55d57911078f5c3a4b637048eb

samin-z:
hey @AndyScherzinger , currently going through the pr but one hing to mention, NODE_TYPE_CONTEXT is 2 (number), the middleware only whitelisted (0,1) for numeric types, so the "$nodeType === 'context'" inside "if (!is_numeric...)" is skipped, so as a result when i call the api to archive a context i get 'Invalid node type', an extra elseif condition that if $nodeType === Application::NODE_TYPE_CONTEXT then set $isContext = true can fix it.

lib/Db/UserArchiveMapper.php
```diff
@@ -0,0 +1,142 @@
...
+		return $entities[0] ?? null;
+	}
+
+	/**
+	 * Fetch all per-user archive overrides for a given node.
+	 *
+	 * @return UserArchive[]
+	 * @throws Exception
+	 */
+	public function findAllForNode(int $nodeType, int $nodeId): array {
```
samin-z: this function is unused outside tests, you can either remove for YAGNI, or better add a comment that it is intentionally kept for future diagnostics (better to be removed if there is no need for it), no strong security issue today since nothing calls it.

lib/Db/UserArchiveMapper.php
```diff
@@ -0,0 +1,142 @@
...
+	 *
+	 * @param int[] $nodeIds IDs to restrict the lookup to
+	 * @return array<int, UserArchive> Keyed by node_id for O(1) map lookup
+	 * @throws Exception
+	 */
+	public function findAllOverridesForUser(string $userId, int $nodeType, array $nodeIds): array {
+		if (empty($nodeIds)) {
+			return [];
+		}
+
```
samin-z, line 63: we can also check for duplicates with adding this line as precaution
$nodeIds = array_values(array_unique($nodeIds));

lib/Db/UserArchiveMapper.php
```diff
@@ -0,0 +1,142 @@
...
+		$results = [];
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->table)
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId, IQueryBuilder::PARAM_STR)))
+			->andWhere($qb->expr()->eq('node_type', $qb->createNamedParameter($nodeType, IQueryBuilder::PARAM_INT)))
+			->andWhere($qb->expr()->in('node_id', $qb->createParameter('chunk')));
+
+		foreach (array_chunk($nodeIds, 997) as $chunk) {
```
samin-z: suggestion: as a maintainability improvement we can have the limit (997) in a private const (with a short explanation comment).

lib/Db/UserArchiveMapper.php
```diff
@@ -0,0 +1,142 @@
...
+		return $results;
+	}
+
+	/**
+	 * Insert or update a per-user archive override.
+	 *
+	 * @throws Exception
+	 */
+	public function upsert(string $userId, int $nodeType, int $nodeId, bool $archived): void {
+		$existing = $this->findForUser($userId, $nodeType, $nodeId);
```
samin-z: we can have an improvement here: instead of 2 DB calls (read-modify or write), better to have 'INSERT... ON DUPLICATE KEY UPDATE'.

lib/Migration/Version2400Date20260904000000.php
```diff
@@ -0,0 +1,190 @@
...
+		$result->closeCursor();
+
+		if (empty($archivedTables)) {
+			$output->info('No archived tables found, skipping data migration.');
+			return;
+		}
+
+		$inserted = 0;
+
+		foreach ($archivedTables as $tableRow) {
```
samin-z, line 157: insetad of loop per table i would suggest a bulk 'INSERT … SELECT'.

lib/Migration/Version2020Date20260428000000.php
```diff
@@ -0,0 +1,190 @@
...
+
+			// Insert owner record
+			$inserted += $this->upsertArchiveRecord($ownerId, 0, $tableId, true);
+
+			// Insert direct user-share recipient records
+			$shareQb = $this->connection->getQueryBuilder();
+			$shareQb->select('receiver')
+				->from('tables_shares')
+				->where($shareQb->expr()->eq('node_id', $shareQb->createNamedParameter($tableId, IQueryBuilder::PARAM_INT)))
+				->andWhere($shareQb->expr()->eq('node_type', $shareQb->createNamedParameter(0, IQueryBuilder::PARAM_INT)))
```
samin-z: better to use 'Application::NODE_TYPE_TABLE' for hard coded '0'.

lib/Service/ArchiveService.php
```diff
@@ -0,0 +1,213 @@
...
+	public function enrichContextsWithArchiveState(array $contexts, string $userId): array {
+		$nodeIds = array_map(fn (Context $c) => $c->getId(), $contexts);
+		$overrides = $this->userArchiveMapper->findAllOverridesForUser($userId, Application::NODE_TYPE_CONTEXT, $nodeIds);
+
+		foreach ($contexts as $context) {
+			$override = $overrides[$context->getId()] ?? null;
+			$archived = $override !== null ? $override->isArchived() : $context->isArchived();
+			$context->setArchived($archived);
+		}
+
```
samin-z, line 133: not nit-picking or blocker just a suggestion,the “load overrides and foreach merge” block in enrichTablesWithArchiveState and enrichContextsWithArchiveState is duplicated, we can extract a private helper that takes $nodeType, $userId, and an iterable of entities and call it from both methods.

lib/Service/ArchiveService.php
```diff
@@ -0,0 +1,213 @@
...
+	 * Must be called inside the same atomic block as the ownership-column
+	 * update so that archive state and ownership are always consistent.
+	 *
+	 * Returns the new entity-level archived value. The caller must call
+	 * `$entity->setArchived($newArchived)` when the returned value differs
+	 * from $entityArchived, before persisting the entity.
+	 *
+	 * @throws Exception
+	 */
+	public function prepareOwnershipTransfer(
```
samin-z, line 157: there is an edge case for this functionality, after ownership transfer this function may upsert a preservation override for the outgoing owner even if they lose access to the table (no row in tables_shares). the row added to atchive_user then has no practical effect in the UI. we could consider skipping preservation when the outgoing owner does not have read access, or a also add outgoing owner to the tables_shares, not a blocker but otherwise this is harmless dead data.

lib/Service/ContextService.php
```diff
@@ -117,7 +126,55 @@ public function findById(int $id, ?string $userId): Context {
...
+	 *
+	 * @throws Exception
+	 * @throws NotFoundError
+	 * @throws InternalError
+	 */
+	public function archiveContext(int $contextId, string $userId): Context {
+		// Load directly from mapper to get entity-level archived (bypasses per-user enrichment)
+		$context = $this->contextMapper->findById($contextId, $userId);
+		$isOwner = $context->getOwnerId() === $userId;
+		$this->archiveService->archiveForUser($userId, Application::NODE_TYPE_CONTEXT, $contextId, $isOwner);
```
samin-z: in TableService.php it wraps archiveForUser in try/catch for OCP\DB\Exception, but archiveContext / unarchiveContext do not, aligning would keep behaviour and API errors consistent between tables and contexts.

lib/Service/ContextService.php
```diff
@@ -117,7 +126,55 @@ public function findById(int $id, ?string $userId): Context {
...
+		return $this->findById($contextId, $userId);
+	}
```
samin-z, line 179: findById is a heavy query, running it twice here is fine in regard to correctness, but it does cost more DB work. for later we can have a lighter reload after the write, or a short note in the PR that this tradeoff is intentional.

samin-z:
@AndyScherzinger i might be nit picking at times ^^'.
some are not blockers and just suggestions.

jancborchardt (approved):
Design-wise 👍

AndyScherzinger:
> @AndyScherzinger i might be nit picking at times ^^'.

@samin-z Please never worry about nit picking, that is perfectly fine 👍 Happy to fix and improve whatever you find and I am able to fix 🎉

enjeck:
@AndyScherzinger I fixed merge conflicts

lib/Controller/ContextController.php
```diff
@@ -255,6 +256,62 @@ public function transfer(int $contextId, string $newOwnerId, int $newOwnerType =
...
+	 * @param int $contextId ID of the context
+	 * @return DataResponse<Http::STATUS_OK, TablesContext, array{}>|DataResponse<Http::STATUS_FORBIDDEN|Http::STATUS_INTERNAL_SERVER_ERROR|Http::STATUS_NOT_FOUND, array{message: string}, array{}>
+	 *
+	 * 200: Context returned with updated archived state
+	 * 403: No permissions
+	 * 404: Context not found or not available
+	 */
+	#[NoAdminRequired]
+	#[UserRateLimit(limit: 20, period: 60)]
+	#[RequirePermission(permission: Application::PERMISSION_READ, type: Application::NODE_TYPE_CONTEXT, idParam: 'contextId')]
```
enjeck: does this actually work? This attribute passes type: `Application::NODE_TYPE_CONTEXT` (integer 2). `PermissionMiddleware::checkPermission()` only recognizes a context when the type arrives as the string `'context'`.

lib/Controller/ContextController.php
```diff
@@ -255,6 +256,62 @@ public function transfer(int $contextId, string $newOwnerId, int $newOwnerType =
...
```
enjeck: same as at https://github.com/nextcloud/tables/pull/2476/changes#r3887748276

tests/unit/Controller/ContextControllerArchiveTest.php
```diff
@@ -0,0 +1,129 @@
...
+		$n = $this->createMock(IL10N::class);
+		$n->method('t')->willReturnArgument(0);
+
+		$this->controller = new ContextController(
+			$this->createMock(IRequest::class),
+			$this->createMock(LoggerInterface::class),
+			$n,
+			'alice',
+			$this->contextService,
+		);
```
enjeck, line 50: called with 5 arguments (request, logger, n, userId, contextService). The real constructor requires 9

tests/unit/Controller/ApiTablesControllerArchiveTest.php
```diff
@@ -0,0 +1,155 @@
...
+			$this->createMock(IRequest::class),
+			$this->createMock(LoggerInterface::class),
+			$this->tableService,
+			$this->createMock(ColumnService::class),
+			$this->createMock(ViewService::class),
+			$n,
+			$this->createMock(IAppManager::class),
+			$this->createMock(IDBConnection::class),
+			'alice',
+		);
```
enjeck, line 53: The constructor call passes 9 arguments and omits `StructureService`. The real ApiTablesController constructor takes 10

lib/Controller/ApiTablesController.php
enjeck: This calls `$this->service->find($id)` (and does not pass a userId), so it returns the entity level archived flag, not the per-user resolved value. I think `getTableForUser($id, $this->userId)` works better?

enjeck: same in lib/Controller/TableController.php:41

lib/Service/TableService.php
```diff
@@ -157,6 +158,14 @@ public function findAll(?string $userId = null, bool $skipTableEnhancement = fal
 			}
 		}
 
+		if ($userId !== '') {
+			try {
+				$this->archiveService->enrichTablesWithArchiveState(array_values($allTables), $userId);
+			} catch (OcpDbException $e) {
+				$this->logger->error($e->getMessage(), ['exception' => $e]);
+			}
+		}
```
enjeck, line 167: what happens if `$userId` is null?

lib/Migration/Version2400Date20260904000000.php
enjeck, line 1: it's been so long, best to rename this file too

src/store/store.js
```diff
@@ -346,6 +346,70 @@ export const useTablesStore = defineStore('store', {
...
+			try {
+				await axios.post(generateOcsUrl(`/apps/tables/api/2/tables/${id}/archive`))
+			} catch (e) {
+				displayError(e, t('tables', 'Could not archive table.'))
+				return false
+			}
+
+			const index = this.tables.findIndex(t => t.id === id)
+			const table = this.tables[index]
+			table.archived = true
```
enjeck, line 362: this could throw an error if `table` is `undefined` (const `table = this.tables[index]` after `findIndex` returns -1 gives `undefined`)

AndyScherzinger:
Thanks @enjeck

Beware that while it generally works for items you have direct access to, there is a lack of houses keeping of you losses access due to a circle or group change, this would need addition logic to removed archive recordings when a user is thrown out of a circle or a circle of is removed in any way.

src/modules/navigation/partials/NavigationTableItem.vue
```diff
@@ -109,7 +109,7 @@
 			</NcActionButton>
 
 			<!-- ARCHIVE -->
-			<NcActionButton v-if="canManageElement(table) && !table.archived && !table.favorite"
```
enjeck, line 112: Removing `canManageElement(table)` means read-only shared users now see Archive and Unarchive? Is this intentional?

lib/Service/ContextService.php
enjeck, line 108: Do we want archived applications to show in the global nav bar?

src/store/store.js
```diff
@@ -125,7 +125,7 @@ export const useTablesStore = defineStore('store', {
 		},
 		setContext(context) {
 			const index = this.contexts.findIndex(c => c.id === context.id)
-			this.contexts[index] = context
+			this.contexts.splice(index, 1, context)
```
enjeck: This swap to splice looks like it was meant to fix reactivity??  But on Vue 3 `this.contexts[index] = context` already updates the view, so it is not needed.

enjeck (changes requested):

src/modules/navigation/partials/NavigationTableItem.vue
AndyScherzinger: I'd say so, yes. The gola of the PR is to have an archiving/unarchiving option on user-level, so also a ready-only share reciever should be able to (un)archive for them since this is about the meta-data belonging to the sharee, the table/view is still read-ony

lib/Controller/ContextController.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/95c6667724ff104de62b8fe594596098a26d137c
`fix(archiving): pass context type as typeParam in RequirePermission`

lib/Controller/ContextController.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/95c6667724ff104de62b8fe594596098a26d137c
`fix(archiving): pass context type as typeParam in RequirePermission`

lib/Controller/ApiTablesController.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/fb90274c4ac8d3ab0465f74602daee0f7cef3d18
`fix(archiving): resolve per-user archived state in table show endpoints`

lib/Service/ContextService.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/1f7777936bd4e0e314552292c7e9a61c90c234c6
`fix(archiving): hide archived applications from the app navigation`

lib/Migration/Version2400Date20260904000000.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/803118c9ef47a61fffb6044f716a56e6cb6ecebf
`chore(archiving): rename migration to match the upcoming app version`

tests/unit/Controller/ContextControllerArchiveTest.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/5d65a9cdeddab0339b3e1a877eb9664804a6839f
`fix(archiving): pass all constructor dependencies in ContextController archive test`

tests/unit/Controller/ApiTablesControllerArchiveTest.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/3382d1960226503f8a8d321bb5089870290966d8
`fix(archiving): add missing StructureService mock in ApiTablesController archive test`

src/store/store.js
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/7217133f62dcc6730f86b1feb3cc886d0f565da1
`fix(archiving): guard store archive actions against missing items`

src/store/store.js
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/4cb79368449d9366216498341e1a4eae8dc69251
`fix(store): keep direct index assignment in setContext`

lib/Db/UserArchiveMapper.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/8a36c987708f2bfc017794c8b642b8f11843411c
`refactor(archiving): remove unused findAllForNode mapper method`

lib/Db/UserArchiveMapper.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/d90e8885c1db274fc354cabae01150e42fc52e51
`refactor(archiving): extract IN-clause chunk size into a named constant`

lib/Db/UserArchiveMapper.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/97f5805a2058b6e9039864e8671c9c19a56f11b1
`perf(archiving): upsert archive overrides atomically via setValues`

lib/Migration/Version2400Date20260904000000.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/8da798f0daabcc2b6dd67f3e56b28a0e1fa01d85
`perf(archiving): build migration queries once outside the loop`

lib/Migration/Version2020Date20260428000000.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/53c90805a03fcd53bf6f521e3489593dc2097ccd
`refactor(archiving): use NODE_TYPE_TABLE constant in migration`

lib/Service/ArchiveService.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/56b97993d99a8bfa48ef1239dabdb85fbd8b797a
`refactor(archiving): deduplicate per-user archive state enrichment`

lib/Controller/ApiTablesController.php
AndyScherzinger: TableController.php:41 is covered by the same commit https://github.com/nextcloud/tables/commit/fb90274c4ac8d3ab0465f74602daee0f7cef3d18

lib/Db/UserArchiveMapper.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/2c1622b93398ed46e6b93c6314a0db2e5f790307
`fix(archiving): deduplicate node IDs before chunked override lookup`

lib/Service/ContextService.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/207bdf6b099dc35b3c7548c0679445c39976076f
`fix(archiving): align context archive error handling with tables`

lib/Service/ArchiveService.php
AndyScherzinger: Fixed in https://github.com/nextcloud/tables/commit/63fb0e3ab2903f64299aecf3d670306a94ea2b6f
`fix(archiving): drop outgoing owner's override when transfer removes access`

lib/Service/TableService.php
AndyScherzinger: $userId can't be null at this point. findAll() normalizes it right at the top via preCheckUserId():
https://github.com/nextcloud/tables/blob/db76d801961063b4cebf89e17018553bcae1e983/lib/Service/TableService.php#L94

preCheckUserId() has a string return type — a null argument is replaced by the session user, and if there is no user at all it throws an InternalError; the only special value it can return is '' for CLI/occ context:
https://github.com/nextcloud/tables/blob/db76d801961063b4cebf89e17018553bcae1e983/lib/Service/PermissionsService.php#L66-L83

That '' case is exactly what the guard here skips, because per-user archive enrichment is meaningless without a user:
https://github.com/nextcloud/tables/blob/db76d801961063b4cebf89e17018553bcae1e983/lib/Service/TableService.php#L161

So no code path reaches the enrich call with null.

lib/Service/ContextService.php
AndyScherzinger: This is an intentional tradeoff. The first read fetches the entity-level archived flag before the write (bypassing per-user enrichment on purpose), and the second read returns the authoritative post-mutation state — including the freshly resolved per-user archive value — as the API response body:
https://github.com/nextcloud/tables/blob/db76d801961063b4cebf89e17018553bcae1e983/lib/Service/ContextService.php#L168-L178

It mirrors the table side, where archiveTable()/unarchiveTable() also re-read after the write via getTableForUser():
https://github.com/nextcloud/tables/blob/db76d801961063b4cebf89e17018553bcae1e983/lib/Service/TableService.php#L319-L328

Archiving is a rare interaction, so the extra query cost is negligible; a lighter reload could be a follow-up optimization if it ever shows up in profiling 👍

AndyScherzinger:
@samin-z regarding https://github.com/nextcloud/tables/pull/2476#issuecomment-4430562621

Fixed in https://github.com/nextcloud/tables/commit/6f4140f65a3a664d8a76236408a4e69be0d3d9b3
fix(archiving): pass context type as typeParam in RequirePermission

Instead of extending the middleware's numeric whitelist as suggested, the two endpoints now pass typeParam: 'context' like every other context endpoint in the controller, so the existing string-based context detection in PermissionMiddleware applies.

lib/Service/TableService.php
```diff
@@ -536,7 +713,7 @@ public function update(int $id, ?string $title, ?string $emoji, ?string $descrip
 			$table->setEmoji($emoji);
 		}
 		if ($archived !== null) {
-			$table->setArchived($archived);
+			$this->applyArchivedOnUpdate($table, $archived, $userId);
```
enjeck: should we gate this of `$isOwner && $entityArchived !== $archived`? Otherwise, one user's private archive shows up as a shared change that everyone sees?

lib/Service/TableService.php
enjeck: do we want to update this even when the only change was a non-owner's private archive setting?

tests/unit/Controller/ApiTablesControllerArchiveTest.php
enjeck, line 1: we should probably have a test for applyArchivedOnUpdate() for owners and non-owners

lib/Service/TableService.php
AndyScherzinger: Fixed in feat(archive): apply per-user archive state in table and context services (81e20476).

lib/Service/TableService.php
AndyScherzinger: True,  we don't want that...

Fixed in feat(archive): apply per-user archive state in table and context services (81e20476).

tests/unit/Controller/ApiTablesControllerArchiveTest.php
AndyScherzinger: Added in test(archive): add unit and e2e tests for archiving (794e3e82)

PR #2480, opened 2026-04-14
fix(datetime): parse stored values as UTC in edit form getters

### LINKED ISSUE #2479: Datetime column shifts time by timezone offset on every save (DST-aware, Tables 2.0.x)
### Steps to reproduce

1. Set up Nextcloud 33 with Tables 2.0.1
2. User timezone: Europe/Zurich (CEST, UTC+2 in summer / CET, UTC+1 in winter)
3. Server timezone: Europe/Zurich, PHP: Europe/Zurich, DB: SYSTEM (= Europe/Zurich)
4. Create a table with a datetime column
5. Enter a time (e.g. 8:00) and save
6. Open the row for editing again (modal or inline)

### Expected behavior

The saved time (8:00) is shown correctly in the edit form.

### Actual behavior

Every time you open the edit form and save, the time is shifted by the 
current UTC offset:
- In summer (CEST, UTC+2): shifts by -2h each save
- In winter (CET, UTC+1): shifts by -1h each save

This means the shift is DST-aware — it uses the local timezone offset 
of the *date being edited*, not a fixed offset. Dates created before 
the March DST switch shift by -1h, dates after shift by -2h.

### Tables app version

2.0.1

### Browser

Firefox 149, Chrome 146

### Client operating system

Windows

### Operating system

Ubuntu 24.04.4 LTS (up to date)

### Web server

Apache

### PHP engine version

PHP 8.3

### Database

MariaDB

### Additional info

The bug does NOT occur on:
- Nextcloud 32.0.6 with Tables 1.0.5 (same user timezone, same server config)

This suggests a regression introduced in Tables 2.0.x, likely related to 
the new inline datetime editing feature introduced in this version.

The time is shifted correctly for display (DST-aware), but the edit form 
appears to interpret the stored naive datetime string as local time and 
converts it to UTC on save — effectively subtracting the UTC offset each 
time the row is saved.

samuelx12:
## What this fixes
Every time a row with a datetime column is opened for editing and saved,
the time shifts by the current UTC offset (−2h in summer CEST, −1h in
winter CET). The shift is DST-aware — dates created before the March DST
switch shift by −1h, dates after by −2h.

## Root cause
Commit 78f8131a moved datetime handling from a backend `TimezoneHelper`
to frontend-side UTC conversion. The display formatters and form setters
were correctly updated to use `.utc()` / `.utc().local()`, but the form
getters (which load the stored value into the date picker) were missed.
They still parsed the stored UTC string as local time, causing an offset
drift on every open→save cycle.

## Fix
Update the three affected getters to use `Moment.utc(...).local().toDate()`
— matching the pattern already used by the setters and display formatters.

**Files changed:**
- `DatetimeForm.vue` — modal edit form getter
- `DatetimeTimeForm.vue` — time-only modal edit form getter
- `TableCellDateTime.vue` — inline edit getter (both datetime and time-only)

`DatetimeDateForm.vue` (date-only) is not affected — its setter never had
`.utc()` added, so getter and setter were already symmetric.

## Testing
Manually tested with Docker (Nextcloud 33, Europe/Zurich timezone):
- Datetime values remain stable across multiple open→save cycles
- Both modal edit and inline edit verified
- Summer (CEST, UTC+2) and winter (CET, UTC+1) dates tested

No unit tests added — the affected components (Vue form getters) are not
covered by the existing test suite. A Playwright E2E test for datetime
round-trip stability would be a useful follow-up.

Fixes #2479

### DIFF
--- a/src/shared/components/ncTable/partials/rowTypePartials/DatetimeForm.vue
+++ b/src/shared/components/ncTable/partials/rowTypePartials/DatetimeForm.vue
@@ -42,7 +42,7 @@ export default {
 		localValue: {
 			get() {
 				if (this.value !== null && this.value !== 'none') {
-					return Moment(this.value, 'YYYY-MM-DD HH:mm').toDate()
+					return Moment.utc(this.value, 'YYYY-MM-DD HH:mm').local().toDate()
 				} else if (this.value === null && this.column.datetimeDefault === 'now') {
 					const dt = Moment()
 					this.$emit('update:value', dt.format('YYYY-MM-DD HH:mm'))
--- a/src/shared/components/ncTable/partials/rowTypePartials/DatetimeForm.vue
+++ b/src/shared/components/ncTable/partials/rowTypePartials/DatetimeForm.vue
@@ -45,7 +45,7 @@ export default {
 					return Moment.utc(this.value, 'YYYY-MM-DD HH:mm').local().toDate()
 				} else if (this.value === null && this.column.datetimeDefault === 'now') {
 					const dt = Moment()
-					this.$emit('update:value', dt.format('YYYY-MM-DD HH:mm'))
+					this.$emit('update:value', dt.clone().utc().format('YYYY-MM-DD HH:mm'))
 					return dt.toDate()
 				} else {
 					return null
--- a/src/shared/components/ncTable/partials/rowTypePartials/DatetimeTimeForm.vue
+++ b/src/shared/components/ncTable/partials/rowTypePartials/DatetimeTimeForm.vue
@@ -50,7 +50,7 @@ export default {
 		localValue: {
 			get() {
 				if (this.value !== null && this.value !== 'none') {
-					return Moment(this.value, 'HH:mm').toDate()
+					return Moment.utc(this.value, 'HH:mm').local().toDate()
 				} else if (this.value === null && this.column.datetimeDefault === 'now') {
 					const dt = Moment()
 					this.$emit('update:value', dt.format('HH:mm'))
--- a/src/shared/components/ncTable/partials/rowTypePartials/DatetimeTimeForm.vue
+++ b/src/shared/components/ncTable/partials/rowTypePartials/DatetimeTimeForm.vue
@@ -53,7 +53,7 @@ export default {
 					return Moment.utc(this.value, 'HH:mm').local().toDate()
 				} else if (this.value === null && this.column.datetimeDefault === 'now') {
 					const dt = Moment()
-					this.$emit('update:value', dt.format('HH:mm'))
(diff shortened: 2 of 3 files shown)


src/shared/components/ncTable/partials/rowTypePartials/DatetimeTimeForm.vue
```diff
@@ -50,7 +50,7 @@ export default {
 		localValue: {
 			get() {
 				if (this.value !== null && this.value !== 'none') {
-					return Moment(this.value, 'HH:mm').toDate()
+					return Moment.utc(this.value, 'HH:mm').local().toDate()
 				} else if (this.value === null && this.column.datetimeDefault === 'now') {
 					const dt = Moment()
 					this.$emit('update:value', dt.format('HH:mm'))
```
enjeck: Do we need to update this too? 
```suggestion
					this.$emit('update:value', dt.clone().utc().format('HH:mm'))
```

src/shared/components/ncTable/partials/rowTypePartials/DatetimeForm.vue
```diff
@@ -42,7 +42,7 @@ export default {
 		localValue: {
 			get() {
 				if (this.value !== null && this.value !== 'none') {
-					return Moment(this.value, 'YYYY-MM-DD HH:mm').toDate()
+					return Moment.utc(this.value, 'YYYY-MM-DD HH:mm').local().toDate()
 				} else if (this.value === null && this.column.datetimeDefault === 'now') {
 					const dt = Moment()
 					this.$emit('update:value', dt.format('YYYY-MM-DD HH:mm'))
```
enjeck: do we need to update this too?

```suggestion
					this.$emit('update:value', dt.clone().utc().format('YYYY-MM-DD HH:mm'))
```

src/shared/components/ncTable/partials/rowTypePartials/DatetimeTimeForm.vue
samuelx12: Yes, thanks!! Didn't think about the creation with default time "now".

samuelx12:
Not directly related to this PR, but a question came up to me:
There's no migration strategy for switching to Tables 2.x.x. People will see all their existing times shifted by their UTC offset (unless they live in England 😉). That's bad UX. Just throwing it out there, even though I don't have a good solution either.

enjeck:
i suppose we no longer need this after https://github.com/nextcloud/tables/pull/2583 ? Please reopen if I'm wrong

samuelx12:
> i suppose we no longer need this after #2583 ? Please reopen if I'm wrong

Thats correct. Sorry for not closing it myself.

PR #2481, opened 2026-04-14
🔢 Reorder and sort table columns

### LINKED ISSUE #607: An option to edit a default view

Right now it is impossible to reorder columns or show metadata columns on a table without creating a separate view.

Create a hidden view object for each table, which would be a default view for this table. Allow users to edit it by clicking Edit table in the menu. Renaming the default view would rename the table, making this solution a complete replacement for the current editing window, which allows renaming only.

Allow to save a view as a default view, itstead of replacing editing dialogue.

### Additional context

_No response_

AndyScherzinger:
Resolves #607 implementing the same mechanism views already offer, but for tables

- **Column Order:** Added `column_order` to `tables_tables` (DB migration), extended the `Table` entity, `TableService`, and controllers to persist and apply it, and exposed a drag-and-drop `ColumnOrderList.vue` in the "Edit table" modal.
- **Default Sort:** Added `sort` to `tables_tables`, wired it through the `Table` entity, `TableService`, and controllers, and added a `DefaultSortRules.vue` component to the modal.
- **Import/Export:** Both fields are preserved through scheme export/import and user-migration, with round-trip tests confirming correctness.
- **Testing:** Added / adapted unit up to e2e tests.
- **OpenAPI Spec:** Updated and regenerated openapi.json and its TypeScript counterpart.

Dialog old|Dialog new (2 sort items)|Dialog new (1 sort item)|Dialog view new (2 sort items)|
---|---|---|---
2026-04-16 14_13_08-Tabellen - Nextcloud — Mozilla Firefox|2026-04-16 14_21_35-Welcome to Nextcloud Tables! - Tables - Nextcloud — Mozilla Firefox|2026-04-16 14_10_12-Welcome to Nextcloud Tables! - Tables - Nextcloud — Mozilla Firefox|2026-04-16 14_22_00-Check yourself! - Tables - Nextcloud — Mozilla Firefox

**JSON Export**
See last section for order/sorting block (like for views)
```
{
   "title":"Welcome to Nextcloud Tables!",
   "emoji":"\ud83d\ude80",
   "columns":[
      ...
   ],
   "views":[
      ...
   ],
   "description":"A B C D",
   "tablesVersion":"2.0.1",
   "columnOrder":[
      {
         "columnId":4,
         "order":1,
         "readonly":false,
         "mandatory":false
      },
      {
         "columnId":2,
         "order":2,
         "readonly":false,
         "mandatory":false
      },
      {
         "columnId":1,
         "order":3,
         "readonly":false,
         "mandatory":false
      },
      {
         "columnId":3,
         "order":4,
         "readonly":false,
         "mandatory":false
      }
   ],
   "sort":[
      {
         "columnId":3,
         "mode":"DESC"
      }
   ]
}
```


jancborchardt (changes requested):
Looks very nice! 2 things:
- [x] Wording detail: Heading should be "Default sorting" instead of "Default sort"
- [ ] The default sorting can be deleted via the trash icon on the right. In Tables right now the default is new additions on the bottom, something we could also include there? That way it would be more obvious than being an "empty" state.

marcoambrosini (commented):
Nice one! 
One comment about the info text of default sorting: I wonder if there's a way to make this explanation a bit more clear, I don't understand it 😅 
Another minor improvement would be moving it to right after the heading instead of after the add rule button.

AndyScherzinger:
Implemented @jancborchardt first comment and @marcoambrosini's feedback

jancborchardt (approved):
Noice

marcoambrosini (approved):
Thanks for the addition :)

lib/UserMigration/TablesMigrator.php
```diff
@@ -199,6 +199,28 @@ public function import(
...
+					$this->tableMapper->update($newTable);
+				}
+				if (!empty($table['sort'])) {
+					$remapped = array_map(function (array $entry) use ($columnIdMap): array {
+						if (isset($entry['columnId'])) {
+							$entry['columnId'] = $columnIdMap[$entry['columnId']] ?? $entry['columnId'];
+						}
+						return $entry;
+					}, $table['sort']);
+					$newTable->setSort(json_encode($remapped));
```
samin-z, line 221: in line 211 and 220, we're having two BD writes, we can add something like '$updateSortOrder = true;' for each condition and then if true, we can have
if ($needsUpdate) {
$this->tableMapper->update($newTable);
}
outside the condition, this way we only made one call

lib/Service/RowService.php
```diff
@@ -97,7 +97,10 @@ public function findAllByTable(int $tableId, string $userId, ?int $limit = null,
 				$tableColumns = $this->columnMapper->findAllByTable($tableId);
 				$showColumnIds = array_map(fn (Column $column) => $column->getId(), $tableColumns);
 
-				return $this->row2Mapper->findAll($showColumnIds, $tableId, $limit, $offset, null, null, $userId);
+				$table = $this->tableMapper->find($tableId);
+				$sort = !empty($table->getSortArray()) ? $table->getSortArray() : null;
```
samin-z: here getSortArray() is called twice, we can simplify it with:
`$sort = $table->getSortArray() ?: null;`

src/modules/main/partials/editTablePartials/DefaultSortRules.vue
```diff
@@ -0,0 +1,91 @@
+<!--
+  - SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
```
samin-z: i think this should be 2026 :D

samin-z (changes requested):

src/modules/main/partials/editTablePartials/DefaultSortRules.vue
AndyScherzinger: Good point, let me update that. It is based on the sorting from the view implementation hence one could argue for 2023 as a date reference point, but might also be confusing. So let's stick to 2026 👍

lib/Service/RowService.php
AndyScherzinger: Right 🙈 Good catch

lib/UserMigration/TablesMigrator.php
AndyScherzinger: Looking into it 👍

samin-z (approved):

lib/Controller/ApiTablesController.php
```diff
@@ -130,12 +130,14 @@ public function showScheme(int $id): DataResponse {
...
 	 * @param list<TablesView> $views views
+	 * @param list<array{columnId: int, order: int, readonly: bool}> $columnOrder Default column order settings
+	 * @param list<array{columnId: int, mode: 'ASC'|'DESC'}> $sort Default sort rules
 	 * @return DataResponse<Http::STATUS_OK, TablesTable, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
 	 *
 	 * 200: Tables returned
 	 */
 	#[NoAdminRequired]
-	public function createFromScheme(string $title, string $emoji, string $description, array $columns, array $views): DataResponse {
+	public function createFromScheme(string $title, string $emoji, string $description, array $columns, array $views, array $columnOrder = [], array $sort = []): DataResponse {
```
blizzz, line 140: did not check further down the code, but `$columnOrder` and `$sort` are not tested or sanitized for expected input format.

lib/Controller/ApiTablesController.php
```diff
@@ -270,9 +289,15 @@ public function create(string $title, ?string $emoji, ?string $description, stri
 	 */
 	#[NoAdminRequired]
 	#[RequirePermission(permission: Application::PERMISSION_MANAGE, type: Application::NODE_TYPE_TABLE, idParam: 'id')]
-	public function update(int $id, ?string $title = null, ?string $emoji = null, ?string $description = null, ?bool $archived = null): DataResponse {
+	public function update(int $id, ?string $title = null, ?string $emoji = null, ?string $description = null, ?bool $archived = null, null|array|string $columnSettings = null, null|array|string $sort = null): DataResponse {
```
blizzz, line 292: same here for `$sort`

lib/Controller/TableController.php
```diff
@@ -70,9 +70,15 @@ public function destroy(int $id): DataResponse {
 
 	#[NoAdminRequired]
 	#[RequirePermission(permission: Application::PERMISSION_MANAGE, type: Application::NODE_TYPE_TABLE, idParam: 'id')]
-	public function update(int $id, ?string $title = null, ?string $emoji = null, ?bool $archived = null): DataResponse {
-		return $this->handleError(function () use ($id, $title, $emoji, $archived) {
-			return $this->service->update($id, $title, $emoji, null, $archived, $this->userId);
+	public function update(int $id, ?string $title = null, ?string $emoji = null, ?bool $archived = null, null|array|string $columnSettings = null, null|array|string $sort = null): DataResponse {
```
blizzz, line 73: and here

lib/Service/ValueObject/ViewColumnInformation.php
```diff
@@ -59,10 +59,10 @@ public function isMandatory(): bool {
...
 	public static function fromArray(array $data): static {
 		$vci = new static(
-			$data[self::KEY_ID],
-			$data[self::KEY_ORDER],
-			$data[self::KEY_READONLY] ?? false,
-			$data[self::KEY_MANDATORY] ?? false,
+			(int)$data[self::KEY_ID],
+			(int)$data[self::KEY_ORDER],
+			(bool)($data[self::KEY_READONLY] ?? false),
+			(bool)($data[self::KEY_MANDATORY] ?? false),
```
blizzz, line 65: please revert this changes as well. It used to throw a PHP Error on construction, but now it silently succeeds with potentially flawed data.

lib/Db/Table.php
```diff
@@ -121,4 +132,53 @@ private function getSharePermissions(): ?Permissions {
 	private function getViewsArray(): array {
 		return $this->getViews() ?: [];
 	}
+
+	/**
+	 * @psalm-suppress MismatchingDocblockReturnType
```
blizzz, line 137: red flag

lib/Db/Table.php
```diff
@@ -121,4 +132,53 @@ private function getSharePermissions(): ?Permissions {
...
+
+		$result = [];
+		foreach ($columns as $index => $columnId) {
+			$result[] = new ViewColumnInformation($columnId, order: (int)$index + 1);
+		}
+		return $result;
+	}
+
+	/**
+	 * @psalm-suppress MismatchingDocblockReturnType
```
blizzz, line 169: red flag

blizzz: I am also not entirely confident about the code generated here, but it is a bit vague gut feeling yet.

lib/Migration/Version2010Date20260414000000.php
```diff
@@ -0,0 +1,50 @@
...
+namespace OCA\Tables\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+use Override;
+
+class Version2010Date20260414000000 extends SimpleMigrationStep {
```
blizzz, line 19: now too late, but it should have been Version2002, because it follows a funny `XYYY` pattern. From my today`s attempt:

```
$ occ migrations:generate tables 2020
Unexpected migration version for current version: 2.1.1
 - Pattern:  XYYY 
 - Expected: 2001
 - Actual:   2020
```

Now gotta stick with this for sorting. The whole thing will become interesting when exceeding v9. Main mistake by me in 5f871ad18fea021c61697e5d329522f1db2879b0 (and a smaller one in ca314e85224b11d6a065c3ab843ddd0061f01245, did not spot it either).

AndyScherzinger: Ouch, well we could generally rethink the naming adding more digits in total. On the other hand it is not yet released so we might be able to still rename it.

Wdyt? @blizzz

AndyScherzinger: Ah, wait this is 2.1.0 released... Sorry mě being confused, yeah too late, sorry for this 😣

Issue #1508, opened 2024-12-16
Limit fetching of rows to a certain threshold


blizzz:
Contributes to #941 

Currently the frontend tries to load all rows from a table or view. This is currently not a problem in rendering, but can be a problem for loading times, client RAM usage, and limit of roughly 65k/3 in Postgres that we may run against in the current approach.

In #1507 we prepare a backend API that can take necessary parameters to do effective pagination, even with dynamic values provided.

In a first step we want to limit the maximum number of rows to be fetched, while keeping the current pagination system. When requesting the value, dynamic filters and sorting rules shall also be supplied with the query.

The frontend shall request to load 1101 rows (eleven complete pages, plus and indicator of 1 for more rows). If there are more data rows available, the user shall be informed by a message with the hint that filtered loading and sorting can finetune the result set.

Internally, we store the information whether this table or view is loaded fully or just partially.

ISC-PRO:
Good morning. Can the number of lines be limited in customizing or table settings?

blizzz:
> > to load 1101 rows (eleven complete pages, plus and indicator of 1 for more rows)
> 
> Why/How did we choose this number? As in, why not 1001, for example?

The thought behind is is the current page (100 rows) plus 5 pages before and after  (2*5*100), plus 1 for an indicator whether there are more entries.

blizzz:
> Good morning. Can the number of rows be limited in customizing or table settings?

At the moment it is unlimited, so anything that we come up with should be more performant. It has the most effect on the browser, and that can be very individual from laptops and PCs with loads of RAM to puny phones and old devices with little RAM. So ideally we can figure out a good number dynamically, if needed. I would refrain from a setting for this.

ISC-PRO:
> > Good morning. Can the number of rows be limited in customizing or table settings?
> 
> At the moment it is unlimited, so anything that we come up with should be more performant. It has the most effect on the browser, and that can be very individual from laptops and PCs with loads of RAM to puny phones and old devices with little RAM. So ideally we can figure out a good number dynamically, if needed. I would refrain from a setting for this.

Paging is currently displayed. But 100 rows are too many for our users. We would like to limit the number to 25 per page. Can I make this setting?

blizzz:
It is displayed, but it is fake in so far as all rows are being loaded at once (but it is not heavy on the browser). The number is hard coded.

PR #2483, opened 2026-04-15
feat: add permissions for public link shares

### LINKED ISSUE #2389: Permissions for Public link shares
We would like to allow users on public link shares to **create**, **modify**, **delete** rows in tables, when the table manager enables those permissions in sharing. By default, new link shares should still be empty.

Continuation of #67.

benjaminfrueh:
Adds view, create, update and delete permissions for public table shares.

- UI changes are kept minimal for now, as the sharing sidebar should be updated to the unified sharing component in the future
- Rows created/edited via public link shares use `public-<token>` for the `created_by` and `last_edit_by`value.
- If a public link share has only create permission, the UI handles it as a public form with adjusted icon and text

Closes #2389

### DIFF
--- a/lib/Controller/PublicRowOCSController.php
+++ b/lib/Controller/PublicRowOCSController.php
@@ -9,12 +9,14 @@ declare(strict_types=1);
 namespace OCA\Tables\Controller;
 
 use OCA\Tables\AppInfo\Application;
+use OCA\Tables\Db\Row2Mapper;
 use OCA\Tables\Errors\BadRequestError;
 use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Errors\PermissionError;
 use OCA\Tables\Helper\ConversionHelper;
 use OCA\Tables\Middleware\Attribute\AssertShareAccessIsAccessible;
+use OCA\Tables\Model\RowDataInput;
 use OCA\Tables\ResponseDefinitions;
 use OCA\Tables\Service\RowService;
 use OCA\Tables\Service\ShareService;
@@ -37,11 +39,13 @@ class PublicRowOCSController extends AOCSController {
 	public function __construct(
 		protected ShareService $shareService,
 		protected RowService $rowService,
+		protected Row2Mapper $row2Mapper,
 		IRequest $request,
 		LoggerInterface $logger,
 		IL10N $l,
 	) {
 		parent::__construct($request, $logger, $l, '');
+		$this->rowService->setPublicContext();
 	}
 
 	/**
@@ -68,6 +72,10 @@ class PublicRowOCSController extends AOCSController {
 			$shareToken = new ShareToken($token);
 			$share = $this->shareService->findByToken($shareToken);
 
+			if (!$share->getPermissionRead()) {
+				return $this->handlePermissionError(new PermissionError('No read permission on this share'));
+			}
+
 			$limit = $limit !== null ? max(0, min(500, $limit)) : null;
 			$offset = $offset !== null ? max(0, $offset) : null;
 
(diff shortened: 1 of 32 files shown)


lib/Service/PermissionsService.php
```diff
@@ -549,6 +551,11 @@ public function getPermissionArrayForNodeFromContexts(int $nodeId, string $nodeT
 		);
 	}
 
+	public function setPublicContext(): void {
+		$this->userId = '';
```
samin-z, line 555: isn't it better if instead of passing an empty string as userId, pass the share context? it would make the flow more smooth and easier to follow and reduce chances of side effect (if any) in the future.

lib/Controller/ShareController.php
```diff
@@ -89,6 +89,28 @@ public function updatePermission(int $id, string $permission, bool $value): Data
...
 
+	#[NoAdminRequired]
+	public function updatePermissions(
+		int $id,
+		bool $permissionRead = false,
+		bool $permissionCreate = false,
+		bool $permissionUpdate = false,
+		bool $permissionDelete = false,
+	): DataResponse {
+		return $this->handleError(function () use ($id, $permissionRead, $permissionCreate, $permissionUpdate, $permissionDelete) {
```
samin-z, line 100: i think its better to wrap 'updatePermissions' in a transaction, so in case of any permission update fail we have the roll back.

lib/Controller/ShareController.php
```diff
@@ -89,6 +89,28 @@ public function updatePermission(int $id, string $permission, bool $value): Data
...
+			$permissions = [
+				'read' => $permissionRead,
+				'create' => $permissionCreate,
+				'update' => $permissionUpdate && $permissionRead,
+				'delete' => $permissionDelete && $permissionRead,
+			];
+			foreach ($permissions as $permission => $value) {
```
samin-z, line 107: do we need to loop over each permissions? it's better to have a separate 'updatePermission' method in 'shareService.php' to updaze all the permissions in one operation, or extend the existing method to handle bulk update.

lib/Controller/PublicRowOCSController.php
```diff
@@ -90,4 +98,152 @@ public function getRows(string $token, ?int $limit, ?int $offset): DataResponse
...
+
+			$row = $this->rowService->delete($rowId, $viewId, '');
+			return new DataResponse($this->rowService->formatRowsForPublicShare([$row])[0]);
+		} catch (PermissionError $e) {
+			return $this->handlePermissionError($e);
+		} catch (NotFoundError $e) {
+			return $this->handleNotFoundError($e);
+		} catch (InternalError|\Exception $e) {
+			return $this->handleError($e);
+		}
```
samin-z, line 247: for all three of these functions (update/create/delete row) we have few steps repeated, i think a shared helper can reduce duplication and keep the behaviour consistent.

src/store/data.js
```diff
@@ -302,6 +344,21 @@ export const useDataStore = defineStore('data', {
...
+		async removePublicRow({ token, rowId }) {
+			try {
+				await axios.delete(generateOcsUrl('/apps/tables/api/2/public/' + token + '/rows/' + rowId))
+			} catch (e) {
+				displayError(e, t('tables', 'Could not remove row.'))
+				return false
+			}
+			const stateId = 'public-' + token
+			if (this.rows[stateId]) {
+				const filteredRows = this.rows[stateId].filter(r => r.id !== rowId)
```
samin-z, line 356: not a blocker, but is it possible this in bigger tables could slow things down? going over all the rows and comparing the id?

lib/Controller/PublicRowOCSController.php
benjaminfrueh: Thank you for this feedback! The catch block pattern is consistent with the rest of the codebase and changes here would require refactoring of AOCSController and other Controllers which is maybe out of scope of this PR. The token resolution is 2 or 3 lines and I prefer keeping it explicit for readability.

lib/Controller/ShareController.php
benjaminfrueh: Thank you for this feedback! With the change to the `updatePermisson` in one database operation it is already atomic and a transaction should not be necessary anymore.

lib/Controller/ShareController.php
benjaminfrueh: Thank you for this feedback! I changed the `ShareService->updatePermission()` to accept a permissions array instead of just one key and value. The existing single-permission endpoints and the new bulk endpoint now share the same logic and pass it as array. Controller signatures are unchanged to avoid breaking changes.

src/store/data.js
benjaminfrueh: This is true, on big tables this could turn into a performance bottleneck. The problem is that always all rows are fetched into the store. This is also currently the same for authenticated tables. To solve this, we would have to change how pagination works in the tables app, and not fetch all rows at once. But this is out of scope of this PR, and there are already open issues about this, e. g. https://github.com/nextcloud/tables/issues/1508

lib/Service/PermissionsService.php
benjaminfrueh: Good suggestion for a future improvement. The empty string approach is currently consistent with how CLI context is handled in the codebase, so I kept it consistent here. But it is a magic value that bypasses permission checks and it is not obvious why, so maybe worth discussing as a follow-up how we would like to change it, as it would require bigger refactoring.

samin-z (approved):

lib/Service/PermissionsService.php
```diff
@@ -634,7 +641,7 @@ private function basisCheck(Table|View|Context $element, string $nodeType, ?stri
 		}
 
 		if ($userId === '') {
-			return true;
+			return $this->isCli || $this->isPublicContext;
```
enjeck, line 644: i wonder if this breaks existing shares 🤔

benjaminfrueh: It should not break existing shares, anything going through PublicRowOCSController, so also the existing getRows() will have isPublicContext set to true. I just wanted to be more explicit and safe here, in what cases a empty userId should be allowed.

Was there any other case than CLI and isPublicContext when a empty userId should bypass the permission check? We can of course also revert this line and always return true here again, just to be sure.

I think we should think about a way to refactor this later and maybe use something else than an empty userId to bypass the permission checks.

enjeck (dismissed):
When a public share has delete permissions, the guest is able to delete rows from other tables 🙀 😱 

To see, create a public share, open in an incognito/private tab and run in the console:

```
const TOKEN = location.pathname.match(/\/s\/([A-Za-z0-9]{16})\/?$/)?.[1] || 'TOKEN_HERE'
const ROW_ID = 16 // row from some other table

const res = await fetch(
  `${location.origin}/ocs/v2.php/apps/tables/api/2/public/${TOKEN}/rows/${ROW_ID}?format=json`,
  {
    method: 'DELETE',
    headers: {
      'OCS-APIRequest': 'true',
      ...(window.OC?.requestToken ? { requesttoken: OC.requestToken } : {}),
    },
  },
)
```

benjaminfrueh:
Hi @enjeck and thank you for catching the delete issue, I updated the PR. For now to keep it consistent with the `updateSet` method, I passed the `tableId` to validate that the row belongs to the table.

lib/Service/RowService.php
```diff
@@ -712,6 +712,12 @@ public function delete(int $id, ?int $viewId, string $userId): Row2 {
 				throw new InternalError(get_class($this) . ' - ' . __FUNCTION__ . ': ' . $e->getMessage());
 			}
 		} else {
+			if ($tableId !== null && $tableId !== $item->getTableId()) {
+				$e = new \Exception('Row does not belong to table with id ' . $tableId);
+				$this->logger->error($e->getMessage(), ['exception' => $e]);
+				throw new InternalError(get_class($this) . ' - ' . __FUNCTION__ . ': ' . $e->getMessage());
+			}
+
```
blizzz, line 728: theoretically both table and view can be null. Not from the calling Controller, given it is internally called after validation there is no much pressure… but it leaves a funny feeling in the tummy. #paranoia

blizzz (approved):
What we discussed earlier is good, with a non-blocking remark.

lib/Service/RowService.php
benjaminfrueh: As a defensive fix, I now added a simple check to the create, update and delete in the `PublicRowOCSController` to be sure they can not call the service with both `tableId` and `viewId` being `null`.

blizzz (approved):

blizzz:
cypress: it is always the last three cases that fail. So far.

benjaminfrueh:
> cypress: it is always the last three cases that fail. So far.

@blizzz The toast message "Row successfully created." was stacking up ~5 times and covering elements required for cypress tests. The test opens the create row modal, fills and sends it in fast succession. I now added to the cypress helper that any success toast from creating a row will be closed before continue.

Cypress should pass again with that change now.

Before this PR, the success toast was only shown when the "Add more" option is active in the create modal. I changed this with the newly introduced form mode (create only permission), to always give feedback even if no added row is visually shown without the read permission.

blizzz:
> /backport to stable1.0

nope, link sharing was introduced in 2.0 only

PR #2504, opened 2026-04-20
🗃️ Structure import

AndyScherzinger:
WIP for CI execution/checks

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist
 
- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


Koc:
hey @AndyScherzinger!

Thank you for working on this. Really good feature that nice to have. Few points from my side:

- Let's wait before https://github.com/nextcloud/tables/pull/2428 and  https://github.com/nextcloud/tables/pull/2547 got merged? I guess it's pretty logical to use this `technicalName` as a unique key for upserting columns and view. This will allow us to properly handle columns renaming (for now we're deleting old column and creating new one in case of rename. This causes data loss)
  image
- It's hard to understand what exact was changed in view, see previous screen
- Don't know, is it something wrong with my screen/styles, but new title almost not readable for me
  image
- We already have possibility to create table from scheme, see `\OCA\Tables\Controller\ApiTablesController::createFromScheme`. Do you plan to change this controller as well to use newly created service `ApplySchemeService`? IMHO it has sense to reduce code duplication.
- Not sure that we need this for the 1st iteration, but it has sense to use something like [Optimistic Locking](https://martinfowler.com/eaaCatalog/optimisticOfflineLock.html) to prevent applying changes on a table that was recently changed by another user between we pressed "preview" and "apply"
- For some settings I have changes in diff preview but actually nothing changed (Number column, min 0 -> 0)
  image
- It doesn't support column reordering that you've added recently https://github.com/nextcloud/tables/pull/2481

AndyScherzinger:
Hi @Koc

Thanks a lot for the testing and review. All very good points to look into and address. The clear identification of column would be neat to have and makes updating structures easier. Yet for existing ones some/improved heuristic drill might make sense, else you can't make use of this on legacy table structures but that could be a limitation to accept. Maybe.

The colors aren't correct yet, first wanted to to get it working, right, polished. In that order and the current state is still very much in progress.

The issues at the end are currently issues of default values that don't get exported but a target value gets computed, so a difference is detected...

blizzz:
> Let's wait before https://github.com/nextcloud/tables/pull/2428 and https://github.com/nextcloud/tables/pull/2547 got merged? I guess it's pretty logical to use this technicalName as a unique key for upserting columns and view.Let's wait before https://github.com/nextcloud/tables/pull/2428 and https://github.com/nextcloud/tables/pull/2547 got merged? I guess it's pretty logical to use this technicalName as a unique key for upserting columns and view.

It is user-controlled, default empty (or having the title) and also mutable. I would do another change, and store a UUID (v7 or snowflake) next to the simple numeric ID.

blizzz:
> > Let's wait before #2428 and #2547 got merged? I guess it's pretty logical to use this technicalName as a unique key for upserting columns and view.Let's wait before #2428 and #2547 got merged? I guess it's pretty logical to use this technicalName as a unique key for upserting columns and view.
> 
> It is user-controlled, default empty (or having the title) and also mutable. I would do another change, and store a UUID (v7 or snowflake) next to the simple numeric ID.

That would be UUIDv7 in this case. Nextcloud's Snowflake implementation is not suited to this, as the produced IDs are instance-influenced and may have a (tiny) chance of collision when importing. They would be a good replacement for the `id`, but not as a universal identifier.

From my understanding, we essentially need them for *Columns*, as we would import the structure into a chosen table. We can add them for tables and views as well, it could enable easier imports/exports, but we might hit YAGNI and just cram more stuffs into the DB.

The other place would be inside the `selection_options` payload of columns to support updating those values. The `selection_default` currently could reference the ID of the option, or a value (`true`, `false`), depending on the type. For export/import purposes this is not ideal. Could be changed to match the UUID in this case, or the UUID could be used only within the export and import context. Would tend to the latter as it is potentially less breaking.

enjeck (commented):
no longer necessary after https://github.com/nextcloud/tables/pull/2888

PR #2505, opened 2026-04-21
Reordering validation

AndyScherzinger:
Address post-merge feedback on #2481

### DIFF
--- a/lib/Controller/ApiTablesController.php
+++ b/lib/Controller/ApiTablesController.php
@@ -190,7 +190,7 @@ class ApiTablesController extends AOCSController {
 				$remappedColumnOrder = null;
 				if (!empty($columnOrder)) {
 					$remappedColumnOrder = ColumnSettings::createFromInputArray(array_map(static function (array $entry) use ($colMap): array {
-						if (isset($entry['columnId']) && $entry['columnId'] > 0) {
+						if ($entry['columnId'] > 0) {
 							$entry['columnId'] = $colMap[$entry['columnId']] ?? $entry['columnId'];
 						}
 						return $entry;
@@ -199,7 +199,7 @@ class ApiTablesController extends AOCSController {
 				$remappedSort = null;
 				if (!empty($sort)) {
 					$remappedSort = SortRuleSet::createFromInputArray(array_map(static function (array $entry) use ($colMap): array {
-						if (isset($entry['columnId']) && $entry['columnId'] > 0) {
+						if ($entry['columnId'] > 0) {
 							$entry['columnId'] = $colMap[$entry['columnId']] ?? $entry['columnId'];
 						}
 						return $entry;
--- a/lib/Controller/ApiTablesController.php
+++ b/lib/Controller/ApiTablesController.php
@@ -14,6 +14,8 @@ use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Errors\PermissionError;
 use OCA\Tables\Middleware\Attribute\RequirePermission;
+use OCA\Tables\Model\ColumnSettings;
+use OCA\Tables\Model\SortRuleSet;
 use OCA\Tables\Model\ViewUpdateInput;
 use OCA\Tables\ResponseDefinitions;
 use OCA\Tables\Service\ColumnService;
@@ -139,15 +141,11 @@ class ApiTablesController extends AOCSController {
 	 */
 	#[NoAdminRequired]
 	public function createFromScheme(string $title, string $emoji, string $description, array $columns, array $views, array $columnOrder = [], array $sort = []): DataResponse {
-		foreach ($columnOrder as $entry) {
-			if (!is_array($entry) || !isset($entry['columnId'], $entry['order'])) {
-				return new DataResponse(['message' => 'Invalid columnOrder format: each entry requires columnId (int) and order (int)'], Http::STATUS_BAD_REQUEST);
-			}
-		}
-		foreach ($sort as $entry) {
-			if (!is_array($entry) || !isset($entry['columnId'], $entry['mode']) || !in_array($entry['mode'], ['ASC', 'DESC'], true)) {
-				return new DataResponse(['message' => 'Invalid sort format: each entry requires columnId (int) and mode (ASC or DESC)'], Http::STATUS_BAD_REQUEST);
-			}
+		try {
+			ColumnSettings::createFromInputArray($columnOrder);
+			SortRuleSet::createFromInputArray($sort);
+		} catch (\InvalidArgumentException $e) {
+			return new DataResponse(['message' => $e->getMessage()], Http::STATUS_BAD_REQUEST);
 		}
 		try {
 			$this->db->beginTransaction();
@@ -316,25 +314,21 @@ class ApiTablesController extends AOCSController {
 		if (is_string($sort)) {
 			$sort = json_decode($sort, true) ?? null;
 		}
-		if ($columnSettings !== null) {
-			if (!is_array($columnSettings)) {
-				return new DataResponse(['message' => 'Invalid columnSettings: must be a JSON array'], Http::STATUS_BAD_REQUEST);
-			}
-			foreach ($columnSettings as $entry) {
-				if (!is_array($entry) || !isset($entry['columnId'], $entry['order'])) {
-					return new DataResponse(['message' => 'Invalid columnSettings format: each entry requires columnId (int) and order (int)'], Http::STATUS_BAD_REQUEST);
+		try {
+			if ($columnSettings !== null) {
+				if (!is_array($columnSettings)) {
+					throw new \InvalidArgumentException('Invalid columnSettings: must be a JSON array');
 				}
+				ColumnSettings::createFromInputArray($columnSettings);
 			}
-		}
-		if ($sort !== null) {
-			if (!is_array($sort)) {
-				return new DataResponse(['message' => 'Invalid sort: must be a JSON array'], Http::STATUS_BAD_REQUEST);
-			}
-			foreach ($sort as $entry) {
-				if (!is_array($entry) || !isset($entry['columnId'], $entry['mode']) || !in_array($entry['mode'], ['ASC', 'DESC'], true)) {
-					return new DataResponse(['message' => 'Invalid sort format: each entry requires columnId (int) and mode (ASC or DESC)'], Http::STATUS_BAD_REQUEST);
+			if ($sort !== null) {
+				if (!is_array($sort)) {
+					throw new \InvalidArgumentException('Invalid sort: must be a JSON array');
 				}
+				SortRuleSet::createFromInputArray($sort);
 			}
+		} catch (\InvalidArgumentException $e) {
+			return new DataResponse(['message' => $e->getMessage()], Http::STATUS_BAD_REQUEST);
 		}
 		try {
 			return new DataResponse($this->service->update($id, $title, $emoji, $description, $archived, $this->userId, $columnSettings, $sort)->jsonSerialize());
--- a/lib/Controller/ApiTablesController.php
+++ b/lib/Controller/ApiTablesController.php
@@ -187,18 +187,18 @@ class ApiTablesController extends AOCSController {
 				$colMap[$column['id']] = $col->getId();
 			}
 			if (!empty($columnOrder) || !empty($sort)) {
-				$remappedColumnOrder = !empty($columnOrder) ? array_map(static function (array $entry) use ($colMap): array {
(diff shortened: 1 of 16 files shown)


lib/Controller/ApiTablesController.php
```diff
@@ -132,12 +132,23 @@ public function showScheme(int $id): DataResponse {
...
+	 * 400: Invalid request data
 	 */
 	#[NoAdminRequired]
 	public function createFromScheme(string $title, string $emoji, string $description, array $columns, array $views, array $columnOrder = [], array $sort = []): DataResponse {
+		foreach ($columnOrder as $entry) {
+			if (!is_array($entry) || !isset($entry['columnId'], $entry['order'])) {
+				return new DataResponse(['message' => 'Invalid columnOrder format: each entry requires columnId (int) and order (int)'], Http::STATUS_BAD_REQUEST);
+			}
+		}
+		foreach ($sort as $entry) {
```
blizzz: `OCA\Tables\Service\ValueObject\SortRule` should be used instead of re-implementing and duplicating the check (some for the other occurrances)

lib/Controller/ApiTablesController.php
```diff
@@ -132,12 +132,23 @@ public function showScheme(int $id): DataResponse {
...
-	 * @return DataResponse<Http::STATUS_OK, TablesTable, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
+	 * @return DataResponse<Http::STATUS_OK, TablesTable, array{}>|DataResponse<Http::STATUS_BAD_REQUEST|Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
 	 *
 	 * 200: Tables returned
+	 * 400: Invalid request data
 	 */
 	#[NoAdminRequired]
 	public function createFromScheme(string $title, string $emoji, string $description, array $columns, array $views, array $columnOrder = [], array $sort = []): DataResponse {
+		foreach ($columnOrder as $entry) {
+			if (!is_array($entry) || !isset($entry['columnId'], $entry['order'])) {
```
blizzz: Like for with the SortRule I would prefer to add another value object and isolate its logic there. Like in `OCA\Tables\Service\ValueObject\SortRule`, even if it is "only" the checks right now.

lib/Controller/TableController.php
```diff
@@ -78,6 +78,26 @@ public function update(int $id, ?string $title = null, ?string $emoji = null, ?b
 			$sort = json_decode($sort, true) ?? null;
 		}
 		return $this->handleError(function () use ($id, $title, $emoji, $archived, $columnSettings, $sort) {
+			if ($columnSettings !== null) {
```
blizzz: Same here actually. We have `OCA\Tables\Service\ValueObject\ViewColumnInformation` that does similar. I want to avoid wild arrays being passed around as they do not have a strict structure and are terrible to maintain over the code base. Now `ViewColumnInformation` does not exactly fit here, but a superset would do (relevant for table is only column id and order, other aspects are already implemented differently) , so another class can be added that is parent to it, for example.

blizzz (dismissed):

lib/Controller/ApiTablesController.php
```diff
@@ -178,18 +187,18 @@ public function createFromScheme(string $title, string $emoji, string $descripti
...
 			if (!empty($columnOrder) || !empty($sort)) {
-				$remappedColumnOrder = !empty($columnOrder) ? array_map(static function (array $entry) use ($colMap): array {
+				$remappedColumnOrder = !empty($columnOrder) ? ColumnSettings::createFromInputArray(array_map(static function (array $entry) use ($colMap): array {
 					if (isset($entry['columnId']) && $entry['columnId'] > 0) {
 						$entry['columnId'] = $colMap[$entry['columnId']] ?? $entry['columnId'];
 					}
 					return $entry;
-				}, $columnOrder) : null;
-				$remappedSort = !empty($sort) ? array_map(static function (array $entry) use ($colMap): array {
+				}, $columnOrder)) : null;
```
blizzz: coalescing operater over multiple lines with a callable in between is hard to read, and this is also badly formatted. Then better have another if block.

Same with the next part about the `$sort`.

Not critical, but not well readable and thus maintenance cost.

lib/Controller/ApiTablesController.php
```diff
@@ -178,18 +187,18 @@ public function createFromScheme(string $title, string $emoji, string $descripti
 				$colMap[$column['id']] = $col->getId();
 			}
 			if (!empty($columnOrder) || !empty($sort)) {
-				$remappedColumnOrder = !empty($columnOrder) ? array_map(static function (array $entry) use ($colMap): array {
+				$remappedColumnOrder = !empty($columnOrder) ? ColumnSettings::createFromInputArray(array_map(static function (array $entry) use ($colMap): array {
 					if (isset($entry['columnId']) && $entry['columnId'] > 0) {
```
blizzz: `columnOrder` was already verified, hence the first `isset()` is not necessary anymore and can be dropped (but Psalm might complain).

Not blocking.

lib/Controller/ApiTablesController.php
```diff
@@ -178,18 +187,18 @@ public function createFromScheme(string $title, string $emoji, string $descripti
...
 						$entry['columnId'] = $colMap[$entry['columnId']] ?? $entry['columnId'];
 					}
 					return $entry;
-				}, $columnOrder) : null;
-				$remappedSort = !empty($sort) ? array_map(static function (array $entry) use ($colMap): array {
+				}, $columnOrder)) : null;
+				$remappedSort = !empty($sort) ? SortRuleSet::createFromInputArray(array_map(static function (array $entry) use ($colMap): array {
 					if (isset($entry['columnId']) && $entry['columnId'] > 0) {
```
blizzz: Same here with the `isset()`

lib/Service/ValueObject/ColumnOrderInformation.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```
blizzz:
```suggestion
 * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

lib/Service/ValueObject/ViewColumnInformation.php
```diff
@@ -58,51 +42,34 @@ public function isMandatory(): bool {
 	}
 
 	public static function fromArray(array $data): static {
-		$vci = new static(
+		if (!isset($data[self::KEY_ID], $data[self::KEY_ORDER])) {
```
blizzz: Is now duplicated (same with parent), but here the parent method cannot just be called. Cleaner solution was to move it into a separate method on the parent, and call it from here (and the parent constructor).

lib/Service/ValueObject/ViewColumnInformation.php
```diff
@@ -58,51 +42,34 @@ public function isMandatory(): bool {
...
 
 	/**
-	 * @return array{columnId: int, order: int, readonly?: bool, mandatory?: bool}
+	 * @return array{columnId: int, order: int, readonly: bool, mandatory: bool}
 	 */
 	public function jsonSerialize(): array {
-		return $this->data;
+		return [
+			self::KEY_ID => $this->getId(),
+			self::KEY_ORDER => $this->getOrder(),
```
blizzz: also here parent logic is duplicated

lib/Service/ValueObject/ViewColumnInformation.php
```diff
@@ -58,51 +42,34 @@ public function isMandatory(): bool {
...
+			self::KEY_MANDATORY => $this->isMandatory(),
+		];
 	}
 
 	protected function ensureType(string $offset, mixed $value): int|bool {
 		return match ($offset) {
 			self::KEY_ID,
 			self::KEY_ORDER => (int)$value,
-			self::KEY_READONLY => (bool)$value,
+			self::KEY_READONLY,
```
blizzz: irritating formatting, would revert, not blocking.

lib/Service/ValueObject/ViewColumnInformation.php
```diff
@@ -58,51 +42,34 @@ public function isMandatory(): bool {
...
+			self::KEY_ORDER => $this->getOrder(),
+			self::KEY_READONLY => $this->isReadonly(),
+			self::KEY_MANDATORY => $this->isMandatory(),
+		];
 	}
 
 	protected function ensureType(string $offset, mixed $value): int|bool {
 		return match ($offset) {
 			self::KEY_ID,
 			self::KEY_ORDER => (int)$value,
```
blizzz: also duplicated from parent

blizzz (commented):
We can take it, but it has still room for improvement,

lib/Service/ValueObject/ViewColumnInformation.php
AndyScherzinger: https://github.com/nextcloud/tables/pull/2505/commits/3efa881b93b9ddd137b67c032a0497e141f475d0

lib/Service/ValueObject/ViewColumnInformation.php
AndyScherzinger: https://github.com/nextcloud/tables/pull/2505/commits/3efa881b93b9ddd137b67c032a0497e141f475d0

lib/Service/ValueObject/ViewColumnInformation.php
AndyScherzinger: https://github.com/nextcloud/tables/pull/2505/commits/3efa881b93b9ddd137b67c032a0497e141f475d0

lib/Service/ValueObject/ViewColumnInformation.php
AndyScherzinger: https://github.com/nextcloud/tables/pull/2505/commits/3efa881b93b9ddd137b67c032a0497e141f475d0

lib/Service/ValueObject/ColumnOrderInformation.php
AndyScherzinger: https://github.com/nextcloud/tables/pull/2505/commits/71620ecc7a39cf763d2855161dee24dd9aa3de16

lib/Controller/ApiTablesController.php
AndyScherzinger: https://github.com/nextcloud/tables/pull/2505/commits/7560fbe073c7b9d578f610af3a4ddf82c944bd83

lib/Controller/ApiTablesController.php
AndyScherzinger: https://github.com/nextcloud/tables/pull/2505/commits/7560fbe073c7b9d578f610af3a4ddf82c944bd83

lib/Controller/ApiTablesController.php
AndyScherzinger: https://github.com/nextcloud/tables/pull/2505/commits/03a290f0288548179afc30274d11cf7ed638ad34

lib/Controller/ApiTablesController.php
AndyScherzinger: added again, else psalm breaks: https://github.com/nextcloud/tables/actions/runs/24894918778/job/72897053951?pr=2505

lib/Controller/ApiTablesController.php
AndyScherzinger: added again, else psalm breaks: https://github.com/nextcloud/tables/actions/runs/24894918778/job/72897053951?pr=2505

AndyScherzinger:
Addressed all remaining comment, feel free to review again @blizzz

lib/Service/ValueObject/ColumnOrderInformation.php
```diff
@@ -0,0 +1,94 @@
...
+
+	public function offsetUnset(mixed $offset): void {
+		if (!$this->offsetExists($offset)) {
+			return;
+		}
+		unset($this->data[(string)$offset]);
+	}
+
+	/**
+	 * @return array{columnId: int, order: int, ...}
```
blizzz, line 78:
```suggestion
	 * @return array{columnId: int, order: int}
```

AndyScherzinger: @blizzz seems to not be needed to make psalm happy, but can be changed if you think it is useful anyways.

lib/Service/ValueObject/ViewColumnInformation.php
blizzz: The change in the commit looks different than the effective code here. That caused the psalm trouble?

blizzz (approved):

PR #2513, opened 2026-04-24
build(release): v2.1.0

enjeck:


mejo- (approved):

AndyScherzinger (approved):

appinfo/info.xml
```diff
@@ -26,7 +26,7 @@ Share your tables and views with users and groups within your cloud.
 Have a good time and manage whatever you want.
 
 ]]></description>
-	<version>2.0.1</version>
+	<version>2.0.2</version>
```
AndyScherzinger: @enjeck Given https://github.com/nextcloud/tables/pull/2481/changes#diff-238d8f4f5557b76897722af71f716283c21e2b6545f1ee1e94e3b9f37859f4b8 stays in I would suggest 2.1.0 due to it being a feature and the DB migration file naming also hinting towards it, else future database migrations might not be sorted correctly

blizzz: With the Migration class now in it should be 2.1.0 indeed. 

Feature-wise as well, strictly speaking, though it is not always followed religiously, though would be a good thing to have it consistent of course:  https://semver.org/

AndyScherzinger: DONE https://github.com/nextcloud/tables/pull/2513/commits/d6b52cb962d699a7abd57166f9366708a5eeff4c

PR #2547, opened 2026-04-28
Feat: Add technicalName to views

Koc:
This is similar to #1840/#2384 but for Views.

image

image

image

## :construction: TODO
- [x] Extend tests

### DIFF
--- a/lib/Service/ViewService.php
+++ b/lib/Service/ViewService.php
@@ -21,6 +21,7 @@ use OCA\Tables\Db\Column;
 use OCA\Tables\Db\Table;
 use OCA\Tables\Db\View;
 use OCA\Tables\Db\ViewMapper;
+use OCA\Tables\Errors\BadRequestError;
 use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Errors\PermissionError;
@@ -195,11 +196,13 @@ class ViewService extends SuperService {
 	 * @param string|null $emoji
 	 * @param Table $table
 	 * @param string|null $userId
+	 * @param string|null $technicalName
 	 * @return View
 	 * @throws InternalError
 	 * @throws PermissionError
+	 * @throws BadRequestError
 	 */
-	public function create(string $title, ?string $emoji, Table $table, ?string $userId = null): View {
+	public function create(string $title, ?string $emoji, Table $table, ?string $userId = null, ?string $technicalName = null): View {
 		/** @var string $userId */
 		$userId = $this->permissionsService->preCheckUserId($userId, false); // $userId is set
 
@@ -221,6 +224,10 @@ class ViewService extends SuperService {
 		$item->setLastEditBy($userId);
 		$item->setCreatedAt($time->format('Y-m-d H:i:s'));
 		$item->setLastEditAt($time->format('Y-m-d H:i:s'));
+		if ($technicalName !== null) {
+			$this->assertTechnicalNameValid($technicalName);
+			$item->setTechnicalName($technicalName);
+		}
 		// ownership is not stored with the record, but it might be necessary upon
 		// further interaction with the view in the running process, as the instance
 		// is cached now. The ownership is always inherited from the table.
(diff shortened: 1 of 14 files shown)


enjeck (commented):
i guess we wait for https://github.com/nextcloud/tables/pull/2428 to merge first?

Koc:
@enjeck this is more or less parallel process. My PR is completely independent.

enjeck:
> @enjeck this is more or less parallel process. My PR is completely independent.

There's some designer feedback over there that I think applies here too? @samin-z please review.

samin-z:
@enjeck yes ill aligh the design on column to be the same as view as talked with @nimishavijay

lib/Service/ViewService.php
```diff
@@ -620,9 +643,51 @@ public function importView(int $tableId, array $view, string $userId): void {
...
+			throw new InternalError(get_class($this) . ' - ' . __FUNCTION__ . ': ' . $e->getMessage(), previous: $e);
+		}
+
+		foreach ($views as $view) {
+			if ($excludeCurrentViewId !== null && $view->getId() === $excludeCurrentViewId) {
+				continue;
+			}
+
+			if ($view->getTechnicalName() === $technicalName) {
+				throw new BadRequestError('Technical name must be unique in the table.');
```
samin-z: unique technical name does not take effect, i can still safe the view with an existing technical name.

samin-z:
if there is no technical_name value for view when updating no default is added for that view.

samin-z (changes requested):

samin-z:
when api is called for view, dataByAlias is empty (screenshot 1), but in view it's an empty array (screenshot 2).
Screenshot 2026-07-08 at 16 52 43Screenshot 2026-07-08 at 16 52 24

lib/Service/ViewService.php
```diff
@@ -620,9 +643,51 @@ public function importView(int $tableId, array $view, string $userId): void {
...
+		if (!preg_match('/^[a-z][a-z0-9_]*$/', $technicalName)) {
+			throw new BadRequestError('Technical name must start with a lowercase letter and contain only lowercase letters, numbers, and underscores.');
+		}
+	}
+
+	/**
+	 * @throws BadRequestError
+	 * @throws InternalError
+	 */
+	private function assertTechnicalNameUnique(int $tableId, string $technicalName, ?int $excludeCurrentViewId = null): void {
```
samin-z: maybe we can remove the findAll and foreach uniqueness check and rely on the db unique index on (table_id, technical_name)? on create/update catch the constraint violation and return the same BadRequestError (“Technical name must be unique in the table.”).

lib/Migration/Version2200Date20260708000000.php
```diff
@@ -0,0 +1,84 @@
...
+					->set('technical_name', $updateQb->createNamedParameter('view_' . $viewId, IQueryBuilder::PARAM_STR))
+					->where($updateQb->expr()->eq('id', $updateQb->createNamedParameter($viewId, IQueryBuilder::PARAM_INT)));
+
+				$updatedCount += $updateQb->executeStatement();
+			}
+		} finally {
+			$result->closeCursor();
+		}
+
+		$output->info('Version2011Date20260428000000: backfilled technical_name for ' . $updatedCount . ' views.');
```
samin-z: this also needs to be updated.

lib/Service/ViewService.php
```diff
@@ -620,9 +643,51 @@ public function importView(int $tableId, array $view, string $userId): void {
...
 	}
+
+	private function buildDefaultTechnicalName(int $viewId): string {
+		return 'view_' . $viewId;
+	}
+
+	/**
+	 * @throws BadRequestError
+	 */
+	private function assertTechnicalNameValid(string $technicalName): void {
```
samin-z, line 697: low importance but maybe a check also for the length of it, but nothing worth stopping to go forward.

Koc:
@samin-z @enjeck I did all changes, please re-review. Tests are green, except query count (which is expected)

Koc:
hey @samin-z!

I've squashed commits, rebased PR, fixed conflicts, covered duplicated technical name with Behat test. Here the :movie_camera: demo:

https://github.com/user-attachments/assets/d4a08015-619e-42ac-9dd6-2cd0ea1b686d


Double-check that migration was properly executed.

Other test failures are unrelated, see https://github.com/nextcloud/tables/pull/2829 for fix

lib/Migration/Version2200Date20260708000000.php
```diff
@@ -0,0 +1,84 @@
...
+					$selectQb->expr()->isNull('technical_name'),
+					$selectQb->expr()->eq('technical_name', $selectQb->createNamedParameter('')),
+				)
+			);
+
+		$result = $selectQb->executeQuery();
+		$updatedCount = 0;
+
+		try {
+			while ($row = $result->fetchAssociative()) {
```
samin-z: this could be an N+1 issue, better to use single UPDATE query with CONCAT function.

Koc: fixed and re-tested migration

image

samin-z (approved):

Koc:
rebased, fixed conflicts, tested basic scenarios. Pipeline failures not related to PR

PR #2548, opened 2026-04-28
docs(agents): Tune AGENTS.md to properly execute linters

Koc:


blizzz:
Why, that is already proper and should not need it? It works without, but that is creepy, imo, as those are not native composer commands.

Koc:
hm, indeed :smile:. I had some troubles with those scripts yesterday. And I was thinking that the reason was this extra `run`.

But it was a problem with my setup: I'm using docker and Agent wasn't aware of that.

Sorry for confusion

PR #2554, opened 2026-04-29
📊 feat(Analytics): enhance datasource

Rello:
The datasource for the Analytics app is enhanced to provide more reporting capabilities: 

- A `count` column=1 is added for statistics in Analytics

The datasource now returns analytics-friendly scalar values for:

- single selections as option labels
- multi selections as comma-separated option labels
- checkbox selections as true / false
- text links as Title (URL) when a title exists, or only the URL when title is empty
- rich/long/line text with HTML tags stripped
- user/group/team values as comma-separated display names or IDs
- blank cells backed by column defaults, including number, text, selection, datetime, and usergroup defaults

Unit test was added

### DIFF
--- a/lib/Analytics/AnalyticsDatasource.php
+++ b/lib/Analytics/AnalyticsDatasource.php
@@ -22,8 +22,6 @@ use OCP\IL10N;
 use Psr\Log\LoggerInterface;
 
 class AnalyticsDatasource implements IDatasource {
-	private const COUNT_COLUMN_TITLE = 'count';
-	private const COUNT_COLUMN_VALUE = 1;
 
 	private LoggerInterface $logger;
 	private IL10N $l10n;
@@ -242,7 +240,7 @@ class AnalyticsDatasource implements IDatasource {
 		foreach ($columns as $column) {
 			$header[] = $column->getTitle();
 		}
-		$header[] = self::COUNT_COLUMN_TITLE;
+		$header[] = $this->l10n->t('Count');
 		$data[] = $header;
 
 		// now add the rows
@@ -262,7 +260,7 @@ class AnalyticsDatasource implements IDatasource {
 				}
 				$line[] = $value;
 			}
-			$line[] = self::COUNT_COLUMN_VALUE;
+			$line[] = 1; // constant 1 for the count column
 			$data[] = $line;
 		}
 		return $data;
--- a/lib/Analytics/AnalyticsDatasource.php
+++ b/lib/Analytics/AnalyticsDatasource.php
@@ -8,6 +8,7 @@
 namespace OCA\Tables\Analytics;
 
 use OCA\Analytics\Datasource\IDatasource;
+use OCA\Tables\Db\Column;
 use OCA\Tables\Errors\InternalError;
 use OCA\Tables\Errors\NotFoundError;
 use OCA\Tables\Errors\PermissionError;
@@ -21,6 +22,9 @@ use OCP\IL10N;
 use Psr\Log\LoggerInterface;
 
 class AnalyticsDatasource implements IDatasource {
+	private const COUNT_COLUMN_TITLE = 'count';
+	private const COUNT_COLUMN_VALUE = 1;
+
 	private LoggerInterface $logger;
 	private IL10N $l10n;
 	private TableService $tableService;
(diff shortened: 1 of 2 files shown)


Rello:
Bildschirmfoto 2026-04-29 um 20 11 12

Rello:
Playwright tests are not relevant to this Analytics datasource

lib/Analytics/AnalyticsDatasource.php
```diff
@@ -229,29 +251,171 @@ private function getData(int $nodeId, ?int $limit, ?int $offset, ?string $nodeTy
...
-						}
+						$value = $this->formatValue($column, $datum['value']);
 					}
 				}
-				// Tables does not deliver any values for "blank" default columns
-				if ($value === '' && $column->getType() === 'number') {
-					$value = $column->getNumberDefault();
+				// Tables does not deliver any values for "blank" default columns.
+				if ($value === '') {
+					$value = $this->formatDefaultValue($column);
```
enjeck, line 259: why switch from number to every type? So a cell that's legitimately empty on a column that has a default (text/selection/usergroup) will now render the default instead of blank? Is that the intendion for these types, or should the default only fill truly-missing cells rather than empty-string values?

lib/Analytics/AnalyticsDatasource.php
```diff
@@ -229,29 +251,171 @@ private function getData(int $nodeId, ?int $limit, ?int $offset, ?string $nodeTy
...
+
+	private function formatTextValue(Column $column, mixed $value): string {
+		if ($value === null || $value === '') {
+			return '';
+		}
+
+		$value = (string)$value;
+		if ($column->getSubtype() === 'link') {
+			return $this->formatLinkValue($value);
+		}
```
enjeck, line 349: can we use the predefined constants instead of hardcoding e.g `SUBTYPE_TEXT_LINK`

lib/Analytics/AnalyticsDatasource.php
```diff
@@ -229,29 +251,171 @@ private function getData(int $nodeId, ?int $limit, ?int $offset, ?string $nodeTy
...
+			if ((int)$option['id'] === (int)$value) {
+				return $option['label'];
+			}
+		}
+
+		return '';
+	}
+
+	private function isMultiSelection(Column $column): bool {
+		return in_array($column->getSubtype(), [Column::SUBTYPE_SELECTION_MULTI, 'multi'], true);
```
enjeck, line 311: not sure we need the `multi` option

tests/unit/Analytics/AnalyticsDatasourceTest.php
```diff
@@ -0,0 +1,245 @@
...
+			->method('findAllByTable')
+			->with(123, 'user1')
+			->willReturn([
+				$this->createColumn(1, 'Link', 'text', 'link'),
+				$this->createColumn(2, 'Notes', 'text', 'rich'),
+				$this->createSelectionColumn(3, 'Status', '', [
+					['id' => 1, 'label' => 'Open'],
+					['id' => 2, 'label' => 'Closed'],
+				]),
+				$this->createSelectionColumn(4, 'Tags', 'multi', [
```
enjeck: should be `'selection-multi'` instead of `multi`?

enjeck (commented):
some comments

lib/Analytics/AnalyticsDatasource.php
Rello: Hello,
good question. Can a cell value be completely missing? 
Aim is e.g. a count-column for row counts. make it default=1 but the user does not need to fill the column. is this empty or ''?

lib/Analytics/AnalyticsDatasource.php
Rello: thanks for the hint. makes sense. I will change it

Rello: Hello,
I misunderstood this comment. I thought that these were types from the app already. This string is only used one time, so there is no reusable to put it in a constant for just this datasource

lib/Analytics/AnalyticsDatasource.php
Rello: it seems there was some usage so it looks like a legacy case
https://github.com/nextcloud/tables/blob/main/lib/Db/Row2Mapper.php#L477

lib/Analytics/AnalyticsDatasource.php
enjeck: I guess a case where a there's a cell a user explicitly cleared to an empty string. But should be fine, i guess

lib/Analytics/AnalyticsDatasource.php
enjeck: my bad, i thought we had that

enjeck (approved):
thanks

Rello:
🙏🏻

PR #2556, opened 2026-04-30
fix/2555: share delete and public link access

samin-z:
apply share delete and public link access respecting to limitations


benjaminfrueh:
@samin-z as we just discussed, this could maybe be simplified.
The public controller should not use the user session as it should not require a user and only use the token (and optionally a password) to authenticate.

I think the easiest way would be to get the sender user from the share and simply validate this user with `sharingDisabledForUser()` - I didn't test that though.

The files app does the public link checks checks in https://github.com/nextcloud/server/blob/master/apps/files_sharing/lib/Controller/ShareController.php#L263
Maybe it makes sense for tables to also check if the share sender user is disabled, and also deny access then, like the files_app already does.

benjaminfrueh (approved):

PR #2572, opened 2026-05-04
feat: add is not empty filter support

Koc:
image


src/shared/components/ncTable/mixins/filter.js
```diff
@@ -50,86 +50,94 @@ export const FilterIds = {
...
-		incompatibleWith: [FilterIds.Contains, FilterIds.BeginsWith, FilterIds.EndsWith, FilterIds.IsEqual, FilterIds.IsGreaterThan, FilterIds.IsGreaterThanOrEqual, FilterIds.IsLowerThan, FilterIds.IsLowerThanOrEqual, FilterIds.IsEmpty],
+		incompatibleWith: [FilterIds.Contains, FilterIds.BeginsWith, FilterIds.EndsWith, FilterIds.IsEqual, FilterIds.IsGreaterThan, FilterIds.IsGreaterThanOrEqual, FilterIds.IsLowerThan, FilterIds.IsLowerThanOrEqual, FilterIds.IsEmpty, FilterIds.IsNotEmpty],
+		noSearchValue: true,
+	}),
+	IsNotEmpty: new Filter({
+		id: FilterIds.IsNotEmpty,
+		label: t('tables', 'Is not empty'),
+		goodFor: [ColumnTypes.TextLine, ColumnTypes.TextRich, ColumnTypes.Number, ColumnTypes.TextLink, ColumnTypes.NumberProgress, ColumnTypes.DatetimeDate, ColumnTypes.DatetimeTime, ColumnTypes.Datetime, ColumnTypes.SelectionCheck, ColumnTypes.Usergroup, ColumnTypes.Relation],
+		incompatibleWith: [FilterIds.Contains, FilterIds.BeginsWith, FilterIds.EndsWith, FilterIds.IsEqual, FilterIds.IsGreaterThan, FilterIds.IsGreaterThanOrEqual, FilterIds.IsLowerThan, FilterIds.IsLowerThanOrEqual, FilterIds.IsEmpty, FilterIds.IsNotEmpty],
```
enjeck, line 140: creating a filter doesn't work for me. Maybe because this operator is exposed via `getPossibleOperators()`, which also feeds the persisted view-filter editor (FilterEntry.vue), so these filters run server-side in SQL. The backend has no `is-not-empty` and `FilterOperator` enum lacks the case

src/shared/components/ncTable/mixins/columnsTypes/numberStars.js
enjeck, line 1: NumberStars isn't in `IsNotEmpty.goodFor` (nor `IsEmpty.goodFor`), so this handler is never reached? Same applies to selection.js, selectionMulti.js, and textLong.js?

enjeck (changes requested):
Doesn't seem to work yet 🤔

src/shared/components/ncTable/mixins/columnsTypes/numberStars.js
Koc: fixed, tested all possible column types, it works everywhere except selection-check.

I've removed it from selection-check. Anyway "is-empty" is also not works properly for this column type (we always have "true" or "false", without emptiness)

src/shared/components/ncTable/mixins/filter.js
Koc: I've added support filtering on a server side and tested it on view side, please check one more time


https://github.com/user-attachments/assets/1908e690-949f-4467-af80-d7f506240fa4

lib/Db/Row2Mapper.php
```diff
@@ -568,6 +568,17 @@ private function getFilterExpression(IQueryBuilder $qb, Column $column, string $
...
 				break;
+			case 'is-not-empty':
+				$includeDefault = !empty($defaultValue);
+				if ($column->getType() === Column::TYPE_TEXT) {
+					$filterExpression = $qb2->expr()->andX(
+						$qb->expr()->isNotNull('value'),
+						$qb->expr()->neq('value', $qb->createNamedParameter('', $paramType))
+					);
+				} else {
+					$filterExpression = $qb->expr()->isNotNull('value');
```
enjeck, line 589: multi-selection is not `TYPE_TEXT`, so it hits this branch. Empty multi-selects are stored as the string '[]', not NULL. So `isNotNull('value')` returns rows whose multi-select is empty, i think 🤔 

We prolly need a new branch like the TEXT one:

```python
elseif ($column->getType() === 'selection' && $column->getSubtype() === 'multi') {
    $filterExpression = $qb2->expr()->andX(
        $qb->expr()->isNotNull('value'),
        $qb->expr()->neq('value', $qb->createNamedParameter('[]', $paramType))
    );
}
```

lib/Db/Row2Mapper.php
```diff
@@ -568,6 +568,17 @@ private function getFilterExpression(IQueryBuilder $qb, Column $column, string $
 					$filterExpression = $qb->expr()->isNull('value');
```
enjeck, line 573: Similar as i said in https://github.com/nextcloud/tables/pull/2572/changes#r3887425566. This PR adds `SelectionMulti` to `is-empty` too, and `isNull('value')` misses '[]' cells.

lib/Db/Row2Mapper.php
```diff
@@ -568,6 +568,17 @@ private function getFilterExpression(IQueryBuilder $qb, Column $column, string $
...
 				}
 				break;
+			case 'is-not-empty':
```
enjeck, line 576: would be nice to have a test for this new branch

lib/Db/Row2Mapper.php
Koc: added, seems like it works fine

table with empty and non-empty rows
image

views with empty and non-empty filters for multi-selection column

image

image

lib/Db/Row2Mapper.php
Koc: added, see `tests/unit/Db/Row2MapperFilterTest.php`

enjeck (approved):
tysm!

PR #2573, opened 2026-05-05
fix: Adjust date formatting

enjeck:
When you select a date in the date column, it subtracts one day. This only happens with the inline editing for date columns. This fixes it.


benjaminfrueh (approved):

samuelx12:
I think that just fixes a symptom rather than the whole problem.
The same shift also applies to times. I only figured this out this morning, and I think I understand where the bug comes from. Please read my latest comment on this issue #2479

PR #2582, opened 2026-05-07
fix: handle first option select

### LINKED ISSUE #2545: Column not save when select the first option
### Steps to reproduce

When edit a row, the selection type column cannot be saved when select the first option.

1. Create a single selection column in a table
2. Create new row and fill all columns except the new selection type column
3. Edit the new created row and click save without any change
4. Edit that row again and select the first option in the selection type column
5. Click save

### Expected behavior

Save should be successful.

### Actual behavior

The selection type column cannot be saved.

### Tables app version

_No response_

### Browser

_No response_

### Client operating system

_No response_

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

_No response_

luka-nextcloud:
* Resolves: #2545

## Summary
Cell value parsing improvements: adjust the value parsing logic in the `RowService`. Add handle cases where the parsed value is the integer `0`, ensuring it is returned as the string `'0'` instead of being treated as a null value.

### DIFF
--- a/lib/Service/ColumnTypes/SelectionBusiness.php
+++ b/lib/Service/ColumnTypes/SelectionBusiness.php
@@ -24,7 +24,7 @@ class SelectionBusiness extends SuperBusiness {
 
 		foreach ($column->getSelectionOptionsArray() as $option) {
 			if ($option['id'] === $intValue) {
-				return json_encode($option['id']);
+				return json_encode((string)$option['id']);
 			}
 		}
 
--- a/tests/unit/Service/ColumnTypes/SelectionBusinessTest.php
+++ b/tests/unit/Service/ColumnTypes/SelectionBusinessTest.php
@@ -33,9 +33,9 @@ class SelectionBusinessTest extends TestCase {
 
 	public function parseValueProvider(): array {
 		return [
-			'valid integer value' => [2, '2'],
-			'valid string value' => ['2', '2'],
-			'valid string value for numeric option' => ['4', '4'],
+			'valid integer value' => [2, '"2"'],
+			'valid string value' => ['2', '"2"'],
+			'valid string value for numeric option' => ['4', '"4"'],
 			'invalid value' => [5, ''],
 			'null value' => [null, ''],
 			'empty string' => ['', ''],


lib/Service/RowService.php
```diff
@@ -491,7 +491,8 @@ private function parseValueByColumnType(Column $column, $value = null) {
 		try {
 			$columnBusiness = $this->columnsHelper->getColumnBusinessObject($column);
 			if ($columnBusiness->canBeParsed($value, $column)) {
-				return json_decode($columnBusiness->parseValue($value, $column), true);
+				$result = json_decode($columnBusiness->parseValue($value, $column), true);
+				return $result === 0 ? '0' : $result;
```
samin-z: correct me if im wrong, if $result is 0 we return '0' or else it can be a string? or is it a integer? because if it's the second then there is inconsistency.

luka-nextcloud: @samin-z Yes, we only return the string value `0` if the result is number zero. What inconsistency are you referring to?

samin-z: @luka-nextcloud i meant what $result could be if not 0? if it's like 1 or 'string'? because of $result is for example 1 then we return integer 1 but if it's 0 we return '0', not sure if i can get my point across.

benjaminfrueh: @luka-nextcloud @samin-z I understand Samin's point here as a valid concern, it is not ideal and has no clear intent why to have string and integer types mixed here. Also this change might not be scoped enough to selection columns and affects all column types, e. g. a number type 0 could also become string 0.

I tried to understand why this issue is happening, so I think this is the root cause:
The `SelectionBusiness::parseValue()` encodes the option ID as an unquoted integer, so `json_decode`returns integer `0` for the first option. This then hits the intentional loose `!=` comparision in `Row2::insertOrUpdateCell()` so it compares null with 0 and skips the update.

Do we know why this is not `!==` here in `Row2::insertOrUpdateCell()` - as strict comparison here would fix this issue, but as the comment on this line says `// yes, no type safety here` there must be a reason it was added not strict.

One solution would be to make the `SelectionBusiness::parseValue()` already always json_encode the option id as string, so instead of `return json_encode($option['id']);` you could make the option id a string in the json `return json_encode((string)$option['id']);` - this should also fix the issue and make the data more consistent (strings '0', '1', '2' for the options).

benjaminfrueh (approved):

samin-z (approved):

PR #2583, opened 2026-05-07
revert: timezone conversion intended for meta fields, accidentally broke datetime columns

### LINKED ISSUE #2479: Datetime column shifts time by timezone offset on every save (DST-aware, Tables 2.0.x)

samuelx12:
Reverts PR #2210 (commits 6936f84, 78f8131) and PR #2573 (commit 172a97b6).

The change in #2210 was intended to fix UTC display of meta fields (createdAt/lastEditAt), but accidentally applied UTC conversion to all user-facing datetime columns. This causes time values to shift on every save and misinterprets all existing data for users outside UTC.

PR #2573 patches one symptom of the same root cause and is therefore included in this revert.

For the full analysis, see: nextcloud/tables#2479 ([comment](https://github.com/nextcloud/tables/issues/2479#issuecomment-4397087463))

The original meta field display issue that #2210 was trying to fix still exists and should be addressed separately.

 Fixes #2479


enjeck (approved):
Thanks for fixing.

PR #2586, opened 2026-05-09
perf: cache shared view permission lookups

enjeck:
This improves /apps/tables/view performance for users with shared views by avoiding repeated permission/share resolution during one request. Added request-local caching in PermissionsService for user group IDs, user circle IDs, shared permissions per user/node and context permissions per user/node. 

### To see the problem:
1. Create a table as alice.
2. Add 17 rows.
3. Create 8 views for that table.
4. Share each view with a group containing bob.
5. Log in as bob and load /index.php/apps/tables/.
6. Check the GET /apps/tables/view XHR.

Now, the load time is improved. 
**Before:** /apps/tables/view had slow startup, around 1.4s, with many repeated oc_tables_shares queries.

**After:** /apps/tables/view dropped to about 600ms. So we have a 2x speedup

### DIFF
--- a/lib/Service/PermissionsService.php
+++ b/lib/Service/PermissionsService.php
@@ -49,6 +49,18 @@ class PermissionsService {
 
 	private ContextMapper $contextMapper;
 
+	/** @var array<string, list<string>> Per-request group IDs keyed by user ID. */
+	private array $groupIdsByUserId = [];
+
+	/** @var array<string, list<string>> Per-request circle IDs keyed by user ID. */
+	private array $circleIdsByUserId = [];
+
+	/** @var array<string, Permissions|null> Per-request shared permissions keyed by node and user. */
+	private array $sharedPermissionsByNode = [];
+
+	/** @var array<string, int|null> Per-request context permissions keyed by node and user. */
+	private array $contextPermissionsByNode = [];
+
 	public function __construct(
 		LoggerInterface $logger,
 		?string $userId,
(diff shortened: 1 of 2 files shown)


lib/Service/PermissionsService.php
```diff
@@ -522,23 +543,32 @@ public function getSharedPermissionsIfSharedWithMe(int $elementId, string $eleme
...
 
-			return new Permissions(
+			$permissions = new Permissions(
 				read: $read,
 				create: $create,
 				update: $update,
 				delete: $delete,
 				manage: $manage,
 			);
+			$this->sharedPermissionsByNode[$cacheKey] = clone $permissions;
```
blizzz: clone should not be necessary, or do i oversee something?

lib/Service/ViewService.php
```diff
@@ -173,7 +173,7 @@ public function findSharedViewsWithMe(?string $userId = null): array {
 				) {
 					continue;
 				}
-				$sharedViews[$node['node_id']] = $this->find($node['node_id'], false, $userId);
+				$sharedViews[$node['node_id']] = $this->find($node['node_id'], true, $userId);
```
blizzz, line 177: Is this not needed through `ViewController::indexSharedWithMe()`?

lib/Service/PermissionsService.php
enjeck: The clone was only needed because the returned Permissions object was later mutated in ViewService to add the view-specific manageTable flag. 
I changed that so ViewService now creates a separate Permissions object for onSharePermissions instead of mutating the cached permission result. With that fixed, the cache can store and return the Permissions object directly.

lib/Service/ViewService.php
enjeck: `indexSharedWithMe()` still receives enhanced views. `findSharedViewsWithMe()` enhances every collected shared view in the final loop below, so this `find(..., true, ...)` only skips the earlier per-view enhancement for context-derived views. Without it, those views are enhanced twice.

blizzz (approved):

PR #2598, opened 2026-05-13
feat: add UUIDs to columns

blizzz:
Base for structural updates, cf. my comment in https://github.com/nextcloud/tables/pull/2504#issuecomment-4444240971

* [x] Add UUIDs to Tables or consider unnecessary
* [x] Add UUIDs to Views or consider unnecessary
* [x] Add UUIDs to multiselection options
* [x] Take over column UUIDs when creating columns from schema
* [ ] ~~Treat multiselection default on export and import~~

### DIFF
--- a/.scoper-production-dependencies
+++ b/.scoper-production-dependencies
@@ -4,3 +4,5 @@ maennchen/zipstream-php
 markbaker/complex
 markbaker/matrix
 psr/simple-cache
+symfony/uid
+symfony/polyfill-uuid
--- a/lib/Db/Column.php
+++ b/lib/Db/Column.php
@@ -10,6 +10,7 @@ namespace OCA\Tables\Db;
 use JsonSerializable;
 use OCA\Tables\Constants\ColumnType;
 use OCA\Tables\Dto\Column as ColumnDto;
+use OCA\Tables\Model\SelectionOptions;
 use OCA\Tables\ResponseDefinitions;
 use OCA\Tables\Service\ValueObject\ViewColumnInformation;
 use OCA\Tables\Vendor\Symfony\Component\Uid\Uuid;
@@ -67,8 +68,8 @@ use ValueError;
  * @method setTextMaxLength(?int $textMaxLength)
  * @method getTextUnique(): bool
  * @method setTextUnique(?bool $textUnique)
- * @method getSelectionOptions(): string
- * @method setSelectionOptions(?string $selectionOptionsArray)
+ * @method string getSelectionOptions()
+ * @method setSelectionOptions(?string $selectionOptions)
  * @method setSelectionDefault(?string $selectionDefault)
  * @method getSelectionDefault(): ?string
  * @method getDatetimeDefault(): string
@@ -196,6 +197,8 @@ class Column extends EntitySuper implements JsonSerializable {
 		$this->addType('showUserStatus', 'boolean');
 
 		$this->addType('customSettings', 'string');
+
+		$this->addType('selectionOptions', 'string');
 	}
 
 	public static function isValidMetaTypeId(int $metaTypeId): bool {
@@ -252,8 +255,7 @@ class Column extends EntitySuper implements JsonSerializable {
 		$column->setNumberDecimals($data->getNumberDecimals());
 		$column->setNumberPrefix($data->getNumberPrefix() ?? '');
 		$column->setNumberSuffix($data->getNumberSuffix() ?? '');
-		$column->setSelectionOptions($data->getSelectionOptions());
-		$column->setSelectionDefault($data->getSelectionDefault());
+		$column->setSelectionOptionsCollection(SelectionOptions::createFromInputJsonString($data->getSelectionOptions() ?? '[]', $data->getSelectionDefault()));
 		$column->setDatetimeDefault($data->getDatetimeDefault());
 		$column->setUsergroupDefault($data->getUsergroupDefault());
 		$column->setUsergroupMultipleItems($data->getUsergroupMultipleItems());
@@ -279,18 +281,17 @@ class Column extends EntitySuper implements JsonSerializable {
 		$this->setUsergroup($json);
 	}
 
-	public function getSelectionOptionsArray(): array {
-		$options = $this->getSelectionOptions();
-		if ($options !== '' && $options !== null && $options !== 'null') {
-			return \json_decode($options, true);
-		} else {
-			return [];
-		}
+	public function getSelectionOptionsCollection(): SelectionOptions {
+		return SelectionOptions::createFromInputJsonString($this->getSelectionOptions() ?? '[]', $this->getSelectionDefault());
 	}
 
-	public function setSelectionOptionsArray(array $array):void {
-		$json = \json_encode($array);
-		$this->setSelectionOptions($json);
+	public function setSelectionOptionsCollection(SelectionOptions $selectionOptions): void {
+		$this->setSelectionOptions(json_encode($selectionOptions->jsonSerialize()));
+		$this->setSelectionDefault($selectionOptions->defaultSerialized());
+	}
+
+	public function getSelectionOptionsArray(): ?array {
+		return $this->getSelectionOptionsCollection()->jsonSerialize();
 	}
 
 	/**
--- a/lib/Db/Column.php
+++ b/lib/Db/Column.php
@@ -12,6 +12,7 @@ use OCA\Tables\Constants\ColumnType;
 use OCA\Tables\Dto\Column as ColumnDto;
 use OCA\Tables\ResponseDefinitions;
 use OCA\Tables\Service\ValueObject\ViewColumnInformation;
+use OCA\Tables\Vendor\Symfony\Component\Uid\Uuid;
 use ValueError;
 
 /**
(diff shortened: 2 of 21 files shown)


lib/Migration/Version2020Date20260513185340.php
```diff
@@ -0,0 +1,96 @@
...
+		if (!$schema->hasTable(self::TARGET_TABLE)) {
+			return null;
+		}
+
+		$columnsTable = $schema->getTable(self::TARGET_TABLE);
+		if (!$columnsTable->hasColumn(self::COL_UUID)) {
+			$columnsTable->addColumn(self::COL_UUID, Types::STRING, [
+				'notnull' => false,
+				'default' => null,
+				'length' => 36,
```
Koc, line 56: why not `binary (16)`?

.scoper-production-dependencies
```diff
@@ -7,3 +7,6 @@ psr/http-client
 psr/http-factory
 psr/http-message
 psr/simple-cache
+ramsey/uuid
+brick/math
+ramsey/collection
```
Koc: `symfony/uid` or `symfony/polyfill-uuid` has 0 dependencies and provides better performance BTW

blizzz: Out of habit… `symfony/uid` looks good on first sight, would transition towards it.

lib/Migration/Version2020Date20260513185340.php
blizzz: Mostly for simplicity. We would not need it for lookup either, rather in comparison on structural imports, and in JSON we would use the plain string anyways. We don't need it for sql query comparison either, as I imagine, but work on PHP level with it, so we don't need to handle conversion to or fro.

Although we have the `\OCP\DB\Types::BINARY` type, I have not seen it in use anywhere… assuming it works everywhere, it is still is easier to read or debug in a human-readable way without digging out conversion commands across DB engines.

blizzz:
> Add UUIDs to Tables or consider unnecessary
> Add UUIDs to Views or consider unnecessary

Would not implement it now as it is not required currently. If necessary for an auto-update of a table without specifying local ids, this can be always added alter on easily.

Koc:
I guess that uuid should be unique across table (we need add unique id for table_id+uuid).

Here use-cases that we need to support:

### Table update between instance
- a developer designs and evolves a table schema locally using UI
- exports the schema
- imports it into production
- later changes the schema locally again
- and wants to apply those schema updates to the already existing production table without recreating it from scratch

### Table duplocation a single NC instance
A second scenario is especially important for end users who do not have separate environments:
- a developer designs and evolves a table schema using UI
- exports the schema
- imports it into another table on same instance

blizzz:
> I guess that uuid should be unique across table (we need add unique id for table_id+uuid).
> 
> Here use-cases that we need to support:
> ### Table update between instance
> 
>     * a developer designs and evolves a table schema locally using UI
> 
>     * exports the schema
> 
>     * imports it into production
> 
>     * later changes the schema locally again
> 
>     * and wants to apply those schema updates to the already existing production table without recreating it from scratch

Yupp, this scenario is exactly what are having in mind.

> ### Table duplocation a single NC instance
> 
> A second scenario is especially important for end users who do not have separate environments:
> 
>     * a developer designs and evolves a table schema using UI
> 
>     * exports the schema
> 
>     * imports it into another table on same instance

This would also be covered, as importing always requires a target table. So having columns with the same UUID across different tables is intentional. Hence importing, so the plan, would only affect the columns on the targeted table.

lib/Db/Column.php
```diff
@@ -197,8 +199,23 @@ public static function isValidMetaTypeId(int $metaTypeId): bool {
...
+		$this->uuid = Uuid::v7()->toRfc4122();
+	}
+
+	protected function setUuid(string $uuid): void {
+		if ($this->uuid !== null) {
+			throw new \RuntimeException('This column already has a UUID, they are immutable');
+		}
+		$this->uuid = $uuid;
+	}
+
```
enjeck: `QBMapper::insert()` only writes fields returned by `getUpdatedFields()`. So we need to call `markFieldUpdated('uuid')` ?

lib/Db/Column.php
```diff
@@ -197,8 +199,23 @@ public static function isValidMetaTypeId(int $metaTypeId): bool {
...
 	}
 
+	private function assignUuid(): void {
+		if ($this->uuid !== null) {
+			throw new \RuntimeException('This column already has a UUID, they are immutable');
+		}
+		$this->uuid = Uuid::v7()->toRfc4122();
+	}
+
+	protected function setUuid(string $uuid): void {
```
enjeck: do we ever call this? and is it being `protected` useful. Cuz when the mapper reads a row, `Entity::fromRow()` runs (`Entity.php:50`):

```php
phpforeach ($row as $key => $value) {
    $prop = $instance->columnToProperty($key);
    $instance->setter($prop, [$value]);   // <-- calls the generic setter()
}
```

It calls the generic `setter()` method directly with the property name 'uuid'. i don't think it calls `setUuid()`? So the custom method and its RuntimeException are skipped?

lib/Db/Column.php
```diff
@@ -12,13 +12,15 @@
...
 use OCA\Tables\Service\ValueObject\ViewColumnInformation;
+use OCA\Tables\Vendor\Symfony\Component\Uid\Uuid;
 use ValueError;
 
 /**
  * @psalm-suppress PropertyNotSetInConstructor
  *
  * @psalm-import-type TablesColumn from ResponseDefinitions
  *
+ * @method string getUuid()
```
enjeck: Since the property is `?string $uuid`

```suggestion
@method ?string getUuid()
```

lib/Db/Column.php
```diff
@@ -261,6 +278,7 @@ public function setSelectionOptionsArray(array $array):void {
 	public function jsonSerialize(): array {
 		return [
 			'id' => $this->id,
+			'uuid' => $this->uuid,
```
enjeck, line 303: do we need to update `ResponseDefinitions.php`?

lib/Db/Column.php
enjeck, line 1: In `ColumnService.php`, we call `new Column()`. Do we have to do `assignUuid()` there?

lib/Migration/Version2020Date20260513185340.php
```diff
@@ -0,0 +1,96 @@
...
+				'comment' => 'UUIDv7 identifier to support structural updates across instances',
```
enjeck, line 57: so it will be nice to have a a `UNIQUE` index?

enjeck (commented):
some comments

lib/Db/Column.php
blizzz: Thanks for pointing it out. I remembered it wrongly and assumed it would call a customer setter if present.

PS.: it is vice versa :see_no_evil: https://github.com/nextcloud/server/blob/1ab09ec753f106661beadee1794b390ebec455dd/lib/public/AppFramework/Db/Entity.php#L169-L171

lib/Db/Column.php
blizzz: you refer to `import()`? I did not intend to cover import/export logic here, yet. But it should take over the UUID from the import data (if given) or assign a newly generated one, yes.

lib/Migration/Version2020Date20260513185340.php
blizzz: This is not the idea here, only within the same table. Several users may import (and eventually update) the same table from a template or different system, then you would have the same (or similar) column rows, but associated with different tables.

blizzz: of course that can be in combination with the table id. Would be nice to have he check in place, but there is also the question whether it is worth to have the table bookkeeping.

blizzz: added

blizzz:
@enjeck resolved all your comments. For you convenience, I have the essential changes still left in the unstashed fixup-PRs, and then there is just one additional commit. I would leave this now as the basework and would implement the create and update logic in a seperate PR.

blizzz:
Locally, all Playwright tests pass…

blizzz:
* [x] have to generate UUIDs also for selection options during migration

blizzz:
* [x] UUID information of multi-selection options is lost on export

blizzz:
> > Add UUIDs to Tables or consider unnecessary
> > Add UUIDs to Views or consider unnecessary
> 
> Would not implement it now as it is not required currently. If necessary for an auto-update of a table without specifying local ids, this can be always added alter on easily.

I am challenging my statement now, ~~but would rather do this in a follow up PR.~~ Added with latest commit.

lib/Db/Column.php
```diff
@@ -205,8 +211,41 @@ public static function isValidMetaTypeId(int $metaTypeId): bool {
...
+		}
+		$this->applyUuid(Uuid::v7()->toRfc4122());
+	}
+
+	private function setOrAssignUuid(?string $uuid): void {
+		if ($this->uuid !== null) {
+			throw new \RuntimeException('This column already has a UUID, they are immutable');
+		}
+		if ($uuid === null) {
+			$this->assignUuid();
```
luka-nextcloud: The function `assignUuid` seems redundant. `$this->applyUuid(Uuid::v7()->toRfc4122());` should be enough I think.

luka-nextcloud (commented):
@blizzz I have tested and it works great. I just have a minor comment.

lib/Db/Column.php
blizzz: Indeed, i have it in a simplified version in the Views class :+1:

luka-nextcloud (approved):

blizzz:
Thx, rebased and squashed everything.

PR #2599, opened 2026-05-14
style(icon): Use outline variant

AndyScherzinger:


AndyScherzinger:
tested and "just merged" given it is just a visual fix of using the correct icon (outline variant, not filled)

PR #2603, opened 2026-05-14
Test/noid/playwright test fix and hardening

AndyScherzinger:


enjeck (commented):
not sure we need this anymore. I already made some changes and the tests are now more robust

AndyScherzinger:
@enjeck I agree, let's close this one 👍

PR #2658, opened 2026-05-15
Fix does not contain filter 

### LINKED ISSUE #2597: Filter Does-not-Contains for groups and user column does not work
### Steps to reproduce

1. Create a column user
2. Make it possible to add multiple users 

Image

3. Create a view 
4. use a filter column users
5. select Does not contain userid

### Expected behavior

Filter works if there are multiple users 

### Actual behavior

Filter does not work it only works if the column defaults to one user. I thought it was possible but looks like i cannot see below....

### Tables app version

2.1.1

### Browser

150

### Client operating system

Windows

### Operating system

_No response_

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

seen in lib/Db/ColumnTypes/UsergroupColumnQB.php
```php
<?php

/**
 * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
 */

namespace OCA\Tables\Db\ColumnTypes;

use OCP\DB\QueryBuilder\IQueryBuilder;

class UsergroupColumnQB extends SuperColumnQB implements IColumnTypeQB {
	public function passSearchValue(IQueryBuilder $qb, string $unformattedSearchValue, string $operator, string $searchValuePlaceHolder): void {
		// TODO how to handle searching for multiple users/groups?
		$qb->setParameter($searchValuePlaceHolder, $unformattedSearchValue, IQueryBuilder::PARAM_STR);
	}
}
```

taiebot:
I unfortunately did not manage to test
>App "Tables" cannot be installed because it is not compatible with this version of the server.
Could not enable the tables app.

 that but the fix is so simple it should be easier to review and validate the fix
A row must be excluded if ANY of its usergroup values match the forbidden ones.

(value != forbidden_user) OR (value NOT IN forbidden_groups) OR (value NOT IN forbidden_circles)

Fix https://github.com/nextcloud/tables/issues/2597

### DIFF
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -485,7 +485,7 @@ class Row2Mapper {
 							$qb->expr()->eq('value_type', $qb->createNamedParameter(UsergroupType::CIRCLE, IQueryBuilder::PARAM_INT))
 						);
 					}
-					$filterExpression = $qb2->expr()->andX(...$filterExpressions);
+					$filterExpression = $qb2->expr()->orX(...$filterExpressions);
 					$includeDefault = false;
 
 					break;
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -467,52 +467,49 @@ class Row2Mapper {
 				$filterExpression = $qb->expr()->like('value', $qb->createNamedParameter('%' . $this->db->escapeLikeParameter($value) . '%', $paramType));
 				break;
 			case 'does-not-contain':
-				$filterExpressions = [];
 				if (is_array($value) && $column->getType() === Column::TYPE_USERGROUP) {
-					$filterExpressions[] = $qb2->expr()->andX(
-						$qb->expr()->neq('value', $qb->createNamedParameter($value[UsergroupType::USER])),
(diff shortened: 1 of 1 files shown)


lib/Db/Row2Mapper.php
```diff
@@ -485,7 +485,7 @@ private function getFilterExpression(IQueryBuilder $qb, Column $column, string $
 							$qb->expr()->eq('value_type', $qb->createNamedParameter(UsergroupType::CIRCLE, IQueryBuilder::PARAM_INT))
 						);
 					}
-					$filterExpression = $qb2->expr()->andX(...$filterExpressions);
```
enjeck, line 488: Switching to OR doesn't fix things. To see the issue:

Create:
```
Row A: users = [alice, bob]
Row B: users = [bob]
Row C: users = [alice]
```

Now filter for Does not contain `alice`.

The expected result is Row B only. With the PR’s OR: Row A and Row B can appear.

enjeck (changes requested):
doesn't seem to work

lib/Db/Row2Mapper.php
taiebot: Oh :-( so similar to current behaviour sorry did not manage to test ...

PR #2670, opened 2026-05-18
Fix does not contains function

### LINKED ISSUE #2597: Filter Does-not-Contains for groups and user column does not work

taiebot:
This time this is a tested fix.

This is entirely based from Claude AI. I tested multiple solutions from other AI and this one is the only one which works.

This correctly picks up empty cells as does not contain, if multiple users are present it does select the correct rows etc...
Tested only with groups and not circles

Will fix #2597 

Some screenshots 

image

image


enjeck (commented):
please add video showing the flow I described at https://github.com/nextcloud/tables/pull/2658#discussion_r3249316778

enjeck (commented):
please run `composer run cs:fix` too

enjeck (changes requested):
Now, in `tests/integration/base-query-count.txt` we need to change from 204554 to 207330

enjeck (approved):
thank you so much!

PR #2671, opened 2026-05-18
Fix Add 'does-not-contain' condition to Row2Mapper

taiebot:
Certainly an omission tested here and it works. If you try with current master an error occur if you try to use the filter on creator while with this changes the filter does work..

### DIFF
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -628,6 +628,8 @@ class Row2Mapper {
 				return $qb->expr()->like($columnName, $qb->createNamedParameter($this->db->escapeLikeParameter($value) . '%', $paramType));
 			case 'contains':
 				return $qb->expr()->like($columnName, $qb->createNamedParameter('%' . $this->db->escapeLikeParameter($value) . '%', $paramType));
+			case 'does-not-contain':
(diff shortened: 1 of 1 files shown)


lib/Db/Row2Mapper.php
```diff
@@ -608,6 +608,8 @@ private function getSqlOperator(string $operator, IQueryBuilder $qb, string $col
 				return $qb->expr()->like($columnName, $qb->createNamedParameter($this->db->escapeLikeParameter($value) . '%', $paramType));
 			case 'contains':
 				return $qb->expr()->like($columnName, $qb->createNamedParameter('%' . $this->db->escapeLikeParameter($value) . '%', $paramType));
+			case 'does-not-contain':
```
Koc, line 631: should we add `contains-item` as well that was added in https://github.com/nextcloud/tables/pull/2387?

taiebot: those filters are only for special columns (creator ,etc... ). The grey column visible in a view  they are created automatically on row creation so they will always contains something and cannot be null. i believe you might talk about the function getFilterExpression where contains-item could be added ?

enjeck (approved):
cool, thanks

PR #2456, opened 2026-04-11
📚❌ Add copy/delete row actions

AndyScherzinger:
* Alternative approach to @juliusknorr's #1713 
  * follow-up is opening the edit dialog on the original PR (see link above)
  * this PR opens a create dialog but pre-populated with the row values
* Any of the two would resolve one of the release stretch goals
* PR changes the edit Icon-Button into a 3-dot menu with edit/delete/copy + respective icons and updates the cypress tests

PR #2679, opened 2026-05-20
fix: smartpicker fix row actions and action button z-index

benjaminfrueh:
Fixes pre-existing issues with Smart Picker embedded tables I found while testing PR #2456.
- Row inline edit changes now show in the UI without reload
- Delete now shows in the UI without reload
- Sticky action button is now always clickable, added `z-index: 5` for consistency with `.sticky:first-child`


AndyScherzinger (approved):
works like a charm 🎉

enjeck (approved):

PR #2681, opened 2026-05-22
feat: add notifications

### LINKED ISSUE #1460: Add notifications
It would be nice to have a way to activate a notification if content is introduced. 

Even better, if new content is available on a view (which takes advantages of the filters), to generate a notification to specific contacts/groups

luka-nextcloud:
Resolves: #1460

Demo: 
[demo.webm](https://github.com/user-attachments/assets/50b9ef2b-61a9-4c28-8b07-a15662f030e5)

### DIFF
--- a/lib/Activity/ActivityManager.php
+++ b/lib/Activity/ActivityManager.php
@@ -9,13 +9,19 @@ declare(strict_types=1);
 namespace OCA\Tables\Activity;
 
 use OCA\Tables\AppInfo\Application;
+use OCA\Tables\Constants\ShareReceiverType;
+use OCA\Tables\Db\Column;
 use OCA\Tables\Db\ColumnMapper;
 use OCA\Tables\Db\Row2;
+use OCA\Tables\Db\Share;
 use OCA\Tables\Db\Table;
 use OCA\Tables\Db\TableMapper;
+use OCA\Tables\Db\View;
+use OCA\Tables\Db\ViewMapper;
 use OCA\Tables\Service\ShareService;
 use OCP\Activity\IEvent;
 use OCP\Activity\IManager;
+use OCP\Cache\CappedMemoryCache;
 use OCP\L10N\IFactory;
 use OCP\Server;
 use Psr\Log\LoggerInterface;
@@ -23,7 +29,9 @@ use Psr\Log\LoggerInterface;
 class ActivityManager {
 
 	public const TABLES_OBJECT_TABLE = 'tables_table';
+	public const TABLES_OBJECT_VIEW = 'tables_view';
 	public const TABLES_OBJECT_ROW = 'tables_row';
+	public const TABLES_OBJECT_COLUMN = 'tables_column';
 
 	public const SUBJECT_TABLE_CREATE = 'table_create';
 	public const SUBJECT_TABLE_UPDATE = 'table_update';
@@ -31,18 +39,44 @@ class ActivityManager {
 	public const SUBJECT_TABLE_UPDATE_DESCRIPTION = 'table_update_description';
 	public const SUBJECT_TABLE_DELETE = 'table_delete';
 
+	public const SUBJECT_VIEW_CREATE = 'view_create';
+	public const SUBJECT_VIEW_UPDATE = 'view_update';
+	public const SUBJECT_VIEW_UPDATE_TITLE = 'view_update_title';
+	public const SUBJECT_VIEW_UPDATE_DESCRIPTION = 'view_update_description';
+	public const SUBJECT_VIEW_DELETE = 'view_delete';
+
 	public const SUBJECT_ROW_CREATE = 'row_create';
 	public const SUBJECT_ROW_UPDATE = 'row_update';
 	public const SUBJECT_ROW_DELETE = 'row_delete';
+	public const SUBJECT_ROW_ASSIGN = 'row_assign';
+
+	public const SUBJECT_COLUMN_CREATE = 'column_create';
+	public const SUBJECT_COLUMN_UPDATE = 'column_update';
+	public const SUBJECT_COLUMN_DELETE = 'column_delete';
 
 	public const SUBJECT_IMPORT_FINISHED = 'import_finished';
+
+	public const SUBJECT_SHARE_CREATE = 'share_create';
+	public const SUBJECT_SHARE_UPDATE = 'share_update';
+	public const SUBJECT_SHARE_DELETE = 'share_delete';
+
+	public const RECEIVER_TYPE_UNKNOWN = 'unknown';
+	public const RECEIVER_ID_UNKNOWN = 'unknown';
+
+	public const EVENT_TYPE_SHARING = 'tables_sharing';
+	public const EVENT_TYPE_TABLE_ROW = 'tables_row_table';
+	public const EVENT_TYPE_VIEW_ROW = 'tables_row_view';
+	public const EVENT_TYPE_TABLE_COLUMN = 'tables_column_table';
+	public const EVENT_TYPE_VIEW_COLUMN = 'tables_column_view';
 
 	public function __construct(
 		private readonly IManager $manager,
 		private readonly IFactory $l10nFactory,
 		private readonly TableMapper $tableMapper,
+		private readonly ViewMapper $viewMapper,
 		private readonly ColumnMapper $columnMapper,
 		private readonly ShareService $shareService,
+		protected readonly CappedMemoryCache $cache,
 		private readonly ?string $userId,
 	) {
 	}
@@ -104,8 +138,14 @@ class ActivityManager {
 		if ($object instanceof Table) {
 			$objectTitle = $object->getTitle();
 			$table = $object;
+		} elseif ($object instanceof View) {
+			$objectTitle = $object->getTitle();
+			$table = $this->tableMapper->find($object->getTableId());
 		} elseif ($object instanceof Row2) {
 			$objectTitle = '#' . $object->getId();
+			$table = $this->tableMapper->find($object->getTableId());
+		} elseif ($object instanceof Column) {
+			$objectTitle = $object->getTitle();
 			$table = $this->tableMapper->find($object->getTableId());
 		} else {
 			Server::get(LoggerInterface::class)->error('Could not create activity entry for ' . $subject . '. Invalid object.', (array)$object);
@@ -119,23 +159,46 @@ class ActivityManager {
 		$eventType = 'tables';
 		$subjectParams = [
 			'author' => $author === null ? $this->userId : $author,
-			'table' => $table
+			'table' => $table,
+			'objectType' => $objectType,
 		];
+		if ($object instanceof View) {
+			$subjectParams['view'] = $object;
+		}
 		switch ($subject) {
 			// No need to enhance parameters since entity already contains the required data
 			case self::SUBJECT_TABLE_CREATE:
+			case self::SUBJECT_TABLE_UPDATE:
 			case self::SUBJECT_TABLE_DELETE:
+			case self::SUBJECT_VIEW_CREATE:
+			case self::SUBJECT_VIEW_UPDATE:
+			case self::SUBJECT_VIEW_DELETE:
 				break;
 			case self::SUBJECT_TABLE_UPDATE_DESCRIPTION:
+			case self::SUBJECT_VIEW_UPDATE_DESCRIPTION:
 				$subjectParams['after'] = $additionalParams['after'] ?? null;
 				break;
 			case self::SUBJECT_TABLE_UPDATE_TITLE:
+			case self::SUBJECT_VIEW_UPDATE_TITLE:
 				$subjectParams['before'] = $additionalParams['before'] ?? null;
 				break;
 			case self::SUBJECT_ROW_CREATE:
 			case self::SUBJECT_ROW_UPDATE:
 			case self::SUBJECT_ROW_DELETE:
+				$eventType = self::EVENT_TYPE_TABLE_ROW;
 				$subjectParams['row'] = $object;
+				break;
+			case self::SUBJECT_COLUMN_CREATE:
+			case self::SUBJECT_COLUMN_UPDATE:
+			case self::SUBJECT_COLUMN_DELETE:
+				$eventType = self::EVENT_TYPE_TABLE_COLUMN;
+				$subjectParams['column'] = $object;
+				break;
+			case self::SUBJECT_SHARE_CREATE:
+			case self::SUBJECT_SHARE_UPDATE:
+			case self::SUBJECT_SHARE_DELETE:
+				$eventType = self::EVENT_TYPE_SHARING;
+				$subjectParams['sharedWith'] = $this->buildSharedWithParam($additionalParams['share'] ?? []);
 				break;
 			case self::SUBJECT_IMPORT_FINISHED:
 				$subjectParams['importStats'] = $additionalParams['importStats'] ?? null;
@@ -145,28 +208,7 @@ class ActivityManager {
 		}
 
 		if ($subject === self::SUBJECT_ROW_UPDATE) {
-			$subjectParams['changeCols'] = [];
-			foreach ($additionalParams['before'] as $index => $colData) {
-				if ($additionalParams['after'][$index] === $colData) {
-					continue; // No change, skip
-				} else {
-					try {
-						$column = $this->columnMapper->find($colData['columnId']);
-						$subjectParams['changeCols'][] = [
-							'id' => $column->getId(),
-							'name' => $column->getTitle(),
-							'before' => $colData,
-							'after' => $additionalParams['after'][$index]
-						];
-					} catch (\Exception $e) {
-						Server::get(LoggerInterface::class)->error('Could not find column for activity entry.', [
-							'columnId' => $colData['columnId'],
-							'exception' => $e->getMessage()
-						]);
-						continue; // Skip if column not found
-					}
-				}
-			}
+			$subjectParams['changeCols'] = $this->extractChangeColsData($additionalParams);
 			unset($additionalParams['before'], $additionalParams['after']);
 		}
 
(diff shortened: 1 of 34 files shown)


src/modules/modals/sections/NotificationsSettings.vue
```diff
@@ -0,0 +1,263 @@
...
+		targetDescription() {
+			return this.isView
+				? t('tables', 'Choose which changes in this view should notify you.')
+				: t('tables', 'Choose which changes in this table should notify you.')
+		},
+		notificationSettings() {
+			return [
+				{
+					key: 'notify-row',
+					label: t('tables', 'Row changes'),
```
blizzz: 💄 less technical could be: `Content changes`?

src/modules/modals/sections/NotificationsSettings.vue
```diff
@@ -0,0 +1,263 @@
...
+					description: t('tables', 'Notify me when rows are created, updated or deleted.'),
+				},
+				{
+					key: 'notify-column',
+					label: t('tables', 'Column changes'),
```
blizzz: 💄 Less technical wording could be `Structure changes`?

src/modules/modals/sections/NotificationsSettings.vue
```diff
@@ -0,0 +1,263 @@
...
+			return this.isView
+				? t('tables', 'Choose which changes in this view should notify you.')
+				: t('tables', 'Choose which changes in this table should notify you.')
+		},
+		notificationSettings() {
+			return [
+				{
+					key: 'notify-row',
+					label: t('tables', 'Row changes'),
+					description: t('tables', 'Notify me when rows are created, updated or deleted.'),
```
blizzz: Here maybe `row values`, or just `values`, or `values in a row`.

lib/Activity/ActivityManager.php
blizzz, line 137: 💡 This method collected more than one hundred lines now. Maybe it can be split up?

lib/Activity/ActivityManager.php
```diff
@@ -181,85 +233,535 @@ private function createEvent($objectType, $object, $subject, $additionalParams =
...
+				'name' => $receiverName,
+				'type' => $receiverType,
+			];
+		}
+
+		return [
+			'id' => 'unknown',
+			'name' => 'unknown',
+			'type' => 'unknown',
+		];
```
blizzz: 💡  move `'unknown'` into a constant, it is used in more places.

lib/Activity/ActivityManager.php
```diff
@@ -181,85 +233,535 @@ private function createEvent($objectType, $object, $subject, $additionalParams =
...
+
+			$changeCols[] = [
+				'id' => $changeCol['id'],
+				'name' => $changeCol['name'],
+				'before' => $changeCol['before'],
+				'after' => $changeCol['after'],
+			];
+		}
+
+		// @Note: old data might not have the view parameter, in that case we assume the table is accessible and show all changed columns
```
blizzz: 🗨️ can you elaborate on that? Generally it sounds like a risky assumption.

lib/Activity/ActivityManager.php
```diff
@@ -181,85 +233,535 @@ private function createEvent($objectType, $object, $subject, $additionalParams =
...
+		}
+
+		return $visibleChangeCols;
+	}
+
+	/**
+	 * @param mixed $view
+	 * @return list<int>|null
+	 */
+	private function getViewColumnIds(mixed $view): ?array {
```
blizzz: 💡  could be simpler when the return value is always an array, not null.

lib/Controller/ConfigController.php
```diff
@@ -0,0 +1,50 @@
...
+	#[NoCSRFRequired]
+	public function getViewConfig(int $id): DataResponse {
+		$config = $this->configService->getViewConfig($id);
+		return new DataResponse($config);
+	}
+
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	public function setValue(string $key, mixed $value): DataResponse|NotFoundResponse {
+		$result = $this->configService->set($key, $value);
```
blizzz: 💡 I have the preference to check inputs (valid key and value types, e.g. through a value object) before passing them to the service. Others may disagree.

lib/Notification/NotificationHelper.php
```diff
@@ -0,0 +1,702 @@
...
+	) {
+	}
+
+	/**
+	 * @param mixed $object
+	 * @param mixed $subject
+	 * @param array<string, mixed> $additionalParams
+	 * @param mixed $author
+	 */
+	public function sendNotification($objectType, $object, $subject, $additionalParams = [], $author = null): void {
```
blizzz: type hints are missing, `mixed` can also be used if this is really the case. But `objectType` should be string, like `subject`, while `object` is either `Row2` or `Column`. And `auther` should be `?string?.

lib/Notification/NotificationHelper.php
```diff
@@ -0,0 +1,702 @@
...
+				'objectId' => is_object($object) && method_exists($object, 'getId') ? $object->getId() : null,
+				'subject' => $subject,
+				'error' => $e->getMessage(),
+			]);
+		}
+	}
+
+	/**
+	 * @param mixed $author
+	 */
```
blizzz, line 77: 🔥

lib/Notification/NotificationHelper.php
```diff
@@ -0,0 +1,702 @@
...
+	private function sendColumnNotification(Column $column, string $subject, $author): void {
```
blizzz:
```suggestion
	private function sendColumnNotification(Column $column, string $subject, ?string $author): void {
```

lib/Notification/NotificationHelper.php
```diff
@@ -0,0 +1,702 @@
...
+		$columnsCount = max(count($additionalParams['before']), count($additionalParams['after']));
+
+		for ($i = 0; $i < $columnsCount; $i++) {
+			$before = $additionalParams['before'][$i] ?? null;
+			$after = $additionalParams['after'][$i] ?? null;
+			$columnId = $before['columnId'] ?? $after['columnId'] ?? null;
+
+			if ($before === $after) {
+				continue; // No change, skip
+			} else {
```
blizzz: the else branch is not necessary, as the previous case is an early skip. So the pure `if` would be and we save some indent.

lib/Notification/NotificationHelper.php
```diff
@@ -0,0 +1,702 @@
...
+						'type' => $column->getType(),
+						'before' => $before,
+						'after' => $after
+					];
+				} catch (\Exception $e) {
+					Server::get(LoggerInterface::class)->error('Could not find column for activity entry.', [
+						'columnId' => $columnId,
+						'exception' => $e->getMessage()
+					]);
+					continue; // Skip if column not found
```
blizzz: should not be necessary as last instruction in the for loop?

lib/Notification/NotificationHelper.php
```diff
@@ -0,0 +1,702 @@
...
+			return null;
+		}
+
+		if (isset($this->columnCache[$columnId])) {
+			return $this->columnCache[$columnId];
+		}
+
+		try {
+			$column = $this->columnMapper->find($columnId);
+			$this->columnCache[$columnId] = $column;
```
blizzz: the columnMapper already has a cache

lib/Notification/NotificationHelper.php
```diff
@@ -0,0 +1,702 @@
...
+	): void {
+		$recipientSet = array_fill_keys(array_filter($recipients, static fn (mixed $id): bool => is_string($id) && $id !== ''), true);
+		$sent = [];
+
+		foreach ($assignedTargets as $target) {
+			foreach (array_unique($target['userIds']) as $receiverId) {
+				if ($receiverId === '' || $receiverId === $authorId || !isset($recipientSet[$receiverId])) {
+					continue;
+				}
+				if (!$this->configService->isNotifyEnabledForScope($receiverId, 'table', $tableId, 'notify-assigned')) {
```
blizzz: please use constants when values like `notify-assigned` are used more than once and especially around multiple places. This will save use from errors through type and simplifies refactoring and understanding of the code.

lib/Service/ConfigService.php
```diff
@@ -0,0 +1,132 @@
...
+				[$tableId, $configKey] = explode(':', $configKey, 2);
+				if (in_array($configKey, self::NOTIFY_CONFIG_KEYS)) {
+					return $this->userConfig->getValueBool($userId, Application::APP_ID, $key);
+				}
+		}
+
+		return false;
+	}
+
+	public function set($key, $value) {
```
blizzz: type hints

src/modules/modals/sections/NotificationsSettings.vue
```diff
@@ -0,0 +1,263 @@
...
+		return {
+			loading: false,
+			loadRequestId: 0,
+			config: this.getDefaultConfig(),
+			savingConfig: this.getDefaultConfig(),
+		}
+	},
+
+	computed: {
+		// ...mapState(useTablesStore, ['activeElement', 'isView']),
```
blizzz: leftover?

blizzz (changes requested):
Thanks  @luka-nextcloud! I did only little smoke testing, but that was already working. I left some comments around here in the code, the ones marked with 💡 are not blocking, but for consideration.

What I also did not check yet, question back for you, do you think this scales with a bigger number of users, or users in a group?

My impression is that in some places there is code that is quite similar, an almost-duplication. There might be potential to reduce this, but do not consider this critical.

blizzz:
P.S.: what I forgot to missed to add: :lipstick: better to not include the row ID in the text of the notification, it does not really help people and looks technical. Just link `row`, there is a direct link already anyway (at least for the creates and updates).

enjeck (commented):
not sure if should be done in another PR, but could we have notifications for asynchronous import? Right now, users never know exactly when import is done and it would be great

lib/Notification/NotificationHelper.php
```diff
@@ -0,0 +1,702 @@
...
+
+			if ($before === $after) {
+				continue; // No change, skip
+			} else {
+				try {
+					$column = $this->findColumnById((int)$columnId);
+					$changedColumns[] = [
+						'id' => $column->getId(),
+						'name' => $column->getTitle(),
+						'type' => $column->getType(),
```
enjeck: `$column` could be `null`, so trying to access these properties errors out. I don't think this is caught by following exception

src/shared/components/ActivityList.vue
```diff
@@ -81,14 +78,17 @@ export default {
 			}
 
 			let activities = response.data.ocs.data
-			if (this.objectType === 'tables_row') {
+			if (this.objectType === 'table') {
 				activities = activities.filter((activity) => {
-					return (activity.object_type === 'tables_row' && activity.object_id === this.objectId)
+					return (activity.subject_rich[1].table?.id === this.objectId.toString() && !activity.subject_rich[1].view)
```
enjeck: Is `subject_rich[1]` guaranteed to exist. Do we need to guard it with `subject_rich?.[1]?.table?.id` too?

lib/Notification/NotificationHelper.php
```diff
@@ -0,0 +1,702 @@
...
+				$this->notificationManager->notify($notification);
+			}
+		}
+	}
+
+	/**
+	 * @param array<string, mixed> $additionalParams
+	 * @param mixed $author
+	 */
+	private function sendRowNotification(Row2 $row, string $subject, array $additionalParams, $author): void {
```
enjeck, line 168: this function iterates `viewMapper->findAll($tableId)` twice (regular notifications, then assigned), calling `findSharedWithUserIds()` per view.
For each shared recipient it calls `row2Mapper->isRowInViewPresent()`, which executes `getWantedRowIds()`. So a single row edit on a table with V views and U shares can fire on the order of V×U filter evaluation queries. On busy/large tables this will have slow writes?

Can we fetch views once and reuse?

lib/Notification/Notifier.php
```diff
@@ -0,0 +1,300 @@
...
+use OCP\L10N\IFactory;
+use OCP\Notification\INotification;
+use OCP\Notification\INotifier;
+use OCP\Notification\UnknownNotificationException;
+
+class Notifier implements INotifier {
+	public function __construct(
+		protected readonly IFactory $l10nFactory,
+		private IURLGenerator $urlGenerator,
+		protected readonly IURLGenerator $url,
```
enjeck: why are we injecting `IURLGenerator` twice? Can we use once?

lib/Notification/NotificationHelper.php
enjeck, line 1: there is a lot of logic duplication between ActivityManager (view-context resolution, changed-column filtering, shared-user fan-out) and NotificationHelper. If i understand correctly, in both, we want to figure out which views/users are affected by a specific row/column? Can we reuse?

lib/Activity/Filter.php
```diff
@@ -60,7 +60,7 @@ public function getIcon(): string {
 	 * @since 11.0.0
 	 */
 	public function filterTypes(array $types): array {
-		return $types;
+		return array_merge($types, ['tables_sharing', 'tables_row_table', 'tables_row_view', 'tables_column_table', 'tables_column_view']);
```
enjeck: these constants are reused in various places, would be nice to have shared constants/variables

enjeck (commented):
some comments

luka-nextcloud:
> not sure if should be done in another PR, but could we have notifications for asynchronous import? Right now, users never know exactly when import is done and it would be great

I think the asynchronous import notification should be another PR since this PR is already big enough.
I've created a followup task here - https://github.com/nextcloud/tables/issues/2764

lib/Notification/NotificationHelper.php
blizzz: I meant the whole block could go, as the information is already in the method signature. Not blocking.

lib/Service/ConfigService.php
```diff
@@ -0,0 +1,149 @@
...
+			$this->userId = $user ? $user->getUID() : null;
+		}
+
+		return $this->userId;
+	}
+
+	/**
+	 * @return string|bool
+	 */
+	public function get(string $key) {
```
blizzz:
```suggestion
	public function get(string $key): string|bool {
```

lib/Service/ConfigService.php
```diff
@@ -0,0 +1,149 @@
...
+	/**
+	 * Set a configuration value for the current user
+	 *
+	 * @param string $key
+	 * @param mixed $value
+	 * @return void
+	 * @throws BadRequestError
+	 * @throws PermissionError
+	 */
+	public function set(string $key, mixed $value) {
```
blizzz:
```suggestion
	/**
	 * Set a configuration value for the current user
	 *
	 * @throws BadRequestError
	 * @throws PermissionError
	 */
	public function set(string $key, mixed $value): void {
```

lib/Notification/Notifier.php
```diff
@@ -0,0 +1,299 @@
...
+						$richParams['user']['name'] ?? '',
+						$richParams['view']['name'] ?? '',
+					])
+					: $l->t('A table row has been updated in table %2$s by %1$s.', [
+						$richParams['user']['name'] ?? '',
+						$richParams['table']['name'] ?? '',
+					]);
+				$subject = $hasViewContext
+					? $l->t('{user} has updated {row} in view {view}')
+					: $l->t('{user} has updated {row} in table {table}');
```
blizzz:
```suggestion
					? $l->t('{user} has updated this {row} in view {view}')
					: $l->t('{user} has updated this {row} in table {table}');
```

lib/Notification/Notifier.php
```diff
@@ -0,0 +1,299 @@
...
+						$richParams['user']['name'] ?? '',
+						$richParams['view']['name'] ?? '',
+					])
+					: $l->t('A table row has been deleted from table %2$s by %1$s.', [
+						$richParams['user']['name'] ?? '',
+						$richParams['table']['name'] ?? '',
+					]);
+				$subject = $hasViewContext
+					? $l->t('{user} has deleted the {row} in view {view}')
+					: $l->t('{user} has deleted the {row} in table {table}');
```
blizzz:
```suggestion
					? $l->t('{user} has deleted this {row} in view {view}')
					: $l->t('{user} has deleted this {row} in table {table}');
```

blizzz (commented):
When both content notifications and assignment notifications are enabled, it leads to two notifications. Imo this is not necessary and the assigned one is already sufficient. Can we only emit this one, if otherwise two would be send?

Image

And one thing design-wise. The settings are not so easy to find, and table managers do not know that they are not global settings:

Image

cc @marcoambrosini

luka-nextcloud:
@blizzz 
> When both content notifications and assignment notifications are enabled, it leads to two notifications. Imo this is not necessary and the assigned one is already sufficient. Can we only emit this one, if otherwise two would be send?

I think we should emit 2 notifications since they are 2 separate configs. BTW, a row changed event might include multiple changed fields.

> And one thing design-wise. The settings are not so easy to find, and table managers do not know that they are not global settings:

We should wait for designer advice from @marcoambrosini

blizzz (commented):
Good for me then, but expecting a word from design on that aspect.

src/modules/modals/sections/NotificationsSettings.vue
```diff
@@ -0,0 +1,262 @@
...
+			return this.isView
+				? t('tables', 'Choose which changes in this view should notify you.')
+				: t('tables', 'Choose which changes in this table should notify you.')
+		},
+		notificationSettings() {
+			return [
+				{
+					key: 'notify-row',
+					label: t('tables', 'Content changes'),
+					description: t('tables', 'Notify me when row values are created, updated or deleted.'),
```
marcoambrosini:
```suggestion
					description: t('tables', 'When row values are created, updated or deleted.'),
```

src/modules/modals/sections/NotificationsSettings.vue
```diff
@@ -0,0 +1,262 @@
...
+				},
+				{
+					key: 'notify-column',
+					label: t('tables', 'Structure changes'),
+					description: t('tables', 'Notify me when columns are created, updated or deleted.'),
```
marcoambrosini:
```suggestion
					description: t('tables', 'When columns are created, updated or deleted.'),
```

src/modules/modals/sections/NotificationsSettings.vue
```diff
@@ -0,0 +1,262 @@
...
+				},
+				{
+					key: 'notify-assigned',
+					label: t('tables', 'Assigned'),
+					description: t('tables', 'Notify me when I am assigned in a user group field.'),
```
marcoambrosini:
```suggestion
					description: t('tables', 'When I you are added to a user or group field.'),
```

src/modules/modals/sections/NotificationsSettings.vue
```diff
@@ -0,0 +1,262 @@
+<!--
+	- SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+	- SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<div v-if="elementId" class="notifications-settings">
+		<p>
```
marcoambrosini: No need for this description line

marcoambrosini (approved):
Very nice, only some string fixes and one redundant description removal :)

blizzz (approved):
Then let's take it as is. Thanks @luka-nextcloud!

marcoambrosini:
Sorry for missing your comment @blizzz 

About the double notification, I agree with youy that that's redundant. Let's always show the most specific one.

About the notification settings, I think that the text is pretty explicit in communicating the fact that these settings affect only the user changing them. 

Something else we could do to make it more obvious is to introduce a "Personal" heading similar to Talk, but given that "notifications" is the only personal setting for now, I would leave it as is and only introduce it when and if we have more.  
  
Screenshot 2026-07-21 at 13 13 34

PR #2710, opened 2026-06-07
enh: update max-version to NC 35

enjeck:
so things work on master


AndyScherzinger (approved):

AndyScherzinger:
And make life easier during development. For now I locally patched my checkout but having master target 35 is way better 👍

PR #2711, opened 2026-06-07
🕵️ Add sharereview support

AndyScherzinger:
* pre-requisite: https://github.com/nextcloud/server/pull/61543 merged

### 🏁 Checklist
 
- [x] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/lib/ShareReview/ShareReviewSource.php
+++ b/lib/ShareReview/ShareReviewSource.php
@@ -1,0 +1,252 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Tables\ShareReview;
+
+use OCA\Tables\Db\ContextMapper;
+use OCA\Tables\Db\ShareMapper;
+use OCA\Tables\Db\TableMapper;
+use OCA\Tables\Db\ViewMapper;
+use OCA\Tables\Service\ShareService;
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\DB\Exception;
+use OCP\EventDispatcher\IEventDispatcher;
+use OCP\IL10N;
+use OCP\Share\IShare;
+use OCP\Share\ShareReview\Events\ShareReviewAccessCheckEvent;
+use OCP\Share\ShareReview\IShareReviewSource;
+use OCP\Share\ShareReview\ShareReviewEntry;
+use OCP\Share\ShareReview\ShareReviewPermission;
+use Psr\Log\LoggerInterface;
+
+class ShareReviewSource implements IShareReviewSource {
+
+	private const NODE_TYPE_TABLE = 'table';
+	private const NODE_TYPE_VIEW = 'view';
+	private const NODE_TYPE_CONTEXT = 'context';
+
+	private const RECEIVER_TYPE_LINK = 'link';
+
+	public const PERMISSION_READ = 'tables:read';
+	public const PERMISSION_UPDATE = 'tables:update';
+	public const PERMISSION_CREATE = 'tables:create';
+	public const PERMISSION_DELETE = 'tables:delete';
+	public const PERMISSION_MANAGE = 'tables:manage';
+
+	/** @var array<string, ShareReviewPermission>|null */
+	private ?array $permissionCatalog = null;
+
+	public function __construct(
+		private ShareMapper $shareMapper,
+		private TableMapper $tableMapper,
+		private ViewMapper $viewMapper,
+		private ContextMapper $contextMapper,
+		private IL10N $l10n,
+		private LoggerInterface $logger,
+		private readonly ShareService $shareService,
+		private readonly IEventDispatcher $eventDispatcher,
+	) {
+	}
+
+	public function getName(): string {
+		return $this->l10n->t('Tables');
+	}
+
+	/**
+	 * @return list<ShareReviewEntry>
+	 */
+	public function getShares(): array {
+		try {
+			$nodeIdsByType = $this->shareMapper->findSharedNodeIdsByType();
+		} catch (Exception $e) {
+			$this->logger->error('Tables ShareReview: failed to fetch shared node IDs: {message}', ['message' => $e->getMessage()]);
+			return [];
+		}
+
+		try {
+			$tableNames = $this->tableMapper->findIdToTitleMap($nodeIdsByType[self::NODE_TYPE_TABLE] ?? []);
+		} catch (Exception $e) {
+			$this->logger->error('Tables ShareReview: failed to fetch table names: {message}', ['message' => $e->getMessage()]);
+			$tableNames = [];
+		}
+
+		try {
+			$viewNames = $this->viewMapper->findIdToTitleMap($nodeIdsByType[self::NODE_TYPE_VIEW] ?? []);
+		} catch (Exception $e) {
+			$this->logger->error('Tables ShareReview: failed to fetch view names: {message}', ['message' => $e->getMessage()]);
+			$viewNames = [];
+		}
+
+		try {
+			$contextNames = $this->contextMapper->findIdToNameMap($nodeIdsByType[self::NODE_TYPE_CONTEXT] ?? []);
+		} catch (Exception $e) {
+			$this->logger->error('Tables ShareReview: failed to fetch application names: {message}', ['message' => $e->getMessage()]);
+			$contextNames = [];
+		}
+
+		$formatted = [];
+		try {
+			foreach ($this->shareMapper->findAllRaw() as $share) {
+				$formatted[] = $this->buildEntry($share, $tableNames, $viewNames, $contextNames);
+			}
+		} catch (Exception $e) {
+			$this->logger->error('Tables ShareReview: failed to fetch shares: {message}', ['message' => $e->getMessage()]);
+			return [];
+		}
+		return $formatted;
+	}
+
+	public function deleteShare(string $shareId): bool {
+		if (!is_numeric($shareId)) {
+			return false;
+		}
+
+		$event = new ShareReviewAccessCheckEvent('Tables', $shareId);
+		$this->eventDispatcher->dispatchTyped($event);
+
+		if (!$event->isHandled() || !$event->isGranted()) {
+			return false;
+		}
+
+		try {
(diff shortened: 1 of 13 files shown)


lib/ShareReview/ShareReviewSource.php
```diff
@@ -0,0 +1,235 @@
...
+		$tableNames = $this->fetchNames(self::TABLES_TABLE, 'title', $idsByType[self::NODE_TYPE_TABLE]);
+		$viewNames = $this->fetchNames(self::VIEWS_TABLE, 'title', $idsByType[self::NODE_TYPE_VIEW]);
+		$contextNames = $this->fetchNames(self::CONTEXTS_TABLE, 'name', $idsByType[self::NODE_TYPE_CONTEXT]);
+
+		$appName = $this->getName();
+		$formatted = [];
+		foreach ($rawShares as $share) {
+			$formatted[] = [
+				'id' => (int)$share['id'],
+				'app' => $appName,
```
enjeck: We don't need this key. The consumer (`ShareService::getAppShares()`) already injects the app name. 
Also, not required to be specified according to https://github.com/nextcloud/sharereview#3-implement-the-source

lib/ShareReview/ShareReviewSource.php
```diff
@@ -0,0 +1,235 @@
...
+				->where($qb->expr()->eq('id', $qb->createNamedParameter((int)$shareId, IQueryBuilder::PARAM_INT)));
+			$deleted = $qb->executeStatement() > 0;
+		} catch (Exception $e) {
+			$this->logger->error('Tables ShareReview: failed to delete share {id}: {message}', ['id' => $shareId, 'message' => $e->getMessage()]);
+			return false;
+		}
+
+		if ($deleted) {
+			$this->deleteContextNavigation((int)$shareId);
+		}
```
enjeck, line 84: is it possible to use Tables' `ShareService::delete()` instead?

lib/ShareReview/ShareReviewSource.php
```diff
@@ -0,0 +1,235 @@
...
+	 * @param array<int, string> $viewNames
+	 * @param array<int, string> $contextNames
+	 */
+	private function resolveObjectName(array $share, array $tableNames, array $viewNames, array $contextNames): string {
+		$nodeId = (int)$share['node_id'];
+		return match($share['node_type']) {
+			self::NODE_TYPE_TABLE => ($tableNames[$nodeId] ?? "Table $nodeId") . ' (Table)',
+			self::NODE_TYPE_VIEW => ($viewNames[$nodeId] ?? "View $nodeId") . ' (View)',
+			self::NODE_TYPE_CONTEXT => ($contextNames[$nodeId] ?? "Context $nodeId") . ' (Context)',
+			default => "Unknown $nodeId",
```
enjeck: we should probably localize these texts

lib/ShareReview/ShareReviewSource.php
```diff
@@ -0,0 +1,235 @@
...
+		};
+	}
+
+	private function mapReceiverType(string $receiverType): int {
+		return match($receiverType) {
+			'user' => IShare::TYPE_USER,
+			'group' => IShare::TYPE_GROUP,
+			'link' => IShare::TYPE_LINK,
+			'circle' => IShare::TYPE_CIRCLE,
+			default => IShare::TYPE_USER,
```
enjeck: silently mislabels bad data. Could have a warning too, if we must

AndyScherzinger: Added logging via https://github.com/nextcloud/tables/pull/2711/changes/d2e8862fbfbe94f06de52c4928c10ee472c644c9

lib/ShareReview/ShareReviewSource.php
AndyScherzinger: Addressed via https://github.com/nextcloud/tables/pull/2711/changes/c73d566694e63c751fb0aafddf8e24f4d686b65d

lib/ShareReview/ShareReviewSource.php
AndyScherzinger: Addressed via https://github.com/nextcloud/tables/pull/2711/commits/cbf2ce997c681a94516550d7abdf7f4c817f61d8

lib/ShareReview/ShareReviewSource.php
AndyScherzinger: Addressed via https://github.com/nextcloud/tables/pull/2711/commits/94ce8538869122a1bff5671131a521db08c230a4

enjeck (commented):
To be sure, it's intentional to be able to delete shares from resources that you don't own/manage? And that usually-secret public share tokens are listed too?

lib/ShareReview/ShareReviewSource.php
```diff
@@ -0,0 +1,259 @@
...
+use OCP\IL10N;
+use OCP\Share\IShare;
+use Psr\Log\LoggerInterface;
+
+class ShareReviewSource implements ISource {
+
+	private const SHARE_TABLE = 'tables_shares';
+	private const TABLES_TABLE = 'tables_tables';
+	private const VIEWS_TABLE = 'tables_views';
+	private const CONTEXTS_TABLE = 'tables_contexts_context';
```
blizzz: I do not like that database resource names leak from their actual place, which are the mappers. 

Alternative: make it a public class constant in the respective mapper.

lib/ShareReview/ShareReviewSource.php
```diff
@@ -0,0 +1,259 @@
...
+				->orderBy('id', 'ASC');
+			$result = $qb->executeQuery();
+			$rows = $result->fetchAll();
+			$result->closeCursor();
+			return $rows;
+		} catch (Exception $e) {
+			$this->logger->error('Tables ShareReview: failed to fetch shares: {message}', ['message' => $e->getMessage()]);
+			return [];
+		}
+	}
```
blizzz: Leaks responsibility, should be provided as a method on (Service and) Mapper level.

A think to consider: can be implemented as Generator (we do it to seldomly) for a smaller memory footprint. Especially since this one reads all shares. Think of big instances. In worst case (not sure it is likely, but depends) this can ran into a memory exhaustion – but also on the callee side if really everything is consumed at once.

lib/ShareReview/ShareReviewSource.php
```diff
@@ -0,0 +1,259 @@
...
+					->where($qb->expr()->in('id', $qb->createNamedParameter($chunk, IQueryBuilder::PARAM_INT_ARRAY)));
+				$result = $qb->executeQuery();
+				$rows = $result->fetchAll();
+				$result->closeCursor();
+				foreach ($rows as $row) {
+					$map[(int)$row['id']] = (string)$row[$nameColumn];
+				}
+			} catch (Exception $e) {
+				$this->logger->error('Tables ShareReview: failed to fetch names from {table}: {message}', ['table' => $table, 'message' => $e->getMessage()]);
+			}
```
blizzz: as mentioned, database queries should be done in the respectives mappers and exposed perhaps/ideally via Service.

lib/ShareReview/ShareReviewSource.php
```diff
@@ -0,0 +1,259 @@
...
+		$rawShares = $this->fetchAllShares();
+
+		$idsByType = $this->groupIdsByNodeType($rawShares);
+		$tableNames = $this->fetchNames(self::TABLES_TABLE, 'title', $idsByType[self::NODE_TYPE_TABLE]);
+		$viewNames = $this->fetchNames(self::VIEWS_TABLE, 'title', $idsByType[self::NODE_TYPE_VIEW]);
+		$contextNames = $this->fetchNames(self::CONTEXTS_TABLE, 'name', $idsByType[self::NODE_TYPE_CONTEXT]);
+
+		$formatted = [];
+		foreach ($rawShares as $share) {
+			$formatted[] = [
```
blizzz: Ideally this is not a random associated array, but a specific class (can be a DTO/data containers without further logic). This is way more maintainable and also has a better memory footprint. 

A factory class or callback could be provided by the ShareReview app when calling, then you have it more formalized/API-like as well.

lib/ShareReview/ShareReviewSource.php
```diff
@@ -0,0 +1,259 @@
...
+		$nodeId = (int)$share['node_id'];
+		$nodeType = (string)$share['node_type'];
+		if ($nodeType === self::NODE_TYPE_TABLE) {
+			return $this->l10n->t('%s (Table)', [$tableNames[$nodeId] ?? $this->l10n->t('Table %s', [$nodeId])]);
+		}
+		if ($nodeType === self::NODE_TYPE_VIEW) {
+			return $this->l10n->t('%s (View)', [$viewNames[$nodeId] ?? $this->l10n->t('View %s', [$nodeId])]);
+		}
+		if ($nodeType === self::NODE_TYPE_CONTEXT) {
+			return $this->l10n->t('%s (Context)', [$contextNames[$nodeId] ?? $this->l10n->t('Context %s', [$nodeId])]);
```
blizzz: `Application` is the user facing term, `Context` is only the internal-technical name.

lib/ShareReview/ShareReviewSource.php
```diff
@@ -0,0 +1,259 @@
...
+
+		if (!$deleted) {
+			$this->logger->warning('Tables ShareReview: share {id} not found', ['id' => $shareId]);
+			return false;
+		}
+
+		try {
+			$this->contextNavigationMapper->deleteByShareId((int)$shareId);
+		} catch (Exception $e) {
+			$this->logger->error('Tables ShareReview: failed to clean up context navigation for share {id}: {message}', ['id' => $shareId, 'message' => $e->getMessage()]);
```
blizzz: This will be more rule than exception, as we can expect this being called for *Tables* and *Views* way more often than for *Contexts*. One or more entries will only be against shared contexts.

Hence it would result in confusing, at worst also spammy, log messages.

blizzz (changes requested):

AndyScherzinger:
@enjeck regarding:

> To be sure, it's intentional to be able to delete shares from resources that you don't own/manage? And that usually-secret public share tokens are listed too?

Yes _BUT_ I will change the implementation in the way that it uses an event mechanism like in other places with similar calling patterns, to ensure the user would have that permission as checked then by the share review app. SO I reset the PR to draft, to reflect the need for further improvements.

blizzz (approved):
everything was addressed 👍

blizzz:
rebased and resolved the conflict

PR #2713, opened 2026-06-08
fix: Chunk IN clause to <1000 items for oracle compatibility

AndyScherzinger:
Assisted-by: Claude Code:claude-sonnet-4-6

### 🏁 Checklist
 
- [x] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -197,6 +197,24 @@ class Row2Mapper {
 	 * @throws InternalError
 	 */
 	private function getRows(array $rowIds, array $columnIds): array {
+		if (empty($rowIds) || empty($columnIds)) {
+			return [];
+		}
+
+		$allRows = [];
+		foreach (array_chunk($rowIds, 999) as $rowIdChunk) {
+			$allRows = array_merge($allRows, $this->getRowsChunk($rowIdChunk, $columnIds));
+		}
+		return $allRows;
+	}
+
+	/**
+	 * @param array $rowIds
+	 * @param array $columnIds
+	 * @return Row2[]
+	 * @throws InternalError
+	 */
+	private function getRowsChunk(array $rowIds, array $columnIds): array {
 		$qb = $this->db->getQueryBuilder();
 
 		$qbSqlForColumnTypes = null;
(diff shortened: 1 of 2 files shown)


lib/Db/Row2Mapper.php
```diff
@@ -197,6 +197,24 @@ public function findAll(array $showColumnIds, int $tableId, ?int $limit = null,
 	 * @throws InternalError
 	 */
 	private function getRows(array $rowIds, array $columnIds): array {
+		if (empty($rowIds) || empty($columnIds)) {
+			return [];
+		}
+
+		$allRows = [];
+		foreach (array_chunk($rowIds, 997) as $rowIdChunk) {
```
blizzz: Doubtful about this hardcoded number. The query consumes two arrays. ColumnId is unlikely to have more than 1k entries, but definitely more than 3. So should be `1k - count($columnIds)`. Then I am not sure how Oracle treats the join, but reducing the other 2 seems safe then: `1k - count($columnIds)` -2.

The bad thing is that the `-2` cannot be computed from here, as it is a detail of the called method.

AndyScherzinger: I think the triggered query has 2 INs while each can have up to 1K items, so we could probably chunk by 1000 if you like. The columns are not relevant for chunking as you said and currently not chunked, just the row IDs, no?

Else if unsure about how Oracle treats multiple INs in a query and if all of them need to be max 1k in total than we could follow your approach of `1k - count($columnIds) -2`

https://stackoverflow.com/questions/2401066/oracle-sql-how-to-use-more-than-1000-items-inside-an-in-clause suggests that many INS are fine as long each is max 1K items.

blizzz: They are passed here: https://github.com/nextcloud/tables/blob/f79864d9b8aa1952e8fa2182edf9460fea70f9ed/lib/Db/Row2Mapper.php#L237-L238

IIRC the 1k is a global threshold for the query. But I might mix it up with another DB. Quick search shows a global limit of 32767 arguments in Postgres, while Oracle should have the 1k limit per IN clause.

AndyScherzinger: Updated the info based on your updated fix via https://github.com/nextcloud/tables/pull/2713/commits/b37f358fec4d2b714d335c863f15bfad6bb98881

blizzz (approved):

PR #2728, opened 2026-06-16
apply remove share for users/teams/groups

samin-z:
IF user/team/groups is removed, all the connections with tables share is also removed.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -42,7 +42,7 @@ use OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent;
 use OCP\Collaboration\Reference\RenderReferenceEvent;
 use OCP\Collaboration\Resources\LoadAdditionalScriptsEvent;
 use OCP\DB\Events\AddMissingIndicesEvent;
-use OCP\Group\Events\BeforeGroupDeletedEvent;
+use OCP\Group\Events\GroupDeletedEvent;
 use OCP\User\Events\BeforeUserDeletedEvent;
 use OCP\User\Events\UserDeletedEvent;
 use Psr\Container\ContainerInterface;
@@ -92,7 +92,7 @@ class Application extends App implements IBootstrap {
 		$context->registerEventListener(TableOwnershipTransferredEvent::class, WhenTableTransferredAuditLogListener::class);
 		$context->registerEventListener(AddMissingIndicesEvent::class, AddMissingIndicesListener::class);
 		$context->registerEventListener(UserDeletedEvent::class, ReceiverCleanupListener::class);
-		$context->registerEventListener(BeforeGroupDeletedEvent::class, ReceiverCleanupListener::class);
+		$context->registerEventListener(GroupDeletedEvent::class, ReceiverCleanupListener::class);
 		$context->registerEventListener(CircleDestroyedEvent::class, ReceiverCleanupListener::class);
 
 		$context->registerSearchProvider(SearchTablesProvider::class);
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -9,6 +9,7 @@ namespace OCA\Tables\AppInfo;
 
 use Exception;
 use OCA\Analytics\Datasource\DatasourceEvent;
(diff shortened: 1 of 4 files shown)


lib/Listener/ReceiverCleanupListener.php
```diff
@@ -0,0 +1,54 @@
...
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\Group\Events\BeforeGroupDeletedEvent;
+use OCP\User\Events\UserDeletedEvent;
+
+/** @template-implements IEventListener<Event|UserDeletedEvent|BeforeGroupDeletedEvent|CircleDestroyedEvent> */
+class ReceiverCleanupListener implements IEventListener {
+	public function __construct(
+		private ShareMapper $shareMapper,
+		private CirclesManager $circlesManager,
```
enjeck: `CirclesManager` is a hard constructor dependency, but Circles is an optional app. When Circles is not available, this might crash?. Should use `CircleHelper.php` instead

lib/Listener/ReceiverCleanupListener.php
```diff
@@ -0,0 +1,54 @@
...
+	) {
+	}
+
+	public function handle(Event $event): void {
+		if ($event instanceof UserDeletedEvent) {
+			$this->cleanupByParticipant('user', $event->getUser()->getUID());
```
enjeck: instead of hardcoding `user`, `group`, `circle`, use the constants e.g `ShareReceiverType::USER`

lib/Listener/ReceiverCleanupListener.php
```diff
@@ -0,0 +1,54 @@
...
+
+		try {
+			$singleId = $this->circlesManager
+				->getFederatedUser($gid, Member::TYPE_GROUP)
+				->getSingleId();
+		} catch (\Throwable) {
+			return;
+		}
+
+		$this->cleanupByParticipant('group', $singleId);
```
enjeck: should this be 'circle' instead?

lib/Listener/ReceiverCleanupListener.php
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	private function cleanupByParticipant(string $type, string $participant): void {
+		$this->shareMapper->deleteByReceiver($participant, $type);
```
enjeck: could throw an `Exception`, and nothing here catches it

enjeck (commented):
not sure if ready for review yet, but I have some preliminary comments

lib/Listener/ReceiverCleanupListener.php
samin-z: ok i see your point, circles is optional in tables and hard injecting CirclesManager does not match our existing CircleHelper pattern. i'll update it.

lib/Listener/ReceiverCleanupListener.php
samin-z: if a group is added to table, in tables_shares table for receiver_type we have 'group'

samin-z:
@enjeck i decided to simplify the logic and do a follow-up ticket on the case for checking groups before adding and also removing them if exist.

lib/AppInfo/Application.php
```diff
@@ -87,6 +91,9 @@ public function register(IRegistrationContext $context): void {
 		$context->registerEventListener(RowDeletedEvent::class, WhenRowDeletedAuditLogListener::class);
 		$context->registerEventListener(TableOwnershipTransferredEvent::class, WhenTableTransferredAuditLogListener::class);
 		$context->registerEventListener(AddMissingIndicesEvent::class, AddMissingIndicesListener::class);
+		$context->registerEventListener(UserDeletedEvent::class, ReceiverCleanupListener::class);
+		$context->registerEventListener(BeforeGroupDeletedEvent::class, ReceiverCleanupListener::class);
```
blizzz: Group deletion may still be aborted at this stage. Like with the other events, `GroupDeletedEvent` should be used here instead, right?

samin-z: good catch, this was the case with the previous solution, i missed to update this one.

blizzz (approved):

Issue #2596, opened 2026-05-12
Embedded tables some column overflow on header


Aveyron-RetD:
### Steps to reproduce

1. Some columns do overflow on embeddded table

### Expected behavior

No overflow on embedded column

### Actual behavior

<img width="753" height="189" alt="Image" src="https://github.com/user-attachments/assets/585b37f0-57de-4ff8-be37-546e74cd94ce" />

### Tables app version

2.1.1

### Browser

150

### Client operating system

Windows

### Web server

None

### PHP engine version

None

### Database

None

### Additional info

Note to myself for testing at home  and create a pull request

Fix suggested by AI

🛠️ How to fix it globally

You can stabilize the header layout by adding this rule in TableHeader.vue:
scss

th {
    white-space: normal;
    min-height: 48px;   // ensures consistent header height
    overflow: hidden;   // hides filter chips or icons that overflow
}

PR #2744, opened 2026-06-17
Fix Embed table full scroll 

taiebot:
* Fix remove sticky and scroll everything 

Widget is too small to let headers stick so let headers scroll, this as well solves issue with overlay on the headers by simplifying the logic. The only issue is the three dots on the right but they are accessible if the user goes back to the top of the table

* Fix issue https://github.com/nextcloud/tables/issues/2596

---------

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist
 
- [ ] 📅 Milestone is set
- [X] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

## 🤖 AI (if applicable)

- [X] The content of this PR was partly or fully generated using AI

### DIFF
--- a/src/views/ContentReferenceWidget.vue
+++ b/src/views/ContentReferenceWidget.vue
@@ -246,70 +246,70 @@ export default {
 </script>
 <style lang="scss" scoped>
 
-	.tables-content-widget {
-		min-height: max(50vh, 200px);
-		height: 50vh;
-		overflow: scroll;
+.tables-content-widget {
+    min-height: max(50vh, 200px);
+    height: 50vh;
+    overflow: auto;
 
-		& .header {
-			position: sticky;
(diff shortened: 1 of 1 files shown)


enjeck (commented):
In the "after" screenshot, I don't see the title?

enjeck:
Not sure how to reproduce the issue since without these changes, it renders for me well. Here, I'm embedding within Notes:

Screenshot 2026-06-20 at 07 36 47

Plus, the before and after are of different tables, so doesn't appropriately reflect the fix. I also don't agree with removing the header

enjeck:
superseded by https://github.com/nextcloud/tables/pull/2868

PR #2763, opened 2026-06-23
Refactor: Extract `getTableCellComponent` helper for better reusability

Koc:
This is needed for #2271 as we should be able to render cells in cells :exploding_head: 

### 🏁 Checklist
 
- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/src/shared/components/ncTable/partials/TableCell.vue
+++ b/src/shared/components/ncTable/partials/TableCell.vue
@@ -1,0 +1,82 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<component :is="cellComponent"
+		:column="column"
+		:row-id="rowId"
+		:value="value"
+		:element-id="elementId"
+		:is-view="isView"
+		:can-edit="canEdit" />
+</template>
+
+<script>
+import { ColumnTypes } from './../mixins/columnHandler.js'
+
+import TableCellHtml from './TableCellHtml.vue'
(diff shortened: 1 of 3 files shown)


src/shared/components/ncTable/partials/TableCell.vue
```diff
@@ -0,0 +1,83 @@
...
+			default: () => {},
+		},
+		rowId: {
+			type: Number,
+			default: null,
+		},
+		value: {
+			type: [String, Number, Array, Object, Boolean, null],
+			default: null,
+		},
```
enjeck: null is not a valid constructor here. . Either drop null from the array, or omit type entirely to allow any value, matching how `cellEditMixin` declares it (`value: { required: true }`)

Koc: done

src/shared/components/ncTable/partials/TableCell.vue
```diff
@@ -0,0 +1,82 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<component :is="cellComponent"
```
AIlkiv: Would it be possible to avoid introducing an additional TableCell component here? This adds one extra Vue component instance for every cell, which could become noticeable for large tables.

We could instead move getTableCell() (or the component mapping) into a regular JS module and import it where needed. This would provide the same reusability without adding an extra component layer to every cell.

Koc: fixed

enjeck (approved):
thanks!

Issue #731, opened 2023-12-04
Table total rows not updated upon add/delete row


whisperdancer:
### Steps to reproduce

1. Open a table
2. Add or delete a row
3. Hover over the table name in left pane and it doesn't change to reflect the change

### Expected behavior

After adding or deleting a row in a table, when hovering over the table name I expect the total rows to update to reflect the change.

### Actual behavior

1. Open a table
2. Add or delete a row
3. Hover over the table name in left pane and the total rows doesn't change to reflect the change

### Tables app version

0.6.4

### Browser

firefox 120.0.1

### Client operating system

Win 10

### Operating system

RaspberryPi OS 64bit

### Web server

Nginx

### PHP engine version

PHP 8.2

### Database

MariaDB

juliushaertl:
Thanks for reporting. Closing as a duplicate of https://github.com/nextcloud/tables/issues/451

Issue #725, opened 2023-11-27
Display of the number of records


FriedhelmLG:
### Steps to reproduce

1. There is a table with x records and 2 views
2. Add a data record in the table or in a view.
3. The correct number of data records is now: x+1.
4. In the sidebar, we can now switch back and forth between displaying the table or the views.

### Expected behavior

If a new data record has been successfully saved, the display of the number of existing data records should then be updated.

### Actual behavior

We can see that the number of rows in the sidebar, as well as in the table view, is not updated. If I manually reload the page in the browser, the correct number of rows is displayed.

### Tables app version

0.6.4

### Browser

Firefox 120

### Client operating system

Win10

### Operating system

Linux

### Web server

Apache

### PHP engine version

PHP 8.1

### Database

MariaDB

juliushaertl:
Thanks for reporting, this is already tracked in https://github.com/nextcloud/tables/issues/451

PR #2781, opened 2026-06-30
fix(rows): keep row counters fresh and reduce count queries

### LINKED ISSUE #451: Update rows counter
- Update if table row count is changed
- Update if views were adjusted (filtering changed, new/edited/deleted rows, ...)
Nearly everything could trigger the rows count. 🙄

hweihwang:
## Summary

Fixes stale row counters in Tables after row changes.

- Update the active table/view counter immediately after row load, create, update, or delete.
- Refresh affected sibling view counters after row changes, debounced per table.
- Count filtered view rows directly in SQL instead of materializing all matching row IDs in PHP.
- Batch table row counts when listing tables to avoid one count query per table.

Closes #451
Related: #725, #731

## Performance impact

This should be a net improvement for counter-heavy paths:

- Table list row counts go from one `COUNT(*)` query per table to one grouped `COUNT(*) ... GROUP BY table_id` query for all readable tables, chunked at 1000 IDs.
- Filtered view row counts now return a scalar SQL `COUNT(*)`; previously they loaded all matching row IDs into PHP and counted the array.
- Row create/update/delete adds at most one debounced view-list refresh per table per 500ms burst, instead of requiring a full page reload for fresh counters.

## Tests

- Added unit coverage for `countRowsForView()` matching filtered `findAll()` results.
- Added unit coverage for batched table row counts.
- Added Playwright coverage for navigation table/view counters after adding a row.

### DIFF
--- a/src/store/data.js
+++ b/src/store/data.js
@@ -11,10 +11,22 @@ import { MetaColumns } from '../shared/components/ncTable/mixins/metaColumns.js'
 import { showError } from '@nextcloud/dialogs'
 import { set } from 'vue'
 import { emit } from '@nextcloud/event-bus'
+import { getCurrentUser } from '@nextcloud/auth'
+import { useTablesStore } from './store.js'
 
 function genStateKey(isView, elementId) {
 	elementId = elementId.toString()
 	return isView ? 'view-' + elementId : elementId
+}
+
+const VIEW_COUNT_REFRESH_DELAY = 500
+const viewCountRefreshTimers = new Map()
+
+function canManageTable(table) {
+	if (!table.isShared) {
+		return true
+	}
+	return !!table.onSharePermissions?.manage || table.ownership === getCurrentUser()?.uid
 }
 
 export const useDataStore = defineStore('data', {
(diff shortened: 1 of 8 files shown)


blizzz:
Thanks for the fix! Which performance impact can we expect from this change?

hweihwang:
> Thanks for the fix! Which performance impact can we expect from this change?

@blizzz Ah thanks. I expect this to be a net positive:

table counters: N per-table count queries become one grouped count query for readable tables.
view counters: filtered views now use SQL COUNT(*) instead of loading all matching row IDs into PHP.
row changes add only one debounced view-counter refresh per table burst, not a full reload.
So the extra refresh is small, and the backend count paths are cheaper than before.

blizzz (approved):

src/store/data.js
```diff
@@ -233,6 +245,30 @@ export const useDataStore = defineStore('data', {
...
+		},
+
+		refreshTableViewCounts(tableId) {
+			if (!tableId) {
+				return
+			}
+			const table = useTablesStore().getTable(tableId)
+			if (!table || !canManageTable(table)) {
+				return
+			}
```
enjeck: with this, a collaborator on a shared table who can create, update, or delete rows but does not have manage permission never gets sibling view counters refreshed after their own edit ??

PR #2786, opened 2026-07-03
feat(federation): Federated table sharing

benjaminfrueh:
## Federated table sharing

Allows sharing tables with users on other Nextcloud instances.

### Todo
- [x] Schema, outgoing share and frontend support
- [x] Incoming share: create local table on receiving instance
- [x] Read support: fetch columns and rows from owner instance
- [x] Write support: create, update and delete rows on remote table
- [x] Permission propagation
- [x] Delete local table on receiving instance when table or share is deleted on owner instance
- [x] Update table metadata (title, emoji) when updated on owner instance
- [x] Update view metadata (title, emoji) when updated on owner instance
- [x] View federation
- [x] Admin config to enable/disable tables federation
- [ ] Unit and E2E tests

### 🏁 Checklist
 
- [x] 📅 Milestone is set

## 🤖 AI (if applicable)


lib/Controller/FederationColumnOCSController.php
```diff
@@ -0,0 +1,49 @@
...
+		private readonly ColumnService $columnService,
+	) {
+		parent::__construct($request, $logger, $n, '');
+	}
+
+	#[PublicPage]
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	#[AssertFederationIsAccessible]
+	public function index(int $tableId): DataResponse {
```
grnd-alt: why are you introducing extra federation endpoints here? Why not have this on the normal API path and adapt the permissionService to verify by federation token?

grnd-alt (changes requested):
Did not entirely check all code paths yet. But the general structure looks good 🔥 

What I noticed though: when creating a federated share it instantly shows as a new table on the receiving user's view. In Deck I think we're sending a notification that the user has to confirm before it shows in the app, so user's don't get spammed.

benjaminfrueh:
> Did not entirely check all code paths yet. But the general structure looks good 🔥
> 
> What I noticed though: when creating a federated share it instantly shows as a new table on the receiving user's view. In Deck I think we're sending a notification that the user has to confirm before it shows in the app, so user's don't get spammed.

Thanks for the feedback! After checking Deck's `shareReceived()`, it also seems to insert the board immediately and just sends an informational notification without accept/decline flow.
An "Pending shares" acceptance flow, similar to how Files handles federated shares would maybe be a good addition to both deck and tables federated sharing - maybe we can consider it as a follow-up feature?

lib/AppInfo/Application.php
```diff
@@ -8,18 +8,22 @@
 namespace OCA\Tables\AppInfo;
 
 use Exception;
+use OC\OCM\OCMSignatoryManager;
```
blizzz, line 11: Would be cool (and more sustainable) when there will be a public interface for this, but definitely out of scope for now, might be accompanied by an issue against server if there is not one yet.

(On second look it might be more complicated than that)

lib/Controller/Api1Controller.php
```diff
@@ -770,6 +772,11 @@ public function updateShareDisplayMode(int $shareId, int $displayMode, string $t
 	#[CORS]
 	#[OpenAPI(scope: OpenAPI::SCOPE_DEFAULT)]
 	public function indexTableColumns(int $tableId, ?int $viewId): DataResponse {
+		if ($this->federationService->isNodeFederated($tableId, 'table')) {
+			$table = $this->tableService->find($tableId, true);
+			return new DataResponse($this->federationService->getColumns($table));
+		}
```
blizzz: Smells a bit as if it wants to be in the try-block, too.

lib/Controller/Api1Controller.php
```diff
@@ -813,6 +820,11 @@ public function indexTableColumns(int $tableId, ?int $viewId): DataResponse {
 	#[RequirePermission(permission: Application::PERMISSION_READ, type: Application::NODE_TYPE_VIEW, idParam: 'viewId')]
 	#[OpenAPI(scope: OpenAPI::SCOPE_DEFAULT)]
 	public function indexViewColumns(int $viewId): DataResponse {
+		if ($this->federationService->isNodeFederated($viewId, 'view')) {
+			$view = $this->viewService->find($viewId, true);
+			return new DataResponse($this->federationService->getColumns($view));
+		}
```
blizzz: same here

lib/Controller/RowController.php
```diff
@@ -24,13 +27,20 @@ public function __construct(
...
 		parent::__construct(Application::APP_ID, $request);
 	}
 
 	#[NoAdminRequired]
 	#[RequirePermission(permission: Application::PERMISSION_READ, type: Application::NODE_TYPE_TABLE, idParam: 'tableId')]
 	public function index(int $tableId): DataResponse {
+		if ($this->federationService->isNodeFederated($tableId, 'table')) {
+			$table = $this->tableService->find($tableId, true);
+			return new DataResponse($this->federationService->getRows($table));
+		}
```
blizzz: and this might want to be within the `handleError()` callback?

lib/Controller/RowController.php
```diff
@@ -98,6 +112,9 @@ public function destroyByView(int $id, int $viewId): DataResponse {
 
 	#[NoAdminRequired]
 	public function presentInView(int $id, int $viewId): DataResponse {
+		if ($this->federationService->isNodeFederated($viewId, 'view')) {
+			return new DataResponse(['present' => true]);
+		}
```
blizzz: same here

lib/Controller/RowOCSController.php
```diff
@@ -88,4 +104,98 @@ public function createRow(string $nodeCollection, int $nodeId, mixed $data): Dat
...
+	 * 400: Invalid request parameters
+	 * 403: No permissions
+	 * 404: Not found
+	 * 500: Internal error
+	 */
+	#[NoAdminRequired]
+	#[RequirePermission(permission: Application::PERMISSION_UPDATE, typeParam: 'nodeCollection')]
+	#[ApiRoute(verb: 'PUT', url: '/api/2/{nodeCollection}/{nodeId}/rows/{rowId}', requirements: ['nodeCollection' => '(tables|views)', 'nodeId' => '(\d+)'])]
+	public function updateRow(string $nodeCollection, int $nodeId, int $rowId, mixed $data): DataResponse {
+		if (is_string($data)) {
```
blizzz, line 127: `$data` could be of type `string|array` then?

lib/Controller/RowOCSController.php
```diff
@@ -88,4 +104,98 @@ public function createRow(string $nodeCollection, int $nodeId, mixed $data): Dat
...
+				$table = $this->tableService->find($nodeId, true);
+				return new DataResponse($this->federationService->updateRow($table, $rowId, $data));
+			}
+		} elseif ($iNodeType === Application::NODE_TYPE_VIEW) {
+			$viewId = $nodeId;
+			if ($this->federationService->isNodeFederated($viewId, 'view')) {
+				$view = $this->viewService->find($nodeId, false, $this->userId);
+				return new DataResponse($this->federationService->updateRow($view, $rowId, $data));
+			}
+		}
```
blizzz, line 139: Here I am also think this should go into the try block

blizzz: perhaps also move the logic into a different method.

lib/Controller/RowOCSController.php
```diff
@@ -88,4 +104,98 @@ public function createRow(string $nodeCollection, int $nodeId, mixed $data): Dat
...
+				$table = $this->tableService->find($nodeId, true);
+				return new DataResponse($this->federationService->deleteRow($table, $rowId));
+			}
+		} elseif ($iNodeType === Application::NODE_TYPE_VIEW) {
+			$viewId = $nodeId;
+			if ($this->federationService->isNodeFederated($viewId, 'view')) {
+				$view = $this->viewService->find($nodeId, false, $this->userId);
+				return new DataResponse($this->federationService->deleteRow($view, $rowId));
+			}
+		}
```
blizzz, line 182: as above

lib/Federation/FederationProvider.php
```diff
@@ -0,0 +1,211 @@
...
+	 */
+	public function shareReceived(ICloudFederationShare $share): string {
+		try {
+			$this->federationService->ensureIncomingFederationEnabled();
+		} catch (FederationDisabledError $e) {
+			throw new ProviderCouldNotAddShareException($e->getMessage());
+		}
+
+		$localUser = $share->getShareWith();
+		if (str_contains($localUser, '@')) {
```
blizzz, line 73: this could also be an email address, @ is not a forbidden character in the user id 🤔

lib/Federation/FederationProvider.php
```diff
@@ -0,0 +1,211 @@
...
+			$localUser = $this->cloudIdManager->resolveCloudId($localUser)->getUser();
+		}
+
+		$metaData = json_decode($share->getDescription(), true) ?? [];
+		if (!isset($metaData['nodeType'])) {
+			throw new ProviderCouldNotAddShareException('Missing node type in share description');
+		}
+
+		$nodeType = $metaData['nodeType'];
```
blizzz, line 82: can we be sure it is a valid node type?

lib/Federation/FederationProvider.php
```diff
@@ -0,0 +1,211 @@
...
+			self::NOTIFICATION_DELETE_NODE => $this->handleNodeDelete($providerId, $notification),
+			self::NOTIFICATION_UPDATE_NODE => $this->handleNodeUpdate($providerId, $notification),
+			default => throw new BadRequestException(['nodeType']),
+		};
+
+		return [];
+	}
+
+	public function getSupportedShareTypes(): array {
+		return ['user'];
```
blizzz, line 111: worth to document this is limited to users at first.

lib/Federation/FederationProvider.php
```diff
@@ -0,0 +1,211 @@
...
+		$node->setOwnership($share->getOwner());
+		$node->setShareToken($share->getShareSecret());
+		$node->setCreatedBy($share->getSharedBy());
+		$node->setCreatedAt($now);
+		$node->setLastEditBy($share->getSharedBy());
+		$node->setLastEditAt($now);
+		$node->setEmoji($metaData['emoji'] ?? null);
+
+		if ($node instanceof View) {
+			$node->setDescription('');
```
blizzz, line 151: now we also need to set the uuid (if provided, older versions may not 🤔 unless both features are released at the same time)

lib/Federation/FederationProxy.php
```diff
@@ -0,0 +1,174 @@
+<?php
```
blizzz, line 1: declare strict is missing (check other new files as well)

lib/Federation/FederationProxy.php
```diff
@@ -0,0 +1,174 @@
...
+			'nextcloud' => [
+				'allow_local_address' => $this->config->getSystemValueBool('allow_local_remote_servers'),
+			],
+			'headers' => [
+				'Accept' => 'application/json',
+				'OCS-APIRequest' => 'true',
+				'Accept-Language' => $this->l10nFactory->getUserLanguage($this->userSession->getUser()),
+				'tables-federation-accesstoken' => $accessToken,
+			],
+			'timeout' => 5,
```
blizzz, line 61: not too low? but Deck does the same.

lib/Federation/FederationProxy.php
```diff
@@ -0,0 +1,174 @@
...
+			$content = $body->getContents();
+			$body->rewind();
+
+			if (!is_array(json_decode($content, true))) {
+				throw new \Exception('Error parsing JSON response', $status);
+			}
+
+			$this->logger->debug('Client error from remote', ['exception' => $e]);
+
+			/** @psalm-suppress InvalidReturnStatement */
```
blizzz, line 102: ?

lib/Migration/Version002003Date20260630000000.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace OCA\Tables\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version002003Date20260630000000 extends SimpleMigrationStep {
```
blizzz: afraid this is sorted not at latest and might not run, but i might be wrong

lib/Service/FederationService.php
```diff
@@ -0,0 +1,258 @@
...
+				FederationProvider::NOTIFICATION_DELETE_NODE,
+				(string)$node->getId(),
+				$share,
+				['nodeType' => $nodeType],
+			);
+		}
+	}
+
+	public function notifyPermissionUpdate(Share $share): void {
+		$this->proxy->sendNotification(
```
blizzz, line 186: what if any of those requests do not get through (maybe the remote nextcloud is being updated), will they be retried?

psalm.xml
```diff
@@ -45,6 +45,10 @@
 				<referencedClass name="Doctrine\DBAL\Schema\Table" />
 				<referencedClass name="Doctrine\DBAL\Types\Types" />
 				<referencedClass name="OC\Core\Command\Base" />
+				<referencedClass name="OC\Http\Client\Response" />
+				<referencedClass name="OC\OCM\OCMSignatoryManager" />
```
blizzz, line 50: stub would be preferred

psalm.xml
```diff
@@ -45,6 +45,10 @@
 				<referencedClass name="Doctrine\DBAL\Schema\Table" />
 				<referencedClass name="Doctrine\DBAL\Types\Types" />
 				<referencedClass name="OC\Core\Command\Base" />
+				<referencedClass name="OC\Http\Client\Response" />
```
blizzz, line 49: this should have an interface?

blizzz (commented):
some things that would be good to have, but no definite blocker by reading the code

lib/Federation/FederationProxy.php
```diff
@@ -0,0 +1,174 @@
...
+			return new Response($e->getResponse(), false);
```
blizzz, line 103: @grnd-alt same in Deck, I wonder what the background is?

lib/Federation/FederationProxy.php
```diff
@@ -0,0 +1,174 @@
...
+				$requestOptions
+			);
+		} catch (ClientException $e) {
+			$status = $e->getResponse()->getStatusCode();
+			$body = $e->getResponse()->getBody();
+			$content = $body->getContents();
+			$body->rewind();
+
+			if (!is_array(json_decode($content, true))) {
+				throw new \Exception('Error parsing JSON response', $status);
```
blizzz: Here and and other places, throwing a generic `\Exception` is broad, and so would be catching them.

lib/Controller/Api1Controller.php
```diff
@@ -770,6 +772,11 @@ public function updateShareDisplayMode(int $shareId, int $displayMode, string $t
 	#[CORS]
 	#[OpenAPI(scope: OpenAPI::SCOPE_DEFAULT)]
 	public function indexTableColumns(int $tableId, ?int $viewId): DataResponse {
+		if ($this->federationService->isNodeFederated($tableId, 'table')) {
```
blizzz: and this actually throws when federation is disabled

blizzz (changes requested):

lib/Controller/Api1Controller.php
blizzz: same in  the method below
and in RowController
and in the places in the RowOCSController

benjaminfrueh: `isNodeFederated()` now simply returns false instead of throwing, when federation is disabled. Thank you @samin-z for this change 💙

lib/Federation/FederationProvider.php
benjaminfrueh: The `FederationService::sendShare()` validates on the sender side, but `shareReceived()` should validate again on the receiver side for better defense against malformed requests.

benjaminfrueh: commit 85c35bfc68fb20b62a08473bc7aefc5c7f74c880

lib/Federation/FederationProxy.php
benjaminfrueh: commit 0f29ad5d3d2aa4e0f94bfc2415ec914617a69a00

lib/Federation/FederationProxy.php
benjaminfrueh: I kept it at 5 seconds to match decks timeout. It could be made configurable if needed.

lib/Service/FederationService.php
benjaminfrueh: Currently no retry mechanism, if the notification fails it is logged as a warning.

lib/Federation/FederationProxy.php
benjaminfrueh: commit 3bad71ca03f08af56727ee03aac468da34c4434d

lib/Federation/FederationProvider.php
benjaminfrueh: The check for `@` here is just a tiny optimization to skip the `resolveCloudId()` call in case of malformed data, this could be removed entirely aswell. In `shareReceived` the data comes from an OCM share, so it should always be a cloud ID.

Additionally `resolveCloudId()` also throws and reject the share if the user is not a valid cloud ID.

lib/Migration/Version002003Date20260630000000.php
benjaminfrueh: I updated the migration as `Version2203Date20260821000000` now.
Commit 63f67ab5b25167a2a6f56755c51e475268124923

lib/Federation/FederationProvider.php
benjaminfrueh: commit c0f2f78ccbb0c8da8783ae7e4eb90a82fc356c2b

blizzz:
*[Follow up/General tech debt]* One thing about Migrations, because i just stumbled across them recently, are the attributes. Can be added as followup, and I do not think we have them in place elsewhere. https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/migrations.html#migrations-and-metadata

blizzz (approved):

PR #2808, opened 2026-07-07
Fix scroll on embedded widget and adjust height to screen size

### LINKED ISSUE #2596: Embedded tables some column overflow on header

taiebot:
### 🖼️ Screenshots

Fix https://github.com/nextcloud/tables/issues/2596 another solution to #2744 

This has been tested on multiple scenarios. 
With ClaudeAI i have done multiple tests cases and this is the solution i came up with.

The table widget embedded in Nextcloud Text/Collectives had several visual bugs when scrolling and missed the three dots to manage the widget. After this PR the widget renders properly.

Interactive cells (star ratings, URLs, richtext editors) scrolled over the sticky header (title + search bar) instead of passing underneath it
The widget had a fixed height of 50vh with no adaptation to screen size
On iOS/iPad, the widget scroll captured the entire page scroll

🏚️ On main| 🏡 With PR 
---|---
image| image

### 🏁 Checklist
 
- [X] ⛑️ Tests (unit and/or integration) are included or not needed
- [X] 🔙 Backport requests are created or not needed: `/backport to stableX.X`
- [ ] 📅 Milestone is set
- [X] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

## 🤖 AI (if applicable)

- [X] The content of this PR was partly or fully generated using AI

What is still broken on embedded table and not part of this pr
- [ ] column ordering do not respect table
- [ ] sort is not even plugged in


enjeck (commented):
Thanks for the fix. Could you please confirm whether you manually tested this with:
- an embedded table where row actions are visible, so the right-side action column also scrolls away instead of staying sticky
- a wide table with horizontal scrolling

I’m especially wondering about the action column, since it still receives sticky styling from the shared table component.

src/views/ContentReferenceWidget.vue
```diff
@@ -249,14 +248,18 @@ export default {
 
 	.tables-content-widget {
 		min-height: max(50vh, 200px);
-		height: 50vh;
+		height: 60vh;
```
Aveyron-RetD: in here should be auto

Aveyron-RetD:
I have extensively tested this, and yes both of your points are fixed  but I have realized that I need to add auto to height for the table to resize. In here I have set the table height which is stupid. 🤦 

The only stuff which is weird regarding this PR, is this
image
Which scroll underneath the header when the embedded is not wide enough, but that is in a much better shape than today.

But I am still not happy with the embedded current state.

I believe to store data in the embedded is not good because a refresh does not bring new data from the tables app itself.
In its current form, the embedded is not dynamic enough.

I wonder if I should close this and bring a Big PR with a rework of embedded, but that will be harder to review.

I say push this with auto and I will bring more ❤️ to embedded

PR #2819, opened 2026-07-17
Add notification for async import finished event

### LINKED ISSUE #2764: Add asynchronous import notification

Add notification for asynchronous import to notify user when importing is done.

_No response_

_No response_

### Additional context

_No response_

luka-nextcloud:
* Resolves: #2764

### 🏁 Checklist
 
- [ ] 📅 Milestone is set

### DIFF
--- a/lib/Notification/NotificationHelper.php
+++ b/lib/Notification/NotificationHelper.php
@@ -43,14 +43,19 @@ class NotificationHelper {
 	}
 
 	/**
-	 * @param Row2|Column $object
+	 * @param Table|Row2|Column $object
 	 * @param string $subject
 	 * @param array<string, mixed> $additionalParams
 	 * @param string|null $author
 	 */
-	public function sendNotification(string $objectType, Row2|Column $object, string $subject, $additionalParams = [], ?string $author = null): void {
+	public function sendNotification(string $objectType, Table|Row2|Column $object, string $subject, array $additionalParams = [], ?string $author = null): void {
 		try {
 			switch ($objectType) {
+				case ActivityManager::TABLES_OBJECT_TABLE:
+					if ($object instanceof Table) {
+						$this->sendTableNotification($object, $subject, $author);
+					}
+					break;
 				case ActivityManager::TABLES_OBJECT_ROW:
 					if ($object instanceof Row2) {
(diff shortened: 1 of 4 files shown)


lib/Notification/NotificationHelper.php
```diff
@@ -72,6 +77,26 @@ public function sendNotification(string $objectType, Row2|Column $object, string
...
+			element: $object,
+			subject: $subject,
+			subjectParams: $subjectParams,
+			objectType: ActivityManager::TABLES_OBJECT_TABLE,
+			objectId: (string)$object->getId(),
+			authorId: null,
+			configKey: null,
+		);
+	}
+
```
enjeck, line 117: i havent tested yet, but it seems this sends the async import completion notification through the table-recipient path. So for imports started from a view, the importing user may only have access to the view and therefore may not be included in `findRecipientsByElement($table)` ? 

Also, doesn't this notify unrelated table recipients about someone else’s import

lib/Notification/NotificationHelper.php
```diff
@@ -72,6 +82,46 @@ public function sendNotification(string $objectType, Row2|Column $object, string
...
+	private function sendTableNotification(Table $object, string $subject, ?string $author): void {
+		$subjectParams = [
+			'author' => $author,
+			'objectType' => ActivityManager::TABLES_OBJECT_TABLE,
+			'table' => [
+				'id' => $object->getId(),
+				'title' => $object->getTitle(),
+			]
+		];
+		$this->sendNotifiesByElement(
```
samin-z, line 107: i believe this notify everyone who this table is shared with that the import is done, doesnt it make sense only the person who did the import be notified?

samin-z:
ok so what i mentioned is already covered, i only found two things which one is not regarding this pr but thought maybe worth mentioning, and the other one could useful:

1. when import is done (specially if the import is very big) in the activity table per each row that is inserted there is a message which for this case can get a bit out of hand, but we can tackle that maybe in another task.

2. worth mentioning, if the import fails for whatever reason there is no notifications, i created a table and imported a file, while the background job was starting i removed the file, but no notifications was returned for the import that it failed. i also tried it with removing the table.

samin-z (approved):

PR #2829, opened 2026-07-18
Fix: Fix pipeline

Koc:
This PR attempts to fix pipeline:
- fix [Psalm](https://github.com/nextcloud/tables/actions/runs/29660633815/job/88122580829?pr=2829) for Nextcloud `main` branch
- fix [undefined steps](https://github.com/nextcloud/tables/actions/runs/29660633820/job/88122594011?pr=2829#step:8:2451)
- fail pipeline if Behat tests scenarios contain undefined steps to prevent merging of incomplete tests
- fix code to pass test scenario `Setting permissions on a context share is forbidden`

image

image

### 🏁 Checklist 
- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Share/SharePageShareDataDecorator.php
+++ b/lib/Share/SharePageShareDataDecorator.php
@@ -230,6 +230,13 @@ class SharePageShareDataDecorator implements IShare {
 		return $this->share->getPassword();
 	}
 
+	/**
+	 * Forward-compatibility with Nextcloud v35
+	 */
+	public function isPasswordProtected(): bool {
+		return !empty($this->share->getPassword());
+	}
+
 	public function setToken($token): never {
 		throw new LogicException('Not implemented: read only object');
 	}
--- a/tests/integration/features/bootstrap/FeatureContext.php
+++ b/tests/integration/features/bootstrap/FeatureContext.php
@@ -2789,6 +2789,108 @@ class FeatureContext implements Context {
 	}
 
 	/**
+	 * @Given user :sharer shares the Context :contextAlias to :shareeType :sharee with permissions :permissions
+	 */
+	public function userSharesTheContextToWithPermissions(string $sharer, string $contextAlias, string $shareeType, string $sharee, string $permissions): void {
+		$this->setCurrentUser($sharer);
+		$context = $this->collectionManager->getByAlias('context', $contextAlias);
+
+		$permissionMap = [
+			'read' => 'permissionRead',
+			'create' => 'permissionCreate',
+			'update' => 'permissionUpdate',
+			'delete' => 'permissionDelete',
+			'manage' => 'permissionManage',
+		];
+
+		$permissionList = explode(',', $permissions);
+		$permissionData = [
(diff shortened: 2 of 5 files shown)


lib/Share/SharePageShareDataDecorator.php
```diff
@@ -230,6 +230,13 @@ public function getPassword(): string {
 		return $this->share->getPassword();
 	}
 
+	/**
+	 * Forward-compatibility with Nextcloud v35
+	 */
+	public function isPasswordProtected(): bool {
+		return $this->share->getPassword() !== '';
+	}
```
enjeck, line 238: this returns true when no password is set? why?

Maybe something like this instead:

```suggestion
	public function isPasswordProtected(): bool {
		return ($this->share->getPassword() ?? '') !== '';
	}
```

tests/integration/run.sh
```diff
@@ -78,12 +78,12 @@ echo ''
...
 echo '#'
-vendor/bin/behat --colors -f junit -f pretty $SCENARIO
+vendor/bin/behat --strict --colors -f junit -f pretty $SCENARIO
 RESULT=$?
 
 echo ''
 echo '#'
-echo '# Stopping PHP webserver and disabling spreedcheats'
+echo '# Stopping PHP webserver and disabling spreadsheets'
 echo '#'
```
enjeck: "spreedcheats" is leftover boilerplate copied from Nextcloud Talk's integration tests, where a helper app is literally named spreedcheats. Nothing named spreedcheats or spreadsheets is actually being disabled in this script, so renaming it to "spreadsheets" is wrong

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -2788,6 +2788,108 @@ public function userSharesTheContextTo(string $sharer, string $contextAlias, str
...
+			};
+
+			Assert::assertEquals($expectedValue, (int)($contextData['onSharePermissions'][$permissionField] ?? 0), "Permission $permissionName mismatch");
+		}
+	}
+
+	/**
+	 * @When user :user tries to set permission :permission to :value for context :contextAlias and user :targetUser
+	 */
+	public function userTriesToSetPermissionToForContextAndUser(string $user, string $permission, string $value, string $contextAlias, string $targetUser): void {
```
enjeck: why pass in `$targetUser` if we don't use it?

enjeck (commented):
some comments:

lib/Share/SharePageShareDataDecorator.php
Koc: ah, I wasn't aware that we can have null in this field (phpdocs are completely broken). Replaced with `empty` check

tests/integration/features/bootstrap/FeatureContext.php
Koc: ah, indeed. Removed

blizzz (approved):

Issue #2767, opened 2026-06-26
Vue3 migration


blizzz:
Vue2 is long outdated, we want to follow NC server (or might have to for compatibility)  and upgrade to Vue3.

blizzz:
completed in https://github.com/nextcloud/tables/pull/2836

PR #2836, opened 2026-07-22
Migrate from Vue2 to Vue3

enjeck:
Migrate from vue 2 to vue 3: #2767 

### 🏁 Checklist
 
- [x] 📅 Milestone is set

## 🤖 AI (if applicable)


package.json
```diff
@@ -44,21 +44,20 @@
...
     "@nextcloud/router": "^3.1.0",
-    "@nextcloud/vue": "^8.39.0",
-    "@tiptap/extension-character-count": "^3.27.4",
-    "@tiptap/extension-task-item": "^3.27.4",
-    "@tiptap/extension-task-list": "^3.27.4",
-    "@tiptap/starter-kit": "^3.27.4",
-    "@tiptap/vue-2": "^3.27.4",
-    "@vueuse/core": "^11.3.0",
+    "@nextcloud/vue": "^9.8.2",
+    "@tiptap/extension-character-count": "^3.27.1",
```
blizzz: is it a downgrade? (Did not check what in the lock file, but looks odd). Same with packages below.

blizzz (commented):
🥳  did not test yet, green CI is promising

samin-z:
gaps found:
1. route name assertion (like table -> tableContent)
2. sort emits (.sync -> update:sort-entry change not validated)
3.share permission toggle (no test for user share permission toggle, only public)

there is also a leftover :checked on NcActionButton in 'TableHeaderColumnOptions.vue' nothing big.

enjeck:
> gaps found:
> 
> 1. route name assertion (like table -> tableContent)
> 2. sort emits (.sync -> update:sort-entry change not validated)
>    3.share permission toggle (no test for user share permission toggle, only public)
> 
> there is also a leftover :checked on NcActionButton in 'TableHeaderColumnOptions.vue' nothing big.

For 3, we should manually test that works as expected since no test for that
For the other points, not sure what their impact is. What's broken if we don't change it? @samin-z

samin-z:
@enjeck 
number 1: 
if route names arent tested and still reference old names like 'table' instead of 'tableContent', navigation by name silently fails or 404. users won't be able to reach tables via programmatic navigation.
number 2:
if sort emit is broken, sort changes will not be saved. if user sort the table, it'll revert on reload.

for 3 then i'll try to tests it manually.

samin-z:
@enjeck i checked all the points, everything works as they should so no need to look into them, everything except point 2 which is regarding sorting, after you sort and refresh the page, the previous selected sorting is gone.
i also checked it on main and it is working as it remembers previous state.
otherwise i also plaied around a little bit and everything was fine.

enjeck:
> i also checked it on main and it is working as it remembers previous state.

i checked on main and when I sort a view column, sorting clears when i refresh the page. Just like it does now

samin-z:
> i checked on main and when I sort a view column, sorting clears when i refresh the page. Just like it does now
you're correct, sorry it was on my side with the data i checked threw me off.

samin-z (approved):

enjeck:
Holding off on merging until a release is done first. So vue 3 has time in main before the next release

PR #2838, opened 2026-07-23
fix: align share response with documented response definition

blizzz:
Any related output of the Api1Controller goes through `formatForOutput` now. It masks the password hash (sets it to 'yes' if present), which ain't an issue for any use case. It's fine to keep that on the server.


enjeck (approved):

blizzz:
failing cypress+playwright tests against master are an unrelated server regression

PR #2857, opened 2026-07-29
perf: Enhance objects using constant sql queries count to get rid of N+1 queries

Koc:
In our organization we operate a large amount of tables/views, so there are performance bottleneck here:

```
PERFORMANCE: Controller OCA\\Tables\\Controller\\Api1Controller::index executed 1883 queries in 20.1 seconds
PERFORMANCE: Controller OCA\\Tables\\Controller\\TableController::index executed 1887 queries in 16.6 seconds
```

- Replaced `TableService::enhanceTable()` in a loop with a single call of the `enhanceTables()`
- Replaced `ViewService::enhanceView()` in a loop with a single call of the`enhanceViews()`
- Added batch helpers in mappers/services and `UserHelper` for row/column counts, share counts and owner display names.

### 🖼️ Screenshots

Here profiler results for `GET /apps/tables/table` endpoint for 18 tables and 9 views:

🏚️ Before | 🏡 After
---|---
121 queries | 29 queries
image | image

### 🏁 Checklist
 
- [x] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Controller/Api1Controller.php
+++ b/lib/Controller/Api1Controller.php
@@ -326,7 +326,8 @@ class Api1Controller extends ApiController {
 	#[OpenAPI(scope: OpenAPI::SCOPE_DEFAULT)]
 	public function indexViews(int $tableId): DataResponse {
 		try {
-			return new DataResponse($this->viewService->formatViews($this->viewService->findAll($this->tableService->find($tableId))));
+			$table = $this->tableService->find($tableId);
+			return new DataResponse($this->viewService->formatViews($this->viewService->findAll($table, tableRowsCount: $table->getRowsCount())));
 		} catch (PermissionError $e) {
 			$this->logger->warning('A permission error occurred: ' . $e->getMessage(), ['exception' => $e]);
 			$message = ['message' => $e->getMessage()];
--- a/lib/Db/ColumnMapper.php
+++ b/lib/Db/ColumnMapper.php
@@ -186,6 +186,41 @@ class ColumnMapper extends QBMapper {
 	}
 
 	/**
+	 * @param int[] $tableIds
+	 * @return array<int, int>
+	 */
+	public function countColumnsForTables(array $tableIds): array {
+		if (empty($tableIds)) {
+			return [];
+		}
+
+		$counts = [];
+		foreach (array_chunk($tableIds, 1000 - 1) as $tableIdsChunk) {
+			$qb = $this->db->getQueryBuilder();
+			$qb->select('table_id', $qb->func()->count('*', 'counter'))
+				->from($this->table)
+				->where($qb->expr()->in('table_id', $qb->createNamedParameter($tableIdsChunk, IQueryBuilder::PARAM_INT_ARRAY)))
+				->groupBy('table_id');
+
+			try {
+				$result = $qb->executeQuery();
+				while ($row = $result->fetch()) {
+					$counts[(int)$row['table_id']] = (int)$row['counter'];
+				}
+				$result->closeCursor();
+			} catch (Exception $e) {
+				$this->logger->warning('Exception occurred: ' . $e->getMessage() . ' Returning 0 for all given tables.');
+			}
+		}
+
+		foreach ($tableIds as $tableId) {
+			$counts[$tableId] ??= 0;
+		}
+
+		return $counts;
+	}
+
+	/**
 	 * Preloads columns data in bulk to optimize caching and reduce database queries.
 	 * This method efficiently loads column data for a given set of columns, filters, and sorts
 	 * by fetching all required data in a single database operation.
--- a/lib/Db/ColumnMapper.php
+++ b/lib/Db/ColumnMapper.php
@@ -165,27 +165,6 @@ class ColumnMapper extends QBMapper {
 	}
 
 	/**
-	 * @param int $tableId
-	 * @return int
-	 */
-	public function countColumns(int $tableId): int {
-		$qb = $this->db->getQueryBuilder();
-		$qb->select($qb->func()->count('*', 'counter'));
-		$qb->from($this->table);
-		$qb->where(
-			$qb->expr()->eq('table_id', $qb->createNamedParameter($tableId))
-		);
-
-		try {
-			$result = $this->findOneQuery($qb);
-			return (int)$result['counter'];
-		} catch (DoesNotExistException|MultipleObjectsReturnedException|Exception $e) {
-			$this->logger->warning('Exception occurred: ' . $e->getMessage() . ' Returning 0.');
-			return 0;
-		}
-	}
-
-	/**
 	 * @param int[] $tableIds
 	 * @return array<int, int>
 	 */
--- a/lib/Db/RowSleeveMapper.php
+++ b/lib/Db/RowSleeveMapper.php
@@ -125,6 +125,41 @@ class RowSleeveMapper extends QBMapper {
 
 	/**
 	 * @param int[] $tableIds
+	 * @return array<int, int>
+	 */
+	public function countRowsForTables(array $tableIds): array {
+		if (empty($tableIds)) {
+			return [];
+		}
+
+		$counts = [];
+		foreach (array_chunk($tableIds, 1000 - 1) as $tableIdsChunk) {
+			$qb = $this->db->getQueryBuilder();
+			$qb->select('table_id', $qb->func()->count('*', 'counter'))
+				->from($this->table, 't1')
+				->where($qb->expr()->in('table_id', $qb->createNamedParameter($tableIdsChunk, IQueryBuilder::PARAM_INT_ARRAY)))
+				->groupBy('table_id');
+
+			try {
+				$result = $qb->executeQuery();
+				while ($row = $result->fetch()) {
+					$counts[(int)$row['table_id']] = (int)$row['counter'];
+				}
+				$result->closeCursor();
+			} catch (Exception $e) {
+				$this->logger->warning('Exception occurred: ' . $e->getMessage() . ' Will return 0 for all given tables.');
+			}
+		}
+
+		foreach ($tableIds as $tableId) {
+			$counts[$tableId] ??= 0;
+		}
+
+		return $counts;
+	}
+
+	/**
+	 * @param int[] $tableIds
 	 *
 	 * @return RowSleeve[]
 	 */
--- a/lib/Db/RowSleeveMapper.php
+++ b/lib/Db/RowSleeveMapper.php
@@ -103,27 +103,6 @@ class RowSleeveMapper extends QBMapper {
 	}
 
 	/**
-	 * @param int $tableId
-	 * @return int
-	 */
-	public function countRows(int $tableId): int {
-		$qb = $this->db->getQueryBuilder();
-		$qb->select($qb->func()->count('*', 'counter'));
-		$qb->from($this->table, 't1');
-		$qb->where(
-			$qb->expr()->eq('table_id', $qb->createNamedParameter($tableId))
-		);
-
-		try {
-			$result = $this->findOneQuery($qb);
-			return (int)$result['counter'];
-		} catch (DoesNotExistException|MultipleObjectsReturnedException|Exception $e) {
-			$this->logger->warning('Exception occurred: ' . $e->getMessage() . ' Will return 0.');
-			return 0;
-		}
-	}
-
-	/**
 	 * @param int[] $tableIds
 	 * @return array<int, int>
 	 */
--- a/lib/Db/ShareMapper.php
+++ b/lib/Db/ShareMapper.php
@@ -7,6 +7,7 @@
 
 namespace OCA\Tables\Db;
 
+use OCA\Tables\Constants\ShareReceiverType;
 use OCA\Tables\Service\ValueObject\ShareToken;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\MultipleObjectsReturnedException;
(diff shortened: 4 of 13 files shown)


lib/Db/ShareMapper.php
```diff
@@ -136,6 +137,77 @@ public function findAllSharesForNode(string $nodeType, int $nodeId, string $send
...
 
+	/**
+	 * @param string $nodeType
+	 * @param int[] $nodeIds
+	 * @param string $sender
+	 * @param array<string> $excluded receiver types to exclude from results
+	 * @return Share[]
+	 * @throws Exception
+	 */
+	public function findAllSharesForNodes(string $nodeType, array $nodeIds, string $sender = '', array $excluded = []): array {
```
enjeck, line 148: Is `sender` applied to the query? otherwise not sure why defined

enjeck: This filters by `sender`, but the singular `findAllSharesForNode` above ignores its `$sender` arg entirely. So countSharesForTables/countSharesForViews now count only shares I sent, while the old `findAll` path counted all shares on the node. This flips `hasShares` to false for a table/view I own that a different manager shared. 

Is this intended? If yes, maybe make `findAllSharesForNode` honor `$sender` too so both paths agree.

lib/Service/ViewService.php
```diff
@@ -511,10 +593,14 @@ private function setIsSharedState(View $view, string $userId): void {
...
 				// (senseless if we have no user in context)
-				try {
-					$allShares = $this->shareService->findAll('view', $view->getId());
-					$view->setHasShares(count($allShares) !== 0);
-				} catch (InternalError $e) {
+				if ($sharesCount !== null) {
+					$view->setHasShares($sharesCount > 0);
+				} else {
+					try {
+						$allShares = $this->shareService->findAll('view', $view->getId(), $userId);
```
enjeck: The batch path (countSharesForViews) is sender filtered, but this calls `findAll('view', ...)` which is not. The same view can get a different `hasShares` depending on which path runs?

lib/Db/RowSleeveMapper.php
```diff
@@ -103,24 +103,38 @@ public function deleteAllForTable(int $tableId): int {
...
+	public function countRowsForTables(array $tableIds): array {
+		if (empty($tableIds)) {
+			return [];
 		}
+
+		$counts = [];
+		foreach (array_chunk($tableIds, 1000 - 1) as $tableIdsChunk) {
+			$qb = $this->db->getQueryBuilder();
+			$qb->select('table_id', $qb->func()->count('*', 'counter'))
+				->from($this->table, 't1')
```
enjeck: not sure we need the `t1` alias

lib/Controller/Api1Controller.php
```diff
@@ -326,7 +326,8 @@ public function deleteTable(int $tableId): DataResponse {
 	#[OpenAPI(scope: OpenAPI::SCOPE_DEFAULT)]
 	public function indexViews(int $tableId): DataResponse {
 		try {
-			return new DataResponse($this->viewService->formatViews($this->viewService->findAll($this->tableService->find($tableId))));
+			$table = $this->tableService->find($tableId);
+			return new DataResponse($this->viewService->formatViews($this->viewService->findAll($table, tableRowsCount: $table->getRowsCount())));
```
enjeck: thinking about performance, `tableService->find()` already enhances the table and loads its views, then `findAll(...)` loads and enhances them again. Is it possible to avoid this duplicate work or no?

lib/Db/RowSleeveMapper.php
Koc: that's removed

lib/Controller/Api1Controller.php
Koc: I did changes, so, now we're calling `$table->getViews()`. Tests are green, but I'm not fully confident that this endpoint is covered by tests. And it seems like we aren't using this endpoint on FE side

lib/Service/ViewService.php
Koc: Now  fallback uses `countSharesForViews(...)` instead of `findAll(...)`

lib/Db/ShareMapper.php
Koc: `findAllSharesForNode()` now honors `$sender`

lib/Db/ShareMapper.php
```diff
@@ -129,13 +130,88 @@ public function findAllSharesForNode(string $nodeType, int $nodeId, string $send
 			->andWhere($qb->expr()->eq('node_type', $qb->createNamedParameter($nodeType, IQueryBuilder::PARAM_STR)))
 			->andWhere($qb->expr()->eq('node_id', $qb->createNamedParameter($nodeId, IQueryBuilder::PARAM_INT)));
 
+		if ($sender !== '') {
+			$qb->andWhere($qb->expr()->eq('sender', $qb->createNamedParameter($sender, IQueryBuilder::PARAM_STR)));
+		}
+
```
enjeck: Should this be moved to plural `findAllSharesForNodes` only (which is what the count methods need)? This adds sender filtering to the shared singular method, which previously ignored `$sender` and returned all shares for a node. now, it changes every existing caller that passes a user id.

lib/Controller/Api1Controller.php
enjeck, line 1: As a downstream effect of the `findAllSharesForNode` change above, the list all shares" endpoint now returns only shares created by the current user. An owner will no longer see shares created by co-managers on their own table or view, so they cannot review or revoke them here?

enjeck: same for sharecontroller.php

lib/Service/ShareService.php
enjeck, line 1: I think we should have a test for `deleteAllForContext` here. Since this now only iterates shares whose sender equals the current user, so for a context with shares from more than one sender the context_navigation entries for the others are not cleaned up before the bulk deleteByNode, leaving orphaned rows???

lib/Controller/Api1Controller.php
```diff
@@ -326,7 +326,8 @@ public function deleteTable(int $tableId): DataResponse {
 	#[OpenAPI(scope: OpenAPI::SCOPE_DEFAULT)]
 	public function indexViews(int $tableId): DataResponse {
 		try {
-			return new DataResponse($this->viewService->formatViews($this->viewService->findAll($this->tableService->find($tableId))));
+			$table = $this->tableService->find($tableId);
+			return new DataResponse($this->viewService->formatViews($table->getViews() ?? []));
```
enjeck, line 304: initially, we return 403 for read-only users . Now, read-only user gets 200 with an empty list, since enhanceTables only loads views for owned or managed tables. So we need to mofify the annotation?

lib/Db/ColumnMapper.php
```diff
@@ -165,24 +165,30 @@ public function getColumnTypes(array $neededColumnIds): array {
...
+			$qb->select('table_id', $qb->func()->count('*', 'counter'))
+				->from($this->table)
+				->where($qb->expr()->in('table_id', $qb->createNamedParameter($tableIdsChunk, IQueryBuilder::PARAM_INT_ARRAY)))
+				->groupBy('table_id');
+
+			$result = $qb->executeQuery();
+			while ($row = $result->fetch()) {
+				$counts[(int)$row['table_id']] = (int)$row['counter'];
+			}
+			$result->closeCursor();
```
enjeck, line 189: without try...catch, a `\OCP\DB\Exception` now propagates?

lib/Db/ViewMapper.php
```diff
@@ -116,6 +118,25 @@ public function findAll(?int $tableId = null): array {
...
+	public function findAllByTableIds(array $tableIds): array {
+		if (empty($tableIds)) {
+			return [];
+		}
+
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('v.*', 't.ownership')
+			->from($this->table, 'v')
+			->innerJoin('v', 'tables_tables', 't', 't.id = v.table_id')
+			->where($qb->expr()->in('v.table_id', $qb->createNamedParameter($tableIds, IQueryBuilder::PARAM_INT_ARRAY)));
```
enjeck: don't we need same `array_chunk` here? Since every other batch method you added chunks the IN list at about 1000, and findMany does too, but this one passes all table ids in a single IN. What if the user has more than ~1000 tables, or  the DB has a 1000 element IN limit (Oracle does, i think?)??

lib/Db/ColumnMapper.php
Koc: do we really need that try/catch here? When exception should be possible here and we need to suppress it?

lib/Db/ViewMapper.php
Koc: yeah, it has sense, added chunking

Koc:
@enjeck I've fixed conflicts introduced by https://github.com/nextcloud/tables/pull/2781 and code review comments, please, check one more time.

lib/Db/RowSleeveMapper.php
```diff
@@ -106,24 +106,30 @@ public function deleteAllForTable(int $tableId): int {
...
-			$qb->expr()->eq('table_id', $qb->createNamedParameter($tableId))
-		);
-
-		try {
-			$result = $this->findOneQuery($qb);
-			return (int)$result['counter'];
-		} catch (DoesNotExistException|MultipleObjectsReturnedException|Exception $e) {
-			$this->logger->warning('Exception occurred: ' . $e->getMessage() . ' Will return 0.');
-			return 0;
+	public function countRowsForTables(array $tableIds): array {
```
enjeck: So `RowSleeveMapper::countRowsForTables` is now unused code that duplicates `countRowsByTableIds`?

enjeck (commented):
a question:

lib/Db/RowSleeveMapper.php
Koc: yeah, good catch, removed!

lib/Controller/Api1Controller.php
```diff
@@ -296,11 +296,12 @@ public function deleteTable(int $tableId): DataResponse {
 	#[NoAdminRequired]
 	#[NoCSRFRequired]
 	#[CORS]
-	#[RequirePermission(permission: Application::PERMISSION_READ, type: Application::NODE_TYPE_TABLE, idParam: 'tableId')]
+	#[RequirePermission(permission: Application::PERMISSION_MANAGE, type: Application::NODE_TYPE_TABLE, idParam: 'tableId')]
```
blizzz: Oh, how come?

lib/Db/ViewMapper.php
```diff
@@ -146,6 +148,30 @@ public function findAll(?int $tableId = null): array {
...
+			return [];
+		}
+
+		$chunks = [[]];
+		foreach (array_chunk($tableIds, self::DB_CHUNK_SIZE - 1) as $tableIdsChunk) {
+			$qb = $this->db->getQueryBuilder();
+			$qb->select('v.*', 't.ownership')
+				->from($this->table, 'v')
+				->innerJoin('v', 'tables_tables', 't', 't.id = v.table_id')
+				->where($qb->expr()->in('v.table_id', $qb->createNamedParameter($tableIdsChunk, IQueryBuilder::PARAM_INT_ARRAY)));
```
blizzz, line 167: The order by statements are missing:

```
$qb->addOrderBy('v.sidebar_order', 'ASC');
$qb->addOrderBy('v.id', 'ASC');
```

lib/Service/ViewService.php
```diff
@@ -415,55 +437,115 @@ public function deleteByObject(View $view, ?string $userId = null): View {
...
+	/**
+	 * add some basic values related to a list of views in context
+	 *
+	 * $userId can be set or ''
+	 *
+	 * @param View[] $views
+	 * @param array<int, int|null> $tableRowsCounts
+	 */
+	private function enhanceViews(array $views, string $userId, array $tableRowsCounts = []): void {
+		if (empty($views)) {
```
blizzz, line 461: this does not have the isFederated guard anymore. Intentional? It puts the responsibility to check about it to the caller, like in `enhanceView()` currently.

lib/Service/ShareService.php
```diff
@@ -256,6 +254,72 @@ public function findTablesSharedWithMe(?string $userId = null): array {
...
+	 * @param string|null $userId
+	 * @return array<int, int>
+	 * @throws InternalError
+	 */
+	public function countSharesForTables(array $tableIds, ?string $userId = null): array {
+		$userId = $this->permissionsService->preCheckUserId($userId);
+
+		try {
+			$excluded = !$this->circleHelper->isCirclesEnabled() ? [ShareReceiverType::CIRCLE] : [];
+			$shares = $this->mapper->findAllSharesForNodes('table', $tableIds, $userId, $excluded);
```
blizzz, line 268: Does this still find all shares for a table manager, i.e. indirect shares?

lib/Service/TableService.php
```diff
@@ -169,55 +161,107 @@ public function formatTables(array $tables): array {
...
-		// add the rows count
-		if ($rowCountsByTableId !== null) {
-			$table->setRowsCount($rowCountsByTableId[$table->getId()] ?? 0);
-		} else {
-			try {
-				$table->setRowsCount($this->rowService->getRowsCount($table->getId()));
-			} catch (PermissionError) {
-				$table->setRowsCount(0);
+		// if the table is shared with me, there are no other shares
+		// will avoid showing the shared icon in the FE nav
```
blizzz, line 233: is this intentional? Sounds strange?

blizzz (changes requested):
Some questions, and view ordering got lost it seems

PR #2878, opened 2026-08-08
fix(relation): Fix ellipsis for long selected relation labels

Koc:
This PR fixes display of the relation fields with long relation label for add/edit modal and inline edit.

### 🖼️ Screenshots

🏚️ Before | 🏡 After | note
---|--- | ---
image | image | relation field has small width, there is an overlap between text and select icons
image | image | column has different width when we selected editing 2 or 3 rows

### 🏁 Checklist
 
- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


samin-z:
ran 'npm run build' and hard refreshed, but i still see the same bug
in chrome:

Screenshot 2026-08-10 at 17 53 36

in firefox:
Screenshot 2026-08-10 at 17 54 46

Koc:
@samin-z I guess because you're using single word. In my test scenario there were multiple words

samin-z:
@Koc the issue is that the input field is flexible and exapnds with the content (has nothing to do with if its a single words or multiple words), so the input box expands to the width of the 'edit row' box and if it's bigger than that, the text still mixes with the 'dropdown' and 'cancel' icon.
"edit row" maximum box input length:
Screenshot 2026-08-11 at 09 53 47
smaller than "edit row" maximum bix input length: 
Screenshot 2026-08-11 at 09 54 07

enjeck (commented):
no longer needed since https://github.com/nextcloud-libraries/nextcloud-vue/pull/8829 is merged???

Koc:
I will update library and double-check

Koc:
I've updated libraries but it seems like this is not enough. Here what I've achieved in current branch with override:

image

image

image

@samin-z please test on your side. If everything ok - I will move changes from the current PR back to libraries.

samin-z:
> he issue is that the input field is flexible and exapnds with the content (has nothing to do with if its a single words or multiple words), so the input box expands to the width of the 'edit row' box and if it's bigger than that, the text still mixes with the 'dropdown' and 'cancel' icon.

@Koc the issue still stands.

Koc:
@samin-z theck one more time, please. I've tested in Firefox and Chrome

blizzz:
@samin-z I cannot repro it either on current state, tried with Firefox.

Koc:
also it seems like fix should stay in this repo and can't be moved into libraries repository

samin-z:
> @samin-z I cannot repro it either on current state, tried with Firefox.

@blizzz the browser is not the case, it's the size of the screen that you can reproduce it with, i tried on monitor other than my laptop.

PR #2882, opened 2026-08-09
fix: date-only column display shifted one day behind in non-UTC timezones

### LINKED ISSUE #2705: Viewing date in column 1 day off from date entered
### Steps to reproduce

1. Create table
2. Create date only column (not time and date, just date when adding column)
3. Create row entry, enter date

### Expected behavior

Date shown to be selected date

### Actual behavior

Date shown in table view is 1 day less (example: "Jun 26, 1985" instead of selected 1985.06.27), but clicking on date to edit or clicking 3 dots/meatball menu and "Edit row" shows correct date in the field/picker (1985.06.27).
(Also shows 1 day less for other dates unless editing, too, not just the example date)

### Tables app version

2.2.0

### Browser

Brave 1.90.128 (Official Build) (arm64) Chromium: 148.0.7778.217

### Client operating system

macOS Tahoe 26.6

### Operating system

AlmaLinux 9.8 (Olive Jaguar) (Linux 5.14.0-687.5.4.el9_8.x86_64 x86_64)

### Web server

Apache

### PHP engine version

PHP 8.3

### Database

PostgreSQL

### Additional info

Nextcloud AIO latest (just updated), Nextcloud Hub 26 Winter (33.0.3).

date and time (instead of just date) appears to work correctly, just the columns where date only is set is where this odd issue is observed.

happens when either entering manually or using api to add row, right value is input, 1 day before is shown, unless editing.

tareko:
## Bug

The discovery of this bug and code to fix it were assisted by AI: GLM 5.2

Date-only columns display the date one day behind the entered value for users in timezones behind UTC (e.g. America/Toronto UTC-4).

**Steps to reproduce:**
1. Create a date-only column (not datetime, just date)
2. Enter a date (e.g. 2025-08-25)
3. The table view shows the previous day (e.g. Aug 24, 2025)
4. Clicking to edit shows the correct date — the bug is only in the display

## Root cause

In `datetimeDate.js`, `formatValue` parsed date-only values (e.g. `"2025-08-25"`) using the **datetime** format `'YYYY-MM-DD HH:mm:ss'`:

```js
formatValue(value) {
    return Moment(value, 'YYYY-MM-DD HH:mm:ss').format('ll')
}
```

Since the input string doesn't match the format (no time component), Moment.js falls back to ISO 8601 parsing, which interprets date-only strings as **UTC midnight**. When `.format('ll')` then converts to the user's local timezone, the date shifts back one day for users behind UTC.

## Fix

Use `'YYYY-MM-DD'` as the Moment parse format — this matches the actual stored value format and ensures local-timezone parsing:

```js
formatValue(value) {
    return Moment(value, 'YYYY-MM-DD').format('ll')
}
```

This is consistent with `getDateFormat()` in `TableCellDateTime.vue`, which already uses `'YYYY-MM-DD'` for the edit mode — explaining why editing showed the correct date while the cell display didn't.

Fixes #2705

### DIFF
--- a/src/shared/components/ncTable/mixins/columnsTypes/datetimeDate.js
+++ b/src/shared/components/ncTable/mixins/columnsTypes/datetimeDate.js
@@ -15,7 +15,7 @@ export default class DatetimeDateColumn extends AbstractDatetimeColumn {
 	}
 
 	formatValue(value) {
-		return Moment(value, 'YYYY-MM-DD HH:mm:ss').format('ll')
+		return Moment(value, 'YYYY-MM-DD').format('ll')
 	}
 
 	sort(mode, nextSorts) {


enjeck (commented):
would be nice to see video of this working properly

samin-z:
not sure if this is still valid as im not able to reproduce it (or maybe it has already been fixed).
here is the screen recording of my local on main branch. were you able to reproduce it? @tareko 

https://github.com/user-attachments/assets/0eea8524-07b8-435c-bf32-4febf78456e2

tareko:
> not sure if this is still valid as im not able to reproduce it (or maybe it has already been fixed).
> here is the screen recording of my local on main branch. were you able to reproduce it? @tareko

What time zone are you in, @samin-z ?

samin-z:
> What time zone are you in, @samin-z ?

@tareko its CET

UPDTAE: i also tried it with different timezones and it still works properly. please let me know of the intended time zone.

PR #2885, opened 2026-08-09
feat(navigation): add LoadAdditionalEntriesListener

enjeck:
Tables' context navigation tests have been broken for a while due to changes in the server. It should be fixed in the server, but it's still flaky on our side, as it passes for me locally but fails here on CI. Hoping that following the advice given to use  LoadAdditionalEntriesListener  fixes it. See https://github.com/nextcloud/server/pull/61120


lib/AppInfo/Application.php
```diff
@@ -85,7 +85,7 @@ public function register(IRegistrationContext $context): void {
 		$context->registerEventListener(BeforeUserDeletedEvent::class, UserDeletedListener::class);
 		$context->registerEventListener(DatasourceEvent::class, AnalyticsDatasourceListener::class);
 		$context->registerEventListener(RenderReferenceEvent::class, TablesReferenceListener::class);
-		$context->registerEventListener(BeforeTemplateRenderedEvent::class, BeforeTemplateRenderedListener::class);
+		$context->registerEventListener(LoadAdditionalEntriesEvent::class, LoadAdditionalEntriesListener::class);
```
mejo-, line 88: I guess that the `BeforeTemplateRenderedListener` class still exists. Is it obsolete now that it's not loaded anymore? Or is it still used somewhere else?

Koc: I've removed it

Koc (approved):

PR #2888, opened 2026-08-12
Table scheme import

### LINKED ISSUE #2840: Structure update
As a process owner, I want to be able to work on a table structure on a cloned table, and only when my changes are ready, apply them on the production table. The cloned table might be a re-imported table on the same instance, or on a staging instance.

Furthermore, I want to be able to do the same against Applications/Context.

With https://github.com/nextcloud/tables/pull/2598 we have unique identifiers of columns, selection options and views, that are abound to the parent table. With this we can identify these elements across instances and even local tables. When exporting and importing a table structures, the UUID are kept in place.

There is prior work at #2504 which was based on a tables version without UUIDs, and could not work in a satisfactory way. Maybe some parts of the PR still can be re-used, perhaps the UI.

What needs to be done:

- Front-end wise
  - [ ] allow a table manager to import a scheme from a json file onto the selected table
  - [ ] allow an application manager to export the scheme of an application
  - [ ] allow an application manager to import a scheme from a json file onto the selected table
  - [ ] on importing, the application or table manager should be able to see the changes, and exclude the ones they do not want to import
- Backend wise
  - [ ] resolve changes between the original and the uploaded table structure
  - [ ] resolve changes between the original and the uploaded application structure
  - [ ] apply selected changes against the specified table
  - [ ] apply selected changes against the specified application

The focus is first on importing against a table. Updating the application structure is a second step.

luka-nextcloud:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist
 
- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI


lib/Migration/Version2200Date20260819000000.php
```diff
@@ -0,0 +1,75 @@
...
+		$select = $qbSelect->executeQuery();
+
+		try {
+			$this->connection->beginTransaction();
+			while (($row = $select->fetchAssociative()) !== false) {
+				$qb->setParameter('tableLocalId', $row['id']);
+				$qb->setParameter('uuid', Uuid::v7()->toRfc4122());
+				$qb->executeStatement();
+			}
+			$this->connection->commit();
```
blizzz, line 78: In `Version2020Date20260513185340` I commit and start a new transaction every 250 writes for performance reasons. Could be an improvement here as well, beneficiary on updates on instances with many tables. But will work, can be addressed later.

lib/Controller/ContextController.php
```diff
@@ -282,6 +300,132 @@ public function updateContentOrder(int $contextId, int $pageId, array $content):
...
+	 * @return DataResponse<Http::STATUS_OK, array<string, mixed>, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR|Http::STATUS_NOT_FOUND|Http::STATUS_FORBIDDEN, array{message: string}, array{}>
+	 *
+	 * @CanManageContext
+	 *
+	 * 200: returning the scheme of the context
+	 * 403: No permissions
+	 * 404: Not found
+	 */
+	#[NoAdminRequired]
+	#[RequirePermission(Application::PERMISSION_MANAGE, null, 'context', 'contextId')]
```
blizzz: nitpick, also can be done in a follow up, just for readability it think this is nicer:

```suggestion
	#[RequirePermission(permission: Application::PERMISSION_MANAGE, typeParam: 'context', idParam: 'contextId')]
```

applies to the other similar lines as well.

(yes it was used somewhere like this here before, that can also be improved)

Again, not a blocker, by-catch, good enough with follow-up.

lib/Controller/ContextController.php
```diff
@@ -282,6 +300,132 @@ public function updateContentOrder(int $contextId, int $pageId, array $content):
...
+	 *
+	 * @CanManageContext
+	 *
+	 * 200: context updated successfully
+	 * 403: No permissions
+	 * 404: Not found
+	 */
+	#[NoAdminRequired]
+	#[RequirePermission(Application::PERMISSION_MANAGE, null, 'context', 'contextId')]
+	public function importScheme(int $contextId, ?string $name, ?string $iconName, ?string $description, ?array $nodes, ?array $tables): DataResponse {
```
blizzz: Node sure I understand correctly the role of `$nodes`  vs `$tables`. Maybe have an idea with the TODO against the nodes.

lib/Service/ColumnService.php
```diff
@@ -820,4 +820,31 @@ private function handleColumnPersistDbException(\OCP\DB\Exception $e, string $co
...
+	 *
+	 * @throws BadRequestError
+	 * @throws InternalError
+	 */
+	public function importColumns(Table $table, array $columns): array {
+		$importedColumns = [];
+		foreach ($columns as $column) {
+			if (isset($column['uuid'])) {
+				try {
+					$existColumn = $this->mapper->findByUuid($column['uuid']);
```
blizzz: Is fetching by UUID important here? My notion is, as with the StructureService, to create a diff between the incoming and existing column and then add, modify (or keep), and remove. This requires checking what is there, then we do need to fetch by UUID (which done one-by-one as also rather expensive, and we have intentionally no index there).

lib/Service/StructureService.php
```diff
@@ -0,0 +1,396 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Tables\Service;
+
+use OCA\Tables\Errors\InternalError;
+use OCA\Tables\Errors\NotFoundError;
+use OCA\Tables\Errors\PermissionError;
+
+class StructureService {
```
blizzz, line 16: I know it comes from me, just to note down one thought that circles my mind is whether to add Stateful to the class name for safer usage 🤔 Again, nothing that cannot be improved upon later on.

lib/Service/TableService.php
```diff
@@ -569,8 +576,8 @@ public function search(string $term, int $limit = 100, int $offset = 0, ?string
 	public function getScheme(int $id, ?string $userId = null): TableScheme {
 		$table = $this->find($id, skipTableEnhancement: true);
 		$columns = $this->columnService->findAllByTable($id, null, $table);
-		$this->enhanceTable($table, $userId);
-		return new TableScheme($table->getTitle(), $table->getEmoji(), $columns, $table->getViews() ?: [], $table->getDescription() ?: '', $this->appManager->getAppVersion('tables'), $table->getColumnOrderSettingsArray(), $table->getSortArray());
+		$this->enhanceTable($table, $userId ?? $this->userId);
```
blizzz: i suppose there is a reason for this line change? ^

lib/Service/ViewService.php
```diff
@@ -709,10 +710,82 @@ private function assertTechnicalNameValid(string $technicalName): void {
 	 */
 	private function handleViewPersistDbException(\OCP\DB\Exception $e, string $context): never {
 		if ($e->getReason() === \OCP\DB\Exception::REASON_UNIQUE_CONSTRAINT_VIOLATION) {
-			throw new BadRequestError('Technical name must be unique in the table.');
+			throw new BadRequestError('Technical name must be unique in the table. ' . $e->getMessage());
```
blizzz: intentional line change?

blizzz (commented):
Early comments, some can be left for later.

UI wise it looks really nice already!

Image

* [x] I would not display the UUID as title in the check box. It looks technical and table managers will have not deeper knowledge about it. Would completely keeps this identifier internal and instead use the **New Title**. 

* [x] In my quick test I also see that Views would be dropped instead of being kept

* [ ] Maybe need to double check the sort order as well when nothing is defined

This is very promising already! :)

lib/Controller/ContextController.php
luka-nextcloud: `$nodes` is array ids of tables and views to be shown. The `$tables` is array of related source tables.

lib/Service/StructureService.php
luka-nextcloud: I think it can be improved later.

luka-nextcloud:
@blizzz 
> In my quick test I also see that Views would be dropped instead of being kept

Does it mean that we should remove the `Remove Views` section?
image

> Maybe need to double check the sort order as well when nothing is defined

Could you please share more details?

lib/Service/TableService.php
luka-nextcloud: It would cause error when `$userId` is null

blizzz:
> @blizzz
> 
> > In my quick test I also see that Views would be dropped instead of being kept
> 
> Does it mean that we should remove the `Remove Views` section? 

No, that should be there!

In my test, the scheme file I uploaded did not have any changes with the views. However the dialog showed they would be removed, instead of kept.

luka-nextcloud:
> > @blizzz
> > > In my quick test I also see that Views would be dropped instead of being kept
> > 
> > 
> > Does it mean that we should remove the `Remove Views` section?
> 
> No, that should be there!
> 
> In my test, the scheme file I uploaded did not have any changes with the views. However the dialog showed they would be removed, instead of kept.

The application scheme includes the table schemes of related source tables. So it also detects the different views of those tables. Should we skip processing removed views when importing application scheme?

blizzz:
> schemes



> > > @blizzz
> > > > In my quick test I also see that Views would be dropped instead of being kept
> > > 
> > > 
> > > Does it mean that we should remove the `Remove Views` section?
> > 
> > 
> > No, that should be there!
> > In my test, the scheme file I uploaded did not have any changes with the views. However the dialog showed they would be removed, instead of kept.
> 
> The application scheme includes the table schemes of related source tables. So it also detects the different views of those tables. Should we skip processing removed views when importing application scheme?

Views that are missing in the scheme that is being imported shall be removed. In my case the detection was not correct, since they were still present, they should not be shown to be removed.

luka-nextcloud:
@blizzz Do you mean that the view does present in the scheme but shown to be removed? It is strange, I cannot reproduce this issue. Could you please share the reproduce steps, as well as the scheme file you used and the uuid of the effected view?

tests/unit/Service/StructureServiceTest.php
```diff
@@ -0,0 +1,35 @@
...
+
+class StructureServiceTest extends TestCase {
+	private readonly StructureService $service;
+	private readonly array $originalSchema;
+	private readonly TableService|MockObject $tableService;
+
+	public function setUp(): void {
+		$this->tableService = $this->createMock(TableService::class);
+
+		$this->originalSchema = json_decode(file_get_contents(__DIR__ . '/../res/feature-list-schema.json'), true);
```
blizzz, line 24: i did not add this file, huh? 🙈

blizzz:
[Screencast_20260821_123155.webm](https://github.com/user-attachments/assets/ce17861b-59bd-4cfe-ba39-4a2481d399d2)

In this recording you see an imported schema with two views.

You also see the modified scheme file with only one selection options removed, another added, otherwise unchanged.

The UI suggests that the views were removed though.

luka-nextcloud:
@blizzz I am still unable to reproduce this issue. In the video, I noticed that 2 views were detected as new views with the same name as 2 removed views. So, the only possible reason is that they have different uuid.

tests/unit/Service/StructureServiceTest.php
luka-nextcloud: @blizzz Yeah... you didn't add that file 😅

blizzz:
> @blizzz I am still unable to reproduce this issue. In the video, I noticed that 2 views were detected as new views with the same name as 2 removed views. So, the only possible reason is that they have different uuid.

In the screen recording you see that there is no difference in the views. I attach the schema files.

But you are right that they differ when you compare them to the database entry:

Screenshot_20260821_194233

That means when doing the original import, the UUIDs were not taken over, but re-generated.

blizzz:
Also checked columns and selection options, but there the UUID is taken over during inital import.

P.S.: The scheme file i tested with did not contain a UUID for the table,  so might apply there as well. And did not try with a Context yet.

P.P.S.: but then it should also not only show the Views to remove, but also the "new" ones  to add, right?

lib/Service/ContextService.php
```diff
@@ -644,4 +651,173 @@ public function importContext(Table $table, array $context, int $tableOldId): ?C
...
+
+		return $changes;
+	}
+
+	public function importScheme(int $contextId, string $name, string $iconName, string $description, array $nodes, array $tables, string $userId): Context {
+		// Validate the structure of the columns and views arrays
+		if (!isset($tables['addTables']) || !is_array($tables['addTables'])
+			|| !isset($tables['modifyTables']) || !is_array($tables['modifyTables'])) {
+			throw new BadRequestError('Invalid tables structure provided.');
+		}
```
blizzz, line 807: removing tables needs to be added as well, right?

luka-nextcloud: We should not remove the tables since they might be used in another application. Updating $node already reflect the elements to show on that application.

blizzz:
I added two small commits with small fixes. 

A bit unsure about https://github.com/nextcloud/tables/pull/2888/commits/f8e2e1c718202b4520660cda59b8862c91ad5c98 – here i soften the check and instead of erroring out, simply discard the submitted UUID. Easier on user front-ends. Also, when i would all passing this parameter, then the remaining call is set on true, but we have more entry points here afaik. cc @samin-z  – FYI you had the error case and we talked about that as well.

More todos:

- [ ] we need https://github.com/nextcloud/tables/pull/2899 extended with a migration step to add UUIDs to selection options where they are missing
  - background: newly created selection options do not get the UUID set upon creation, hence they are not being treated correctly 
- [ ] after an import on the web interface, the data is not being updated/reloaded automatically (e.g. when a table title is being changed). 

I still only tested a little bit, but spent maybe too long on those small changes (investigation time…).

lib/Model/FilterSet.php
```diff
@@ -27,9 +28,20 @@ public function __construct(
 		}
 	}
 
-	public static function createFromInputArray(array $data): self {
+	public static function createFromInputArray(array $data, array $columnsMap = []): self {
 		$filterGroups = [];
 		foreach ($data as $inputFilterGroup) {
+			foreach ($inputFilterGroup as $j => $item) {
+				if (isset($columnsMap[$item['columnUuid']]) && $columnsMap[$item['columnUuid']] instanceof Column) {
```
enjeck: Would this error out in the normal path where filters only have `columnUuid`? Since `$item['columnUuid'] is read with no check, i expect the normal paths where filters only carry columnId triggers an "Undefined array key" warning for every filter item.

enjeck: can add `isset($...['columnUuid'])`

lib/Service/ContextService.php
```diff
@@ -644,4 +651,173 @@ public function importContext(Table $table, array $context, int $tableOldId): ?C
...
+		$columnService = \OCP\Server::get(ColumnService::class);
+
+		foreach ($tables['addTables'] as $tableScheme) {
+			$table = $tableService->importTable($tableScheme, $userId);
+			foreach ($tableScheme['views'] as $viewData) {
+				$viewService->importView($table->getId(), $viewData, $userId);
+			}
+			foreach ($tableScheme['columns'] as $columnData) {
+				$columnService->importColumn($table, $columnData);
+			}
```
enjeck: If we import the view before we import the columns, the new columns get fresh ids, so these references break and view layout, filters, and sorts are lost??

lib/Controller/ContextController.php
```diff
@@ -282,6 +299,102 @@ public function updateContentOrder(int $contextId, int $pageId, array $content):
...
+	#[NoAdminRequired]
+	#[RequirePermission(permission: Application::PERMISSION_MANAGE, typeParam: 'context', idParam: 'contextId')]
+	public function previewSchemeChanges(int $contextId, array $updateScheme): DataResponse {
+		try {
+			$changes = $this->contextService->compareSchemeChanges($contextId, $updateScheme, $this->userId);
+			return new DataResponse($changes);
+		} catch (NotFoundError $e) {
+			return $this->handleNotFoundError($e);
+		} catch (PermissionError $e) {
+			$this->db->rollBack();
```
enjeck: how can we rollBack if we never `beginTransaction()`? 🤔

lib/Model/ContextScheme.php
```diff
@@ -0,0 +1,64 @@
...
+		$this->name = $name;
+		$this->icon = $icon;
+		$this->description = $description;
+		$this->nodes = $nodes;
+		$this->pages = $pages;
+		$this->tables = $tables;
+	}
+
+	public function getName(): ?string {
+		return $this->name | '';
```
enjeck: Since `$this->name | ''` is a bitwise string OR, not a default. And `null | ''` is typeerror:

```suggestion
		return $this->name ?? '';
```

lib/Controller/ApiTablesController.php
```diff
@@ -286,6 +293,91 @@ public function createFromScheme(string $title, string $emoji, string $descripti
...
+			}
+			return $this->handleBadRequestError($e);
+		} catch (InternalError|Exception $e) {
+			try {
+				$this->db->rollBack();
+			} catch (\OCP\DB\Exception $e) {
+				return $this->handleError($e);
+			}
+			return $this->handleError($e);
+		}
```
enjeck: can we also have a `\Throwable` catch that rolls back? Becuse `catch (InternalError|Exception $e)` does not cover `\Throwable`, so a `TypeError` from a bad payload fails without hitting `rollBack()`

lib/Controller/ContextController.php
```diff
@@ -282,6 +299,102 @@ public function updateContentOrder(int $contextId, int $pageId, array $content):
...
+			return $this->handlePermissionError($e);
+		} catch (Exception|InternalError|\Throwable $e) {
+			return $this->handleError($e);
+		}
```
enjeck, line 357: can we also have a `\Throwable` catch that rolls back? Becuse `catch (InternalError|Exception $e)` does not cover `\Throwable`, so a `TypeError` from a bad payload fails without hitting `rollBack()`

tests/unit/Service/StructureServiceTest.php
enjeck, line 1: do we want to add tests for added, removed, and modified columns and views, plus columnOrder and sort resolution?

lib/Service/TableService.php
```diff
@@ -611,4 +620,148 @@ public function importTable(array $table, string $userId): Table {
...
+		}
+
+		// Add new columns
+		foreach ($columns['addColumns'] as $columnData) {
+			$this->columnService->importColumn($table, $columnData);
+		}
+
+		// Remove columns
+		foreach ($columns['removeColumns'] as $columnData) {
+			$this->columnService->delete($columnData['id']);
```
enjeck, line 749: `removeColumns[].id` comes straight from the client and is passed to `delete()` without checking the column belongs to `$tableId`. Risky...

lib/Service/TableService.php
```diff
@@ -611,4 +620,148 @@ public function importTable(array $table, string $userId): Table {
...
+		}
+
+		// Update existing columns
+		foreach ($columns['modifyColumns'] as $columnData) {
+			$fromColumn = $columnData['from'];
+			$toColumn = $columnData['to'];
+			$this->columnService->update($fromColumn['id'], $userId, ColumnDto::createFromArray($toColumn));
```
enjeck, line 760: `modifyColumns[].from.id` comes straight from the client and is passed to `modify()` without checking the column belongs to `$tableId`. Risky?

lib/Service/TableService.php
```diff
@@ -611,4 +620,148 @@ public function importTable(array $table, string $userId): Table {
...
+		foreach ($views['modifyViews'] as $item) {
+			$fromView = $item['from'];
+			$toView = $item['to'];
+			$existingView = $this->viewService->find($fromView['id'], userId: $userId);
+			$this->viewService->update($existingView->getId(), ViewUpdateInput::fromInputArray($toView, $columnsMap), $userId);
+		}
+
+		// Remove views
+		foreach ($views['removeViews'] as $viewData) {
+			$this->viewService->delete($viewData['id']);
```
enjeck, line 806: `removeViews[].id` comes straight from the client and is passed to `delete()` without checking the column belongs to `$tableId`. Risky?

lib/Service/StructureService.php
```diff
@@ -0,0 +1,392 @@
...
+				$this->columnOrderChanges['to'][] = [
+					'columnUuid' => $column['uuid'],
+					'columnTitle' => $column['title'],
+					'order' => $order++,
+				];
+			}
+		} else {
+			foreach ($updateSchema['columnOrder'] as $item) {
+				$this->columnOrderChanges['to'][] = [
+					'columnUuid' => $updateColumnsMap[$item['columnId']]['uuid'],
```
enjeck, line 359: `$updateColumnsMap[$item['columnId']]['uuid']` has no `isset` guard here, unlike the view remap loop above

lib/Model/SelectionOption.php
```diff
@@ -31,7 +31,7 @@ public static function createFromInputArray(array $data, bool $allowPassingUuid
 		}
 
 		if (isset($data['uuid']) && !$allowPassingUuid) {
-			throw new \InvalidArgumentException('It is forbidden to set the Uuid from external');
+			unset($data['uuid']);
```
enjeck, line 34: SHould we only do this for the import path? Cuz otherwise, by silently dropping the an externally supplied `uuid`, we a real client bug now passes quietly

lib/Service/ContextService.php
```diff
@@ -644,4 +651,173 @@ public function importContext(Table $table, array $context, int $tableOldId): ?C
...
+		}
+
+		// Detect modify tables
+		foreach ($updateScheme['tables'] as $tableUpdateScheme) {
+			if (isset($tablesMap[$tableUpdateScheme['uuid']])) {
+				$tableLocalId = $tablesMap[$tableUpdateScheme['uuid']]->getId();
+				$tableLocalScheme = $tableService->getScheme($tableLocalId, $userId)->jsonSerialize();
+				if (json_encode($tableLocalScheme) !== json_encode($tableUpdateScheme)) {
+					$changes['modifyTables'][$tableUpdateScheme['uuid']] = $tableService->compareTableSchemeChanges($tableLocalId, $tableUpdateScheme);
```
enjeck: Comparing tables with `json_encode($a) !== json_encode($b)` is key order sensitive? So will report false modifications?

lib/Service/ContextService.php
```diff
@@ -644,4 +651,173 @@ public function importContext(Table $table, array $context, int $tableOldId): ?C
...
+		];
+
+		$tables = $tableService->findAll();
+		$tablesMap = [];
+		foreach ($tables as $table) {
+			$tablesMap[$table->getUuid()] = $table;
+		}
+
+		// Detect new tables
+		foreach ($updateScheme['tables'] as $table) {
```
enjeck: `$updateScheme['tables']` is read without `isset`, so a malformed body returns a 500 instead of a 400.

lib/Service/ContextService.php
```diff
@@ -644,4 +651,173 @@ public function importContext(Table $table, array $context, int $tableOldId): ?C
...
+			}
+			foreach ($tableScheme['columns'] as $columnData) {
+				$columnService->importColumn($table, $columnData);
+			}
+		}
+
+		foreach ($tables['modifyTables'] as $tableScheme) {
+			$table = $this->tableMapper->findByUuid($tableScheme['uuid']);
+			$tableService->updateTableStructure($table->getId(), $tableScheme['columns'], $tableScheme['views'], $tableScheme['columnOrder'], $tableScheme['sort'], $userId);
+		}
```
enjeck, line 839: $tableScheme['columnOrder'] and ['sort'] are read without `isset`, so a malformed body returns a 500 instead of a 400.

lib/Service/ContextService.php
```diff
@@ -644,4 +651,173 @@ public function importContext(Table $table, array $context, int $tableOldId): ?C
...
+	 * @throws NotFoundExceptionInterface
+	 */
+	public function getScheme(int $contextId, string $userId): ContextScheme {
+		$context = $this->contextMapper->findById($contextId);
+		$tableService = \OCP\Server::get(TableService::class);
+		$viewService = \OCP\Server::get(ViewService::class);
+		$nodes = $context->getNodes();
+		$tables = [];
+
+		foreach ($nodes as &$node) {
```
enjeck, line 675: This leaves a dangling reference, so we should `unset($node)` after the loop?

src/modules/modals/ImportTableScheme.vue
```diff
@@ -0,0 +1,392 @@
...
+				const url = generateOcsUrl('/apps/tables/api/2/tables/' + this.table.id + '/scheme/import')
+				const res = await axios.post(url, this.schemeData)
+
+				if (res.status === 200) {
+					this.importInitialized = true
+					this.result = res.data
+				} else {
+					console.debug('error while importing', res)
+					this.importFailed = true
+					this.errorMessage = t('tables', res.data?.message || 'Could not import scheme due to unknown errors.')
```
enjeck: I don't think this dynamic text is acceptable for translation

src/modules/modals/ImportTableScheme.vue
```diff
@@ -0,0 +1,392 @@
...
+			} else if (res?.status === 403) {
+				console.debug('error while importing', e || res)
+				this.errorMessage = t('tables', 'Could not import, missing needed permission.')
+			} else if (res?.status === 404) {
+				console.debug('error while importing', e || res)
+				this.errorMessage = t('tables', 'Could not import, needed resources were not found.')
+			} else {
+				console.debug('error while importing', e || res)
+				this.errorMessage = t('tables', e?.response?.data?.message || 'Could not import scheme due to unknown errors.')
+			}
```
enjeck, line 342: I don't think this dynamic text is acceptable for translation

enjeck (commented):
lots of comments...

enjeck (dismissed):

lib/Service/ContextService.php
luka-nextcloud: It is unnecessary, we are not using the `$node` after loop.

lib/Model/SelectionOption.php
luka-nextcloud: @blizzz Could you please check?
@enjeck I'm not sure why we have to drop external uuid in this case. For now, I don't see the impact since the SelectionOption has no uuid in the exported table scheme.

blizzz: We change this in https://github.com/nextcloud/tables/pull/2899 and also drop the unsetting. I did not test willfully overwriting it, it requires manager permissions though, hence OK, might re-visit later.

blizzz:
- [x] Views are shown as changed, because the `columnUuid` is not exposed in columnSettings in the export.
  - I am also seeing that in columnSettings the columnUuid is not being exported, hence not present on import. It appears as a changed column. Views are being updated. At least cannot smuggle in a foreign column :)
- [x] It must be possible to import a table with an already present UUID. The idea is having portable tables that can be updated. The issue that we have here is that we require table UUIDs for Contexts, fortunately only for them!  The limitation currently makes it impossible to have a copy of a table on the same instance (main use case is across instances, but not everybody may have that). 
  - My suggestion would be to catch the constrain in Table scheme import and regenerate the UUID. Or generate a new UUID for imported tables in general. But only for the Table! This should not cause any other issues when importing a scheme into a table (and we did not have it before – only added for Contexts.).
- [x] after importing a reload takes place, but it contains the old state 
- [ ] Contexts: *Show in app list* setting is not taken over (survivable)
- :heavy_check_mark: Context shares are not taken over, but this actually makes sense, with importing across instances in mind.
- i am not sure about the column order. I do not see we can change it in tables (but I recall there was a feature). The JSON exports fills the order, but importing shows the changes, but it does not have any effect.

lib/Migration/Version2200Date20260819000000.php
```diff
@@ -0,0 +1,84 @@
...
+		$table = $schema->getTable('tables_tables');
+		if (!$table->hasColumn('uuid')) {
+			$table->addColumn('uuid', Types::STRING, [
+				'notnull' => false,
+				'default' => null,
+				'length' => 36,
+				'comment' => 'UUIDv7 identifier to support structural updates across instances',
+			]);
+		}
+		if (!$table->hasUniqueConstraint('tables_tables_uuid_uniq')) {
```
blizzz: should be `hasIndex`, I did the same mistake before 🙈

playwright/e2e/tables-import-export-scheme.spec.ts
```diff
@@ -265,7 +266,6 @@ test.describe('Import Export Scheme', () => {
 
 	test('Export scheme to json', async ({ userPage: { page } }) => {
 		const sourceScheme = await getTutorialScheme(page)
-		prepareSchemeForImport(sourceScheme, 'Imported scheme export')
```
blizzz, line 268: tests succeed locally when i add this back

blizzz (approved):

Issue #2897, opened 2026-08-18
selection column can not be saved: API rejects option uuid


samin-z:
### Steps to reproduce

**Description**
Saving an existing selection (options) column fails.
The list payload now includes a `uuid` on each option. The edit column dialog sends that same object back on update. The backend then rejects it with:
`It is forbidden to set the Uuid from external`
No column change is saved.

This does **not** happen for every options column. It only happens when the stored options already include a `uuid` (for example after schema import, or after the UUID migration). Newly created options columns without `uuid` on each option still save.
The edit dialog sends the column payload back as returned by the API, including `selectionOptions[].uuid`. `SelectionOption::createFromInputArray()` rejects `uuid` unless `$allowPassingUuid` is true.

**Steps to reproduce**
1. Open a table with a selection/options column (or create one, then edit it).
2. Edit the column (title, options, etc.).
3. Click Save.

### Expected behavior

The column is updated. Existing option UUIDs should be kept/ignored by the update API, not treated as a client trying to set a new UUID.

### Actual behavior

`PUT /apps/tables/api/1/columns/{id}` fails.
Example body:
```json
{
  "mandatory": false,
  "title": "Brand (status)",
  "technicalName": "column_453",
  "description": "",
  "selectionDefault": "null",
  "selectionOptions": "[{\"id\":0,\"label\":\"➡️ Not started \",\"uuid\":\"019f90c5-4576-7d0b-9b14-9bf49b2826cf\"},{\"id\":1,\"label\":\"⌛ In progress\",\"uuid\":\"019390c5-8576-7f0b-9b14-9bfb94975e\"}]"
}

PR #2899, opened 2026-08-19
fix: ignore client sent option UUID and generate new UUID on column update

samin-z:
Issue: #2897 

**Behaviour:**

Upon column update, any UUID sent in the request payload is ignored (to prevent client manipulation and to preserve the server managed identity). existing options keep their stored UUID and new options get a newly generated one.

### 🏁 Checklist
 
- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


lib/Model/SelectionOption.php
```diff
@@ -49,7 +47,29 @@ public function label(): string {
...
+	 */
+	public function ensureUuid(?string $existingUuid = null): void {
+		if ($this->uuid() !== null) {
+			return;
+		}
+		if ($existingUuid !== null && $existingUuid !== '') {
+			$this->setUuid($existingUuid);
+			return;
+		}
+		$this->setUuid(Uuid::v7()->toRfc4122());
```
blizzz, line 69: Not sure having it here. It should be exclusively generated when creating an option first time.

blizzz (commented):
Brief look, left a comment. But a UUID should never be regenerated, it must remain stable once assigned upon creation.

lib/Model/SelectionOption.php
samin-z: thanks for the review, i refactored it.

lib/Model/SelectionOptions.php
```diff
@@ -95,6 +95,34 @@ public static function createFromInputJsonString(?string $data, null|bool|int|st
...
+		if ($this->selectionOptions === null) {
+			return;
+		}
+
+		$existingByKey = [];
+		if ($existing !== null) {
+			foreach ($existing as $existingOption) {
+				$existingByKey[$existingOption->key()] = $existingOption;
+			}
+		}
```
enjeck, line 111: Do we need to worry about protecting agains t `TypeError` if stored options are `null` ? 
```suggestion
		if ($existing !== null && $existing->jsonSerialize() !== null) {
			foreach ($existing as $existingOption) {
				$existingByKey[$existingOption->key()] = $existingOption;
			}
		}
```

lib/Model/SelectionOptions.php
```diff
@@ -95,6 +95,34 @@ public static function createFromInputJsonString(?string $data, null|bool|int|st
 		return self::createFromInputArray($inputArray, $default, $allowPassingUuid);
 	}
 
+	/**
+	 * UUIDs are never regenerated or taken from client input
```
enjeck, line 99: is this comment still true? `Column::fromDto` parses with `allowPassingUuid = true`?

tests/unit/Model/SelectionOptionUuidTest.php
enjeck, line 1: maybe add a test for the existing options being null. The potential issue I mentioned earlier

lib/Model/SelectionOption.php
```diff
@@ -49,7 +47,35 @@ public function label(): string {
...
+			return;
+		}
+		$this->setUuid($uuid);
+	}
+
+	/**
+	 * Generate a UUID when creating an option for the first time, existing ones are not replaced
+	 */
+	public function generateUuid(): void {
+		if ($this->uuid() !== null) {
```
enjeck: Is it necessary to check if `uuid() !== null` ? Wondering if the guard in`setUuid` is enough already

lib/Model/SelectionOptions.php
blizzz: Create from Tables Structure should set them. And upcoming structure update as well.

lib/Model/SelectionOptions.php
blizzz: When `$existing` is not null, it is `SelectionOption` and then it must have a key. So the `jsonSerialize()` check is not necessary.

lib/Model/SelectionOption.php
blizzz: ack, seems duplicated

blizzz (approved):
Tested successfully on top of table-scheme-import

Added a commit to re-run migration for the selection options

lib/Model/SelectionOption.php
samin-z: the duplicated check is already removed, setUuid() is no-ops if a UUID is set.

PR #2900, opened 2026-08-20
chore: add playwright to .nextcloudignore

come-nc:
Same thing as cypress, do not ship tests to production


blizzz (approved):

come-nc:
(force-pushed to add playwright.config.ts as well)

PR #2911, opened 2026-08-23
[main] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency


blizzz:
only there was nothing about nextcloud/ocp :upside_down_face:

PR #2917, opened 2026-08-26
feat(deps): Add Nextcloud 36 support

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


AndyScherzinger (approved):

blizzz:
Wanted to merge it only after the release. Should have rejected then, my fault. Is what it is now.

AndyScherzinger:
Well, would only be an issue if somebody installs 36 and there is no new version by the and this one would be breaking. Else fine, not even a need for a release, so it could also work to one's benefit 😏

PR #2940, opened 2026-08-30
Fix: add contains-item filter support in views

Koc:
This is a follow-up for the #2387. I've realized that BE support is missing for `contains-item` filter.

### 🏁 Checklist
 
- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Db/Row2Mapper.php
+++ b/lib/Db/Row2Mapper.php
@@ -342,6 +342,9 @@ class Row2Mapper {
 	private function replacePlaceholderValues(array &$filters, string $userId): void {
 		foreach ($filters as &$filterGroup) {
 			foreach ($filterGroup as &$filter) {
+				if (is_array($filter['value'])) {
+					continue;
+				}
 				if (str_starts_with((string)$filter['value'], '@')) {
 					$columnId = (int)($filter['columnId'] ?? 0);
 					$column = $columnId > 0 ? $this->columnMapper->find($columnId) : null;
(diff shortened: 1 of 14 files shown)


lib/Db/Row2Mapper.php
```diff
@@ -478,6 +478,36 @@ private function getFilterExpression(IQueryBuilder $qb, Column $column, string $
...
+					if (is_string($filterItems)) {
+						$filterItems = json_decode($filterItems, true) ?? null;
+					}
+					if (!is_array($filterItems) || $filterItems === []) {
+						$filterItems = [(string)$value];
+					}
+				}
+				$filterExpressions = [];
+				$includeDefault = false;
+				foreach ($filterItems as $filterItem) {
```
samin-z, line 510: one small point, when 'value' is an array of option objects something like:
[[{\"columnId\":{columnId},\"operator\":\"contains-item\",\"value\":[{\"id\":0,\"label\":\"Active\"},{\"id\":2,\"label\":\"Pending\"}]}]]"
when saving it can throw 500 error: 'Filter::__construct(): Argument #3 ($value) must be of type string, array given'
we can either reject none-string 'value' with proper error message or normalize the array before construction the 'filer'.

Koc: he-he, this small point changes a lot :smile: . I've updated code to support this one as well, checked all possible scenarios + covered them with Behat end Playwright tests.

image

Pipeline is green except upgrade (which is not related to my changes)

samin-z (approved):

PR #2946, opened 2026-09-02
feat: Allow image previews of Files image links

### LINKED ISSUE #580: Add column types File/Image/Filelist/Imagelist

my usecase: i want to keep track of band shirts i already own and want to see which design i already own at the merch booth of bands/festivals so a image would be way easier then a description of the design

i would like to link a images or a list of images to a row in the table that are already located in my nextcloud
and also see a preview of the/these images in the row view

maybe even create a config option per table how large the preview images per row should be

_No response_

### Additional context

_No response_

enjeck:
You can create an Link column type and enable previews. Images in the Files app would be shown:

Capture d’écran 2026-09-02 à 07 19 55

Shown like so, and the size is customizable:

Capture d’écran 2026-09-02 à 07 12 23
Capture d’écran 2026-09-02 à 07 12 53

Fixes https://github.com/nextcloud/tables/issues/580


samin-z:
@enjeck works like a charm :)
one small nitpick though, maybe for future follow-up we can have MIME-based detection for cases where there is no extension (if that is ever the case), otherwise all good to me.

samin-z (approved):

enjeck:
- [ ] check what happens when you share table to someone without access to the images. I expect broken link, but to confirm the UX is good

AndyScherzinger:
In theory you could do it like the files app and if it is an internal file link of any kind, check for a preview? Than you could also show previews for documents for example 🚀 Could of course also just be a follow-up or a thing "for later". In any case, love the feature 🎉

samin-z:
> * [ ]  check what happens when you share table to someone without access to the images. I expect broken link, but to confirm the UX is good
@enjeck  yes it is a broken link redirecting user to files 
Screenshot 2026-09-04 at 10 07 39

AndyScherzinger (approved):

enjeck:
Now, we preview other files like Markdown and PDF:
Capture d’écran 2026-09-08 à 08 30 59

AndyScherzinger:
Woohoo 🎉

samin-z:
@enjeck i pulled all your latest changes and when i upload a pdf this is how the is shown for me still, is there something im missing?
Screenshot 2026-09-08 at 16 21 04

AndyScherzinger:
@samin-z did you enable the preview provider in the config.php? (If you also only see the mimetype icon in the files view, than the provider is turner off)

samin-z:
> @samin-z did you enable the preview provider in the config.php? (If you also only see the mimetype icon in the files view, than the provider is turner off)

@AndyScherzinger yeah i noticed that and that's what im currently doing

samin-z:
@enjeck all good

blizzz (changes requested):
We had in a conversation before. If the URL is not a local one, then the browser will do an outreach to an external site. This should not be a security issue… but it can be a privacy issue, as you could leak IP or browser details to a foreign website. Whether this is an issue, it depends… if the field is set from outside (e.g. via public link share) than it would not be trustworthy. If from a user on the instance, it depends on the relationship there. In this hindsight, I am not sure whether we should or have to nail this down a little. (The scenario could also be a user on the instance sharing bad links to outside people). 

If we align on having only internal images anyway we can validate against the domain or or just strip it and use a relative path. 

- otherwise, `ColumnService::validateCustomSettings()` could also guard `showPreview/imagePreviewSize`.
- also changing the setting goes not update exisitng rows.
- ResponseDefinitions needs to have `customSettings` updated
- All previews are fetched at once, this could lead to 100 requests at towards /core/preview. Claude suggests  *Adding `loading="lazy"` and `decoding="async"` to the `` is a one-line change that makes this safe on large tables.*
- It also suggested to limit the preview sizes to 32/64/96/128/192 which is generated by NC.

enjeck:
> * It also suggested to limit the preview sizes to 32/64/96/128/192 which is generated by NC.

@blizzz I addressed all except this last point. I don't see why it's better to have these fixed sizes as opposed to a number range between 32 and 192

blizzz:
> > * It also suggested to limit the preview sizes to 32/64/96/128/192 which is generated by NC.
> 
> @blizzz I addressed all except this last point. I don't see why it's better to have these fixed sizes as opposed to a number range between 32 and 192

The previews themselves will be normalized to match those: https://github.com/nextcloud/server/blob/master/lib/private/Preview/Generator.php#L457-L462. Actually 64 is the lowest number, not 32. Of course it is still possible to render them in different size, they will be then re-scaled again, but in the browser.

I am fine to leave it as is.

AndyScherzinger:
I'd vote for merge/ship expecting (community) feedback on it which may or may not lead to further improvements on the feature itself.

blizzz (approved):
There's  two items open about TextLinkForm.vue, but do not sound blocking to me. Not sure how likely the first could be, the second could be a papercut, but nothing terrible either.

- a failed /search/providers request leaves providers = [], and mounted() then forces showPreview = false, writing it into customSettings. canSave stays true, so a transient network blip silently discards the user's setting on save.
- isImagePreviewSizeInvalid only renders a NoteCard; it never sets this.error, so Save stays enabled. The component already owns that channel for the "no provider selected" case.

enjeck:
> The previews themselves will be normalized to match those: https://github.com/nextcloud/server/blob/master/lib/private/Preview/Generator.php#L457-L462. Actually 64 is the lowest number, not 32. Of course it is still possible to render them in different size, they will be then re-scaled again, but in the browser.

Not even sure what the UI would look like if we have a fixed set of sizes. Radio buttons? Dropdown? 🤔

enjeck:
> There's two items open about TextLinkForm.vue, but do not sound blocking to me. Not sure how likely the first could be, the second could be a papercut, but nothing terrible either.
> 
> * a failed /search/providers request leaves providers = [], and mounted() then forces showPreview = false, writing it into customSettings. canSave stays true, so a transient network blip silently discards the user's setting on save.
> * isImagePreviewSizeInvalid only renders a NoteCard; it never sets this.error, so Save stays enabled. The component already owns that channel for the "no provider selected" case.

i fixed these

blizzz:
> > The previews themselves will be normalized to match those: https://github.com/nextcloud/server/blob/master/lib/private/Preview/Generator.php#L457-L462. Actually 64 is the lowest number, not 32. Of course it is still possible to render them in different size, they will be then re-scaled again, but in the browser.
> 
> Not even sure what the UI would look like if we have a fixed set of sizes. Radio buttons? Dropdown? 🤔

My gut feeling would have been a drop down, maybe design would have an opinion on that, too.

blizzz (approved):

blizzz:
Just a sign-off missing at 4d53e349441e1fb5e41a0763b578e6364b5f661a

PR #2948, opened 2026-09-03
fix: Correctly handle selectionDefault initialization and normalizati…

enjeck:
…on in SelectionMultiForm

### to reproduce the bug this fixes

- Create new column with type multiple selection and define a value as default.
- Then save.
- All rows now receive the default value according to the definition.
- Now open the column again to edit.
- The default value is not checked any more
- When I selected the original value again as the default, the default value was then entered twice in the field
- If I add a new row, the same value is entered three times in the field, even though in edit mode it is not selected at all.


blizzz (approved):

blizzz:
@enjeck only a sign-off left

PR #2953, opened 2026-09-04
fix(sidebar): keep views the table listing does not carry

blizzz:
Views come from more source (owned, and shared) and there was a race condition that would reset the list of views instead of extending it.

### 🏁 Checklist
 
## 🤖 AI (if applicable)


blizzz:
Screenshot_20260904_230507

Playwright fails, but is green?

PR #2979, opened 2026-09-06
feat: allow multiple values on relation columns

### LINKED ISSUE #2952: feat: add relation column type (multiple)

in feature - add relation column type - [#2248](https://github.com/nextcloud/tables/pull/2248), the relation to a table/view column was introduced, but only one relation (value) can be chosen, multiple values are not possible

in the column options of an relation type column there should be the option to allow multiple values or a single value to pick from

_No response_

### Additional context

_No response_

he-is-talha:
## Summary
- Add a **Select multiple items** option on relation columns (`customSettings.allowMultiple`)
- Store relation values as one cell row per related id (same pattern as usergroup)
- Update create/edit UI, cell display, validation, import parsing, and analytics formatting for single and multi relations

Closes #2952

### 🖼️ Screenshots
image
image
image

## Notes
- API cell values for relation columns are now always `int[]` (legacy scalars are still accepted and normalized)
- No DB migration; existing single-value cells continue to work

## 🤖 AI (if applicable)
The content of this PR was partly using AI


samin-z:
@he-is-talha please clean you second commit message before merging.

he-is-talha:
@samin-z Cleaned the second commit message and force-pushed.

samin-z:
few points:

1. for multiple item when column is edited and it is switched to single item, table still shows multiple items until page is refreshed and then the correct value is shown.
2. editing multiple items and switching to single item, it shows no warning that existing data might be loss so maybe a nice addition to let users know, or show a warning underneath this option in regard to how toggling this will effect.
3. when the column is mandatory, clearing the relation still save with no error, it's better to make sure empty values are rejected both in the UI and on the API. same point for multi-selection, empty array shouldnt count as a valid value. Inline cell edit should also refuse to clear a mandatory relation. though i think this is pre-existing issue as if we have columns which rows which are empty and then the column is updated to mandatory but existing empty rows columns are still as they are.

for point 4 lets fix it for now for this case though.

he-is-talha:
1. Implemented: Multi → single updates without refresh
2. Warning when turning off multi:
image
3. Mandatory rejects empty
image
4. Enabling mandatory with empty rows blocked
image

he-is-talha:
@samin-z Ready for the review again

Aveyron-RetD:
Maybe @Koc would be a good reviewer

src/shared/components/ncTable/partials/columnTypePartials/forms/RelationForm.vue
```diff
@@ -5,10 +5,10 @@
 <template>
 	<div style="width: 100%">
 		<div class="row space-T">
-			<div class="fix-col-4 title">
+			<div class="form-col-4 title">
```
blizzz, line 8: The rename does not rename the styling. Intentional? Why is the rename necessary in first place?

lib/Service/ColumnTypes/RelationBusiness.php
```diff
@@ -58,42 +47,121 @@ public function canBeParsed($value, ?Column $column = null): bool {
...
+		if ($value === null || $value === '' || $value === []) {
+			return [];
+		}
+
+		if (is_string($value)) {
+			$decoded = json_decode($value, true);
+			if (json_last_error() === JSON_ERROR_NONE) {
+				$value = $decoded;
+			} elseif (str_contains($value, ',')) {
+				$value = array_map(trim(...), explode(',', $value));
```
blizzz, line 120: i.e. labels would not be able to contains `,` I am not aware that such a restriction is in place anyhwere?

lib/Service/ColumnTypes/RelationBusiness.php
```diff
@@ -58,42 +47,121 @@ public function canBeParsed($value, ?Column $column = null): bool {
...
+	 * @return list<int>
+	 * @throws BadRequestError
+	 */
+	private function normalizeToIds(mixed $value, Column $column, bool $throwOnInvalid = false): array {
+		if ($value === null || $value === '' || $value === []) {
+			return [];
+		}
+
+		if (is_string($value)) {
+			$decoded = json_decode($value, true);
```
blizzz, line 116: when this is not a json string, integer values would turn into a string, and then may not survive the check in `resolveRelationId()`

lib/Db/Row2Mapper.php
```diff
@@ -757,7 +794,18 @@ private function parseEntities(IResult $result, array $sleeves): array {
 		}
 
 		foreach ($rowValues as $compositeKey => $value) {
-			$rows[$keyToRowId[$compositeKey]]->addCell($keyToColumnId[$compositeKey], $value);
+			$columnId = $keyToColumnId[$compositeKey];
+			try {
+				$column = $this->columnMapper->find($columnId);
+				if ($column->getType() === Column::TYPE_RELATION
+					&& !(bool)($column->getCustomSettingsArray()[Column::RELATION_ALLOW_MULTIPLE] ?? false)
```
blizzz, line 801: this bleeds customSettings into Row2Mapper which it should not know about.

lib/Db/Row2Mapper.php
```diff
@@ -757,7 +794,18 @@ private function parseEntities(IResult $result, array $sleeves): array {
...
+					&& is_array($value)) {
+					$value = $value[0] ?? null;
+				}
+			} catch (DoesNotExistException) {
+				// Column may have been deleted; keep aggregated value as-is
+			}
```
blizzz, line 807: this column was already fetched in line 776, hence the check is not necessary know (but does not hurt). What would be better is to use the object that was already fetched.

lib/Service/ColumnTypes/RelationBusiness.php
```diff
@@ -58,42 +47,121 @@ public function canBeParsed($value, ?Column $column = null): bool {
...
 	public function validateValue(mixed $value, Column $column, string $userId, int $tableId, ?int $rowId): void {
-		if ($value === null || $value === '') {
+		$ids = ($value === null || $value === '' || $value === [])
+			? []
+			: $this->normalizeToIds($value, $column, throwOnInvalid: true);
+
+		if ($column->getMandatory() && $ids === []) {
+			throw new BadRequestError('Relation column is mandatory and cannot be empty');
+		}
+
```
blizzz, line 86: Isn't this already enforced in `RowService::validateMandatoryColumns()`? Plus more completely (i.e. view specific settings). Can be dropped there then.

lib/Db/RowCellRelationMapper.php
```diff
@@ -21,16 +21,81 @@ public function __construct(IDBConnection $db) {
...
+			$rowId = (int)$row['row_id'];
+			if (isset($seenRows[$rowId])) {
+				$idsToDelete[] = (int)$row['id'];
+			} else {
+				$seenRows[$rowId] = true;
+			}
+		}
+		$result->closeCursor();
+
+		foreach (array_chunk($idsToDelete, 500) as $chunk) {
```
blizzz, line 72: use a constanct with the value of 1k as already in other places, pls: `private const DB_CHUNK_SIZE = 1_000;`

blizzz: Scalingwise it is better to run the delete above, when we already have the number of IDs in `$idsToDelete`. It can be emptied afterwards and will be softer on the ram consumption (think huge tables).

lib/Db/Row2Mapper.php
```diff
@@ -585,6 +609,19 @@ private function getFilterExpression(IQueryBuilder $qb, Column $column, string $
 					$filterExpression = $qb->expr()->neq('value', $qb->createNamedParameter('[' . $this->db->escapeLikeParameter($value) . ']', $paramType));
 					break;
 				}
+				if ($column->getType() === Column::TYPE_RELATION) {
```
blizzz, line 612: duplicated with lines 565ff

lib/Service/ColumnTypes/RelationBusiness.php
```diff
@@ -58,42 +47,121 @@ public function canBeParsed($value, ?Column $column = null): bool {
...
+		if ($value === null || $value === '' || $value === []) {
 			return true;
 		}
 
-		return false;
+		try {
+			$this->normalizeToIds($value, $column, throwOnInvalid: true);
+			return true;
+		} catch (BadRequestError) {
+			return $this->normalizeToIds($value, $column, throwOnInvalid: false) !== [];
```
blizzz, line 74: this probably does not need to be called a second time and can be solved differently.

lib/Service/RowService.php
```diff
@@ -519,6 +519,12 @@ private function isValueValidForMandatoryColumn($value, Column $column, IColumnT
 			$defaultValue = $column->getSelectionDefault();
 			return $defaultValue !== null && $defaultValue !== '' && $defaultValue !== '[]';
 		}
+		if ($column->getType() === Column::TYPE_RELATION) {
+			if (is_array($value)) {
+				return count($value) > 0;
+			}
+			return $value !== null && $value !== '';
```
blizzz, line 526: this check was already run in line 503

src/shared/components/ncTable/partials/TableCellRelation.vue
```diff
@@ -142,7 +169,23 @@ export default {
...
+			newValue = newValue.map(id => parseInt(id)).filter(id => !Number.isNaN(id))
+
+			if (!this.allowMultiple) {
+				newValue = newValue.slice(0, 1)
+			}
+
+			if (this.isMandatory && newValue.length === 0) {
+				showError(t('tables', 'This column is mandatory and cannot be empty.'))
+				this.cancelEdit()
+				this.localLoading = false
```
blizzz, line 184: it is already false (line 168)

src/shared/components/ncTable/mixins/columnsTypes/relation.js
```diff
@@ -72,15 +80,34 @@ export default class RelationColumn extends AbstractColumn {
 	 * @return {boolean} Whether the filter matches
 	 */
 	isFilterFound(cell, filter) {
-		const filterValue = (filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value).toLowerCase()
-		const cellLabel = this.getLabel(cell.value)?.toLowerCase()
+		const rawFilter = filter.magicValuesEnriched ? filter.magicValuesEnriched : filter.value
+		const filterValue = String(rawFilter ?? '').toLowerCase()
+		const ids = this.normalizeIds(cell?.value ?? cell)
```
blizzz, line 85:
```suggestion
		const ids = this.normalizeIds(cell?.value)
```

should be just this?

lib/Db/Row2Mapper.php
blizzz, line 314: fyi, sorting would be by label id not display value.

blizzz (changes requested):
Thanks for the contriubtion!

If it is correct, that the PR is also AI assisted, please add the "Assisted-by: $AGENT_NAME:$MODEL_VERSION" trailer to each commit where it applies to.

src/shared/components/ncTable/partials/columnTypePartials/forms/RelationForm.vue
he-is-talha: Reverted — the rename wasn’t needed.

lib/Service/ColumnTypes/RelationBusiness.php
he-is-talha: Good catch. Names with commas are kept as one label now; we only split when the full name doesn’t match.

lib/Service/ColumnTypes/RelationBusiness.php
he-is-talha: Fixed — number IDs are handled correctly now.

lib/Db/Row2Mapper.php
he-is-talha: Agreed. Moved that logic out of Row2Mapper.

lib/Db/Row2Mapper.php
he-is-talha: Done — we reuse the column we already loaded.

lib/Service/ColumnTypes/RelationBusiness.php
he-is-talha: You’re right. Removed the duplicate check; the existing mandatory handling covers it.

lib/Db/RowCellRelationMapper.php
he-is-talha: Done — now using the shared 1k batch size.

he-is-talha: Done — we clean up in batches as we go so big tables stay light on memory.

lib/Db/Row2Mapper.php
he-is-talha: Fixed — shared that logic in one place.

lib/Service/ColumnTypes/RelationBusiness.php
he-is-talha: Simplified — only one pass now.

lib/Service/RowService.php
he-is-talha: Removed the extra check.

src/shared/components/ncTable/partials/TableCellRelation.vue
he-is-talha: Removed the redundant line.

src/shared/components/ncTable/mixins/columnsTypes/relation.js
he-is-talha: Applied, thanks.

lib/Db/Row2Mapper.php
he-is-talha: Thanks for the note — left as-is for now.

he-is-talha:
@blizzz Thanks for the review! Addressed the inline comments. Also added the `Assisted-by` trailer on the follow-up commit.

PR #2987, opened 2026-09-07
fix(import): avoid crash when table or view is deleted before import job runs

silverkszlo:
The `ImportTableJob` was throwing the following error multiple times:   
```
Did expect one result but found none when executing: query
  "SELECT * FROM tables_tables WHERE id = :dcValue1"
```
If the table (or the view whose table this resolves) is deleted between when the import job is queued and when the background worker picks it up, `TableMapper::find()/ViewMapper::find()` throws `DoesNotExistException`, which propagates uncaught out of `run()`.

Wrap the table/view lookup in a try/catch so a deleted table or view is logged as a warning and the import-finished activity is skipped instead of throwing.

### 🏁 Checklist
 
- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/lib/BackgroundJob/ImportTableJob.php
+++ b/lib/BackgroundJob/ImportTableJob.php
@@ -12,6 +12,7 @@ use OCA\Tables\Db\TableMapper;
 use OCA\Tables\Db\ViewMapper;
 use OCA\Tables\Notification\NotificationHelper;
 use OCA\Tables\Service\ImportService;
+use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Utility\ITimeFactory;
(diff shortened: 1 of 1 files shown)


blizzz (approved):

lib/BackgroundJob/ImportTableJob.php
Koc, line 76: should we move checks for view/table existence above? there is no too much sense to call `importV2` here if target is not exists IMHO

benjaminfrueh (approved):

Koc (approved):

blizzz:
Playwright tests should be unrelated, fix attempt at https://github.com/nextcloud/tables/pull/2995

PR #2988, opened 2026-09-08
fix(tests): toggle link providers by their switch role

AndyScherzinger:
`createTextLinkColumn()` looked the "Allowed types" provider toggles up with `getByRole('checkbox', …)`, but the accessibility tree shows they expose the **switch** role, so the locator matched nothing and `setCheckboxState()` returned without toggling anything. Every pre-activated provider therefore stayed enabled (`preActivatedProviders` is `url, files, contacts`), so a URL-only column rendered the provider `NcSelect` instead of the plain `NcTextField`, the value typed by `.fill()` went into the select's search box and was never committed, and the row saved without it - which is why `column-text-link.spec.ts:16` fails on `main` across all four Nextcloud versions. 

The toggles are now addressed by their switch role and the requested provider is asserted to exist, so a future role change fails the test loudly instead of silently disabling the column setup. 

This is a test-side fix only; no application code is touched.

### 🏁 Checklist
 
- [x] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/playwright/support/commands.ts
+++ b/playwright/support/commands.ts
@@ -605,22 +605,22 @@ export async function createTextLinkColumn(
 		.click()
 	await expect(page.getByText('Allowed types')).toBeVisible()
 
-	await setCheckboxState(
-		page.getByRole('checkbox', { name: /^URL$/i }),
-		false,
-	)
-	await setCheckboxState(
-		page.getByRole('checkbox', { name: /^Files$/i }),
-		false,
-	)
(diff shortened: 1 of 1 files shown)


enjeck (commented):
Looks like the tests are flaky? It seems like they failed on the initial run

blizzz:
Superseded by https://github.com/nextcloud/tables/pull/2995, isn't it?  At least partly. Mind double check whether this is still necessary and needs and update or can be closed? Atm the tests succeed on main.

